PR #15, opened 2020-06-04
Add support for editing tags on resource groups 🏷

EricJizbaMSFT:
I implemented a pared down version of VS Code's `FileSystemProvider` that only supports editing single files (aka no read directory, etc.). This should hopefully replace `AzureBaseEditor` in the UI package - I'll do that once I make sure this works for at least one other scenario (probably App Service since I think that's the only thing using `AzureBaseEditor` at the moment).

I went the file system route because I think it's easier to view/edit tags in a file than a wizard and it felt like displaying tags in the tree would clutter it up too much. Users still get basic validation in the form of standard VS Code problems:

Screen Shot 2020-06-03 at 5 00 40 PM

I plan to support editing tags on resources in a future PR.

### DIFF
--- a/src/commands/tags/TagFileSystem.ts
+++ b/src/commands/tags/TagFileSystem.ts
@@ -1,0 +1,95 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ResourceManagementClient } from "azure-arm-resource";
+import * as jsonc from 'jsonc-parser';
+import * as os from "os";
+import { commands, Diagnostic, FileStat, FileType, MessageItem, Uri, window } from "vscode";
+import { createAzureClient, IActionContext } from 'vscode-azureextensionui';
+import { ext } from "../../extensionVariables";
+import { ResourceGroupTreeItem } from "../../tree/ResourceGroupTreeItem";
+import { localize } from "../../utils/localize";
+import { AzExtFileSystem } from "./AzExtFileSystem";
+import { getTagDiagnostics } from "./getTagDiagnostics";
+
+const insertKeyHere: string = localize('insertTagName', '<Insert tag name>');
+const insertValueHere: string = localize('insertTagValue', '<Insert tag value>');
+
+/**
+ * For now this file system only supports editing tags.
+ * However, the scheme was left generic so that it could support editing other stuff in this extension without needing to create a whole new file system
+ */
+export class TagFileSystem extends AzExtFileSystem<ResourceGroupTreeItem> {
+    public static scheme: string = 'azureResourceGroups';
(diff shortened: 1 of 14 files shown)


src/commands/tags/TagFileSystem.ts
```diff
@@ -0,0 +1,95 @@
...
+                }
+            });
+
+            context.errorHandling.suppressDisplay = true;
+            // This won't be displayed, but might as well track the first diagnostic for telemetry
+            throw new Error(diagnostics[0].message);
+        } else {
+            const message: string = localize('confirmTags', 'Are you sure you want to update tags for resource group "{0}"?', node.name);
+            const update: MessageItem = { title: localize('update', 'Update') };
+            await ext.ui.showWarningMessage(message, { modal: true }, update);
```
nturinski, line 66: No setting to disable the confirmation prompt?  I feel like most of our FileEditors do this-- maybe it could just be a property in the editor itself?

src/commands/tags/registerTagDiagnostics.ts
```diff
@@ -0,0 +1,49 @@
...
+function onDidChangeActiveTextEditor(context: IActionContext, editor: TextEditor | undefined): void {
+    context.telemetry.suppressIfSuccessful = true;
+    context.errorHandling.suppressDisplay = true;
+
+    if (editor?.document.uri.scheme === TagFileSystem.scheme) {
+        if (!ext.diagnosticWatcher) {
+            ext.diagnosticWatcher = workspace.onDidChangeTextDocument(onDidChangeTextDocument);
+        }
+        updateTagDiagnostics(editor.document);
+    } else if (ext.diagnosticWatcher) {
```
nturinski, line 28: How/When does this get created/set?  I guess you're clearing it when the scheme (which I assume is azureResourceGroup) doesn't match?  So you dispose it anytime the person opens a file that is not opened from the extension?

EricJizbaMSFT: It's set here: https://github.com/microsoft/vscode-azureresourcegroups/pull/15/files#diff-852c7ed0a0de11b8c9f3efd1029cd634R25

It's really just a disposable for the `onDidChangeTextDocument` event, which is fired _every time the user types something_. I set it when they open a tags file and dispose it when they open a different file. Just trying to make sure we don't hurt the performance of normal files

src/commands/tags/TagFileSystem.ts
EricJizbaMSFT: If we could do something like this checkbox, I would be all for it (not sure if extensions can do this, though):
Screen Shot 2020-06-04 at 1 43 02 PM

I think this is a little ugly/confusing:
Screen Shot 2020-06-04 at 1 58 22 PM

I would rather wait for users to complain about the confirmation dialog before I spend time making it ugly lol (at least for this scenario)

> maybe it could just be a property in the editor itself?

Not sure what you mean by this. One thing to note is that the warning message is in `TagFileSystem`, not `AzExtFileSystem` - so each extension can do whatever it wants for its specific scenario

Issue #21, opened 2020-06-15
[Suggestion] Add an error in "PROBLEMS" when naming the tag with whitespace/empty


v-ruizh:
**OS:** All
**Build Version:** 20200612.3
**Regression:** Not a Regression

**Repro Steps:**
Right click one resource group -> Edit Tags... -> Add a tag with a whitespace/empty name.

**Actual:**
There is no error with the whitespace/empty name. Ctrl+S and click "Update" button, an error occurs.
![image](https://user-images.githubusercontent.com/45614674/84635142-6288ba80-af25-11ea-93e0-1dba456d50b1.png)

**Suggestion:**
Add the above error in "PROBLEMS" when naming the tag with whitespace/empty, like the error when naming the tag with "<>?/".

PR #23, opened 2020-06-15
Show tag diagnostic if name is empty

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/21


src/commands/tags/getTagDiagnostics.ts
```diff
@@ -82,10 +82,15 @@ class TagVisitor implements jsonc.JSONVisitor {
                 this.addError(range, localize('tagNameTooLong', 'Tag name must be {0} characters or less.', max));
             }
 
+            if (/^\s*$/.test(property)) {
```
wwlorey, line 85: nit: I think using something like `property.trim() === ''` would be a lil more readable than the regex but maybe that wouldn't have the exact same behavior

PR #51, opened 2020-12-03
Hide some resource types

ejizba:
Modeled after the portal:
Screen Shot 2020-12-03 at 2 42 58 PM

Unfortunately I don't know all the types that should be hidden, but I figured I'd at least start with the types I saw on my subscription


nturinski (approved):
If only there was just some property on Azure resources... :|

PR #59, opened 2021-02-16
Add "Help and Feedback" view

ejizba:
And make the extension a bit more generic to prepare for being a dependency of other Azure extensions
- "Resources" instead of "Resource Groups"
- Collapse both views by default

Screen Shot 2021-02-16 at 11 38 47 AM

### DIFF
--- a/src/azureExtensions.ts
+++ b/src/azureExtensions.ts
@@ -46,7 +46,8 @@ export const azureExtensions: IAzExtMetadata[] = [
         tutorial: {
             label: localize('deployStatic', 'Deploy a static app'),
             url: 'https://aka.ms/AAb5xp8'
-        }
+        },
+        reportIssueCommandId: 'staticWebApps.reportIssue'
     },
     {
         name: 'vscode-azureresourcegroups',
@@ -59,14 +60,16 @@ export const azureExtensions: IAzExtMetadata[] = [
         label: 'Storage',
         resourceTypes: [
             'microsoft.storage/storageaccounts'
-        ]
+        ],
+        reportIssueCommandId: 'azureStorage.reportIssue'
     },
     {
         name: 'vscode-azurevirtualmachines',
         label: 'Virtual Machines',
         resourceTypes: [
             'microsoft.compute/virtualmachines'
-        ]
+        ],
+        reportIssueCommandId: 'azureVirtualMachines.reportIssue'
     },
     {
         name: 'vscode-azureeventgrid',
@@ -82,7 +85,8 @@ export const azureExtensions: IAzExtMetadata[] = [
         resourceTypes: [
             'microsoft.documentdb/databaseaccounts',
             'microsoft.dbforpostgresql/servers'
-        ]
+        ],
+        reportIssueCommandId: 'azureDatabases.reportIssue'
     }
 ];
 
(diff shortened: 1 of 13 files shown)


src/azureExtensions.ts
```diff
@@ -0,0 +1,118 @@
...
+        ],
+        reportIssueCommandId: 'azureVirtualMachines.reportIssue'
+    },
+    {
+        name: 'vscode-azureeventgrid',
+        label: 'Event Grid',
+        resourceTypes: [
+            'microsoft.eventgrid/eventsubscriptions',
+            'microsoft.eventgrid/topics'
+        ]
```
nturinski, line 80: lol don't even offer them a report issue for Event Grid, huh

src/commands/helpAndFeedback/reportIssue.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export async function reportIssue(context: IActionContext): Promise<void> {
+    const picks: IAzureQuickPickItem<AzExtWrapper | undefined>[] = getInstalledExtensionPicks();
+    picks.push({
+        label: localize('other', 'Other'),
+        data: undefined
+    });
+
+    const placeHolder: string = localize('selectExtension', 'Select the Azure extension you want to report an issue on');
+    const azExtension: AzExtWrapper | undefined = (await ext.ui.showQuickPick(picks, { placeHolder, suppressPersistence: true })).data;
```
nturinski, line 20: Do we want to turn off persistence?  I could see it being useful if you always report issues on the same extension.

src/commands/helpAndFeedback/reportIssue.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (azExtension) {
+        context.telemetry.properties.extension = azExtension.name;
+        const commandId: string | undefined = await azExtension.getReportIssueCommandId();
+        if (commandId) {
+            await commands.executeCommand(commandId);
+        } else {
+            await openNewIssuePage(azExtension.name);
+        }
+    } else {
+        await openNewIssuePage('vscode-azuretools');
```
nturinski, line 30: I feel like "Other" should direct to here (resource-groups) rather than azuretools personally.

nturinski (approved):

src/azureExtensions.ts
ejizba: Yeah...  that extension is on an old version of the ui package that doesn't support the report issue command. Just not worth it at this point

src/commands/helpAndFeedback/reportIssue.ts
ejizba: Any particular reason?

nturinski: This feels like the extension that kind of encompasses general Azure issues whereas azuretools is kind of more specific to our dev issues.  Also, I don't feel like we actively triage issues in azuretools, but we have a triage set up for Resources already.

ejizba: Discussed offline. Don't want to dilute resourcegroups stuff, but agree azuretools is a bit too "internal". Will use azcode repo

src/commands/helpAndFeedback/reportIssue.ts
ejizba: My thinking on this was - I don't like persistence for seldom-used quick picks because the thing I did a week or two ago probably isn't what I want to do now. If people are _not_ using this seldomly, I feel like we have bigger problems 😝  I also feel like persistence is more important for long wizards and this is pretty simple

Issue #53, opened 2020-12-17
No error notification pops up after saving a tag with "("test"):"123"


v-mengwe:
**OS:** Win10
**Build Version:** 20201216.1

**Repro Steps:**

1. Create a new resource group. -> Right click this resource group --> Click "Edit Tags...".
2. Edit the tag with `"("test"):"123"` -> CTRL+S.
3. Check whether pop up an error notification.

**Expect:**
An error notification pops up.
![image](https://user-images.githubusercontent.com/62053665/102466675-49a2c280-408a-11eb-9073-c2f5979c57bc.png)

**Actual:**
No error notification pops up.
![re](https://user-images.githubusercontent.com/62053665/102466961-ac945980-408a-11eb-8d1b-bf74da4c2ee3.gif)

**More Info:**
The tag changes the default name and value after refreshing.
![image](https://user-images.githubusercontent.com/62053665/102471242-f3388280-408f-11eb-8169-d311e3e39004.png)

v-chenzha:
Verified this issue on the latest main branch build 20230509.1, it still reproduces for now. Details show as below.

1. This issue doesn't reproduce on Vs Code 1.78.0.
2. This issue doesn't reproduce on Linux and Mac OS.

Amar-QA:
This issue has been fixed in the build 20240612.1.
![image](https://github.com/microsoft/vscode-azureresourcegroups/assets/156269329/09e02de9-497f-4605-932f-a2c6310c065a)

Issue #18, opened 2020-06-15
jsonc errors don't block editing tags


v-ruizh:
**OS:** All
**Build Version:** 20200612.3
**Regression:** Not a Regression

**Repro Steps:**
1. Right click one resource group -> Edit Tags... -> Add two tags with the same name in the tags.jsonc file.
![image](https://user-images.githubusercontent.com/45614674/84625064-1897d880-af15-11ea-86be-5ff7007c3115.png)
2. Ctrl+S.

**Expect:**
An error occurs.

**Actual:**
Prompt the below dialog to update tags for resource group. Click "Update", it will successfully updated tags for resource group.
![image](https://user-images.githubusercontent.com/45614674/84625447-d7ec8f00-af15-11ea-9877-14415a30483f.png)

**More Info:**
1. The updated tags only contain the final one of the two duplicate tags.
![image](https://user-images.githubusercontent.com/45614674/84625690-63feb680-af16-11ea-9cc2-c6746825ae07.png)
2. An error occurs when adding a duplicate tag in portal with the "Save" button disabled.
![image](https://user-images.githubusercontent.com/45614674/84624977-f56d2900-af14-11ea-97a3-0610040b9275.png)

v-mengwe:
Verified on the latest build 20201216.1, this issue also reproduces for GitHub codesapces.

PR #61, opened 2021-02-17
Block on jsonc errors as well as our own Azure errors when editing tags

ejizba:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/53
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/18

### DIFF
--- a/src/commands/tags/TagFileSystem.ts
+++ b/src/commands/tags/TagFileSystem.ts
@@ -6,14 +6,13 @@
 import { ResourceManagementClient } from "@azure/arm-resources";
 import * as jsonc from 'jsonc-parser';
 import * as os from "os";
-import { commands, Diagnostic, FileStat, FileType, MessageItem, Uri, window } from "vscode";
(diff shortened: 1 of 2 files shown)


neelip:
When I am adding to duplicated keys, I do see "Duplicate object key" inline warning but I can still save the document successfully. Is the intention to only prompt inline warning?

ejizba:
Uh hmm interesting - jsonc considers those to be "warnings", not "errors". I added an Azure-specific error to catch it, modeled after the portal:
Screen Shot 2021-02-17 at 5 20 51 PM

neelip:
nit: The jsonc warning is case-sensitive
image
image

nturinski (approved):

ejizba:
> nit: The jsonc warning is case-sensitive

Correct, I don't control the jsonc warnings/errors. Azure is more strict than jsonc, which is why I created our own error to enforce that

PR #73, opened 2021-04-29
Switch to eslint

ejizba:


wwlorey (approved):
Should be able to remove this too: https://github.com/microsoft/vscode-azureresourcegroups/blob/ej/eslint/.vscode/settings.json#L17

PR #76, opened 2021-05-06
Update SVG for Azure view

nturinski:
Old----vs----New:
image image

The bad news is, it seems like if you have an extension installed that _doesn't_ rely on Azure Resources (like VMs for instance) the icon will _not_ update.  At least when I tried in debug mode.  I'm not entirely sure what gets priority in this case (as in, maybe if someone installs Resource Groups first and then VMs, they will see the new icon?)


neelip:
I see what you mean. The new icon is showing if the VM ext. is not installed and it stays the same right after I install it. But it goes back to the old icon once I reload vscode.

neelip (commented):
I see what you mean. The new icon is showing up if the VM ext. is not installed and it stays the same right after I install it. But it goes back to the old icon once I reload vscode. 

Maybe VM overwrites Azure icon?

ejizba (approved):
Guess we'll just have to update VMs extension :)

PR #79, opened 2021-05-11
Show total of number of resources getting deleted

### LINKED ISSUE #66: Show the number of resources that will be deleted when confirming deletion
> Are you sure you want to delete resource group "rg"? This will delete all _n_ resources in the resource group.

nturinski:
Fixes #66

I looked into the `ResourceManagementClient` briefly and didn't see any commands to get a total headcount of resources, so I figured the best way to do so was to just load all the children if they aren't loaded and display the children array's length.

Looks like this:
image
image

It looks a little silly when there's 0 or 1 resources, but wasn't sure if it was worth the extra logic to detect that:
image
image


neelip:
The one with 1 resource makes sense but the one with 0 resources might get a little confusing. 
We might need some logic just for that.

src/commands/deleteResourceGroup.ts
```diff
@@ -23,11 +23,13 @@ export async function deleteResourceGroup(context: IActionContext, primaryNode?:
 
     const deleteConfirmation: string | undefined = settingUtils.getWorkspaceSetting('deleteConfirmation');
     for (const node of selectedNodes) {
+        const numOfResources = await node.getNumOfResources(context);
+
         if (deleteConfirmation === 'ClickButton') {
-            const areYouSureDelete: string = localize('areYouSureDelete', 'Are you sure you want to delete resource group "{0}" and all its resources?', node.name);
+            const areYouSureDelete: string = localize('areYouSureDelete', 'Are you sure you want to delete resource group "{0}". This will delete all {1} resource(s) in the resource group?', node.name, numOfResources);
```
neelip, line 29: The one with 1 resource makes sense but the one with 0 resources might get a little confusing. 
We might need some logic just for that.

neelip (approved):

src/commands/deleteResourceGroup.ts
ejizba: What if we go similar to the portal:
Screen Shot 2021-05-11 at 10 15 05 AM

I feel like that sentence would make sense even for 0.

ejizba: Also, either way, the question mark should go on the "Are you sure" sentence, not the second one.

nturinski: Yeah, I like the way that looks quite a bit better.  Though it doesn't read correctly if there is 1 resource.

image

image

It's the kind of thing that kind of bothers me enough to have logic for it, but I dunno 😒

ejizba (approved):

PR #115, opened 2022-02-23
POC for orderBy WIP implementation

nturinski:
🙏


src/extensionVariables.ts
```diff
@@ -21,4 +22,10 @@ export namespace ext {
     export let tagFS: TagFileSystem;
     export let diagnosticWatcher: Disposable | undefined;
     export let diagnosticCollection: DiagnosticCollection;
+
+    export const resourceProviders: { [key: string]: ApplicationResourceProvider } = {};
```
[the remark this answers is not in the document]
nturinski: Well, I thought all extensions would be considered resource providers including our Azure extensions.  Does the SWA extension not register its own resource to provide the SWA tree items?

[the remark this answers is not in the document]
alexweininger: We need to be able to fetch Azure resources separately from the resource extensions (SWA). We need this capability so we can A. list resources quickly without waiting for the other extensions to activate, and B. show resources even if a user doesn't have the associated extension installed. 

So RGs will register an `AzureResourceProvider` which uses the arm-resources SDK to fetch resources.

See my implementation here:
https://github.com/microsoft/vscode-azureresourcegroups/blob/98ee1b6a61097c0ee97cf2ac539d99a4f33c8d0e/src/AzureResourceProvider.ts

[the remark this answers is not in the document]
alexweininger: What SWA will register is a resolver that takes the generic resources fetched by RGs and the arm-resources SDK and "returns" StaticWebAppTreeItems.

See here: https://github.com/microsoft/vscode-azurestaticwebapps/blob/17dfbd82dd8acb167cfafc0472c22f3611109736/src/StaticWebAppResolver.ts

alexweininger (commented):
It's only showing the first child of each type. 

image

alexweininger (approved):

src/api.d.ts
```diff
@@ -12,7 +12,7 @@ export interface TreeNodeConfiguration {
     readonly contextValue?: string;
 }
 
-export interface ResolvableTreeItem {
+export interface IResolvableTreeItem extends GroupableResource {
```
nturinski, line 15: I think we decided to stop using the I convention for interfaces

src/extensionVariables.ts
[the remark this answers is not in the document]
bwateratmsft: Yeah, Alex is right. For now I'm expecting there to be only one resource provider (registered into itself by RG extension), with the feature extensions providing resolvers.

That being said, since _someday_ we might make it a map, why not just make it a map now?

src/extensionVariables.ts
```diff
@@ -21,4 +22,10 @@ export namespace ext {
...
+    export const groupByKeys: { [key: string]: string } = {
```
bwateratmsft, line 27: Seems like these should maybe go into a constants file?

src/tree/GroupTreeItemBase.ts
```diff
@@ -0,0 +1,38 @@
...
+import { localize } from "../utils/localize";
+import { ResolvableTreeItem } from "./ResolvableTreeItem";
+import { ShallowResourceTreeItem } from "./ShallowResourceTreeItem";
+
+export abstract class GroupTreeItemBase extends AzExtParentTreeItem {
+    public readonly childTypeLabel: string = localize('resource', 'Resource');
+    public treeMap: { [key: string]: (ResolvableTreeItem | ShallowResourceTreeItem) } = {};
+    public abstract label;
+
+    public readonly cTime: number = Date.now();
```
bwateratmsft, line 16: Are these really required now?!

src/tree/GroupTreeItemBase.ts
```diff
@@ -0,0 +1,38 @@
...
+            // this.treeMap = {};
+        }
+
+        for (const ti of Object.values(this.treeMap)) {
+            if (ti instanceof ResolvableTreeItem) {
+                void ti.resolve(clearCache, context);
+            }
+        }
+
+        return <AzExtTreeItem[]><unknown>Object.values(this.treeMap);
```
bwateratmsft, line 36: I'm not sure TS will let you, but if it does then IMO this is cleaner:
```suggestion
        return Object.values(this.treeMap) as AzExtTreeItem[];
```

src/tree/ResourceGroupTreeItem.ts
```diff
@@ -10,43 +10,43 @@ import { ext } from "../extensionVariables";
 import { createResourceClient } from "../utils/azureClients";
 import { localize } from "../utils/localize";
 import { treeUtils } from "../utils/treeUtils";
+import { GroupTreeItemBase } from "./GroupTreeItemBase";
 import { ResolvableTreeItem } from "./ResolvableTreeItem";
 import { ShallowResourceTreeItem } from "./ShallowResourceTreeItem";
 
-export class ResourceGroupTreeItem extends AzExtParentTreeItem {
+export class ResourceGroupTreeItem extends GroupTreeItemBase {
```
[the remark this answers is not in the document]
alexweininger: Agreed, should this be done later or in this PR?

src/tree/GroupTreeItemBase.ts
alexweininger: Yes, that works.

src/extensionVariables.ts
alexweininger: Well these aren't "traditional" constants since the contents of the objects will be changing quite a bit, so I think they fit in more here.

src/tree/GroupTreeItemBase.ts
alexweininger: It's used here, but I'm unsure what purpose `cTime` actually serves.

https://github.com/microsoft/vscode-azureresourcegroups/blob/d38b07978ba0c41516f8cf9f9e4eb5d35528ca55/src/commands/tags/TagFileSystem.ts#L31

src/tree/ResourceGroupTreeItem.ts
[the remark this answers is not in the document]
bwateratmsft: It can be later

src/tree/GroupTreeItemBase.ts
nturinski: Yeah, I think it was mostly to see if the tags needed to be updated by comparing the times, but as per this comment we can probably expand its capabilities:

```
/**
 * For now this file system only supports editing tags.
 * However, the scheme was left generic so that it could support editing other stuff in this extension without needing to create a whole new file system
 */
```

PR #116, opened 2022-02-24
PoC for activation manager

bwateratmsft:
This can activate extensions using the custom `x-onAzNodeTypeFetch` and `x-onAzNodeTypeResolve` activation events.

This can support having multiple extensions listening to the same type, and also handles extension enable/disable/install/uninstall gracefully.

### DIFF
--- a/src/utils/ExtensionActivationManager.ts
+++ b/src/utils/ExtensionActivationManager.ts
@@ -1,0 +1,86 @@
+/*!--------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+
+const builtInExtensionIdRegex = /^vscode\./i;
+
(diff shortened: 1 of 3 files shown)


src/utils/ExtensionActivationManager.ts
```diff
@@ -0,0 +1,86 @@
+/*!--------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+
+const builtInExtensionIdRegex = /^vscode\./i;
+
+const activationEventRegex = /^x-onAzNodeType(?<trigger>Fetch|Resolve):(?<type>[\w.\/]+)$/;
```
bwateratmsft, line 10: Should this maybe be `x-ms-onAzNodeTypeFetch`? With `-ms`.

alexweininger: I don't have a preference, I think the `x-` is enough, but also no harm in adding the `-ms`.

bwateratmsft:
Closing in favor of #117

PR #118, opened 2022-03-09
App resource resolvers now return an object instead of an AzExtTreeItem

alexweininger:
SWA branch for testing: https://github.com/microsoft/vscode-azurestaticwebapps/tree/alex/unified-v2

- Changes were fairly simple, a lot of renamed things
- api.d.ts is updated with most recent changes
- Almost no changes to Proxy since it works mostly the same regardless of using a class or plain object
- Largest change was made in the SWA extension itself, in StaticWebAppResolver https://github.com/microsoft/vscode-azurestaticwebapps/blob/alex/unified-v2/src%2FStaticWebAppResolver.ts

Relevant code from the StaticWebAppResolver that returns the needed properties from a StaticWebAppTreeItem as a plain object:
```ts
const ti = new StaticWebAppTreeItem(undefined as unknown as AzExtParentTreeItem, swa);

const resolved: ResolvedStaticWebApp = {
	id: ti.id ?? resource.id,
	contextValue: ti.contextValue,
	description: ti.description,
	iconPath: ti.iconPath,
	label: ti.label,
	name: ti.name,
	branch: ti.branch,
	defaultHostname: ti.defaultHostname,
	repositoryUrl: ti.repositoryUrl,
	resourceGroup: ti.resourceGroup,
	loadMoreChildrenImpl: ti.loadMoreChildrenImpl,
	deleteTreeItemImpl: ti.deleteTreeItemImpl
}
```

Oh yeah, commands haven't been updated in SWA so most are broken since they use instanceof.

I tried not to touch anything that I didn't need to, so grouping by resource group is (still?) broken.

Once grouping by RG is fixed, I'd say this is demo-able.


src/commands/openInPortal.ts
```diff
@@ -6,10 +6,10 @@
 import { openInPortal as uiOpenInPortal } from '@microsoft/vscode-azext-azureutils';
 import { IActionContext } from '@microsoft/vscode-azext-utils';
 import { ext } from '../extensionVariables';
+import { AppResourceTreeItem } from '../tree/AppResourceTreeItem';
 import { ResourceGroupTreeItem } from '../tree/ResourceGroupTreeItem';
-import { ResourceTreeItem } from '../tree/ResourceTreeItem';
 
-export async function openInPortal(context: IActionContext, node?: ResourceGroupTreeItem | ResourceTreeItem): Promise<void> {
+export async function openInPortal(context: IActionContext, node?: ResourceGroupTreeItem | AppResourceTreeItem): Promise<void> {
```
bwateratmsft, line 12: Obviously this is a longer-term thing but actions like open in portal / reveal / remove / etc. are the perfect candidates for having just one command defined in the RG extension.

src/tree/ResolvableTreeItemBase.ts
```diff
@@ -0,0 +1,85 @@
...
+            this.resolveResult = {
+                id: this.data.id,
+                contextValue: 'noResolver',
+                label: 'No resolver',
+                description: 'No resolver found for this resource',
+                iconPath: new ThemeIcon('error'),
+            }
+        }
+
+        await this.runWithTemporaryDescription(context, 'Resolving...', async () => {
```
bwateratmsft, line 47:
```suggestion
        await this.runWithTemporaryDescription(context, 'Loading...', async () => {
```

Users won't know what "Resolving" means

src/utils/ExtensionActivationManager.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
bwateratmsft, line 1: This is disabling across the entire file which is probably not what we want

src/tree/ShallowResourceTreeItem.ts
```diff
@@ -94,12 +94,12 @@ export class ShallowResourceTreeItem extends AzExtTreeItem implements GroupableR
         switch (groupBySetting) {
```
bwateratmsft, line 94: What does this class do again?

src/api.d.ts
```diff
@@ -12,81 +35,168 @@ export interface TreeNodeConfiguration {
...
-export declare function registerApplicationResourceResolver(
-    provider: ApplicationResourceResolver,
-    resourceType: string,
-    resourceKind?: string,
-): vscode.Disposable;
-
 // Resource Groups can have a default resolve() method that it supplies, that will activate the appropriate extension and give it a chance to replace the resolve() method
 // ALSO, it will eliminate that default resolver from future calls for that resource type
 
 // called from host extension (Resource Groups)
```
bwateratmsft, line 196:
```suggestion
```
Not the case anymore

bwateratmsft (approved):

src/commands/openInPortal.ts
bwateratmsft: No action needed, just wanted to put the comment down.

src/tree/ShallowResourceTreeItem.ts
alexweininger: It's for resources that have no extension associated with them. https://github.com/microsoft/vscode-azureresourcegroups/blob/bc86665054c257331e7d6efbdcf4ed28f399a455/src/tree/SubscriptionTreeItem.ts#L55-L66

src/api.d.ts
alexweininger: `registerApplicationResourceProvider` is still called from the host extension right?

bwateratmsft: You're right, my brain was thinking this was the register...Resolver which is called from the client extensions. The host extension will call `registerApplicationResourceProvider` (on itself).

src/tree/ShallowResourceTreeItem.ts
bwateratmsft: We could have a fallback resolver that handles this scenario built into the host extension. I'm imagining that same resolver could either attach a resolved item that will make the item click show the raw JSON (for things that don't have an extension), or the extension marketplace page (for things that do).

alexweininger: Ok, I'm going to merge this and work on that later.

bwateratmsft: Good plan, I was gonna suggest the same.

alexweininger: I think it's actually best to not have a fallback resolver for the resources that don't have extensions, and instead have a fallback class. Because our ResolvableTreeItemBase is an AzExtParentTreeItem so it expects to have children. And I don't think we plan on resources without extensions having children.

We could split `ResolvableTreeItemBase` into `ResolvableParentTreeItemBase`, but I think that would make more sense for VMs (resource with an extension that doesn't have children).

alexweininger: `hasMoreChildrenImpl` is for implementing the load more action. Is there another way?

> hasMoreChildrenImpl
> Implement this as a part of the "Load more..." action. Should not be called directly
> @returns — 'true' if there are more children and a "Load more..." node should be displayed

bwateratmsft: What's the down side of having it be a parent item though? We may also be able to set the collapsible state to `None` which I think makes it appear non-expandable.

bwateratmsft: Gotcha. I'll double check about `collapsibleState` and make sure it's doing what we want.

bwateratmsft: Ok good, it does do what we want!

With `collapsibleState` set to expanded (which we use by default for compose grouping):
image

With it set to none:
image

PR #119, opened 2022-03-10
Changes to the resolve process

bwateratmsft:
This PR makes some changes to the resolve process.

1. `ShallowResourceTreeItem` and `InstallableResourceTreeItem` are more easily accomplished by resolvers, that add the necessary commands.
2. All resources are guaranteed to have a resolver (though it might be the `NoopResolver`).
3. When a new resolver registers, all resources of that type (and maybe later type+kind), that are visible, will be re-resolved. Along with (2), this removes the need to wait for the client extension to activate.


src/api.d.ts
```diff
@@ -157,16 +157,14 @@ export interface AbstractAzExtTreeItem {
...
-export type ResolvedAppResource = ResolvedAppResourceTreeItemBase;
-
-export type ResolvedAppResourceTreeItem<T extends ResolvedAppResourceTreeItemBase> = AppResource & Omit<T, keyof ResolvedAppResourceTreeItemBase>;
+export type ResolvedAppResourceTreeItem<T extends ResolvedAppResourceBase> = AppResource & SealedAzExtTreeItem & Omit<T, keyof ResolvedAppResourceBase>;
 
 export type LocalResource = AzExtTreeItem;
 
 export interface AppResourceResolver {
-    resolveResource(subContext: ISubscriptionContext, resource: AppResource): vscode.ProviderResult<ResolvedAppResource>;
+    resolveResource(subContext: ISubscriptionContext, resource: AppResource): vscode.ProviderResult<ResolvedAppResourceBase>;
```
nturinski, line 167: Should we add `isApplicable` as something that must be defined as well? All resolvers probably need this, right?

src/tree/AppResourceTreeItem.ts
```diff
@@ -34,9 +33,11 @@ export class AppResourceTreeItem extends ResolvableTreeItemBase implements Group
         this.rootGroupConfig = <GroupNodeConfiguration><unknown>parent;
 
         this.data = resource;
-        this.commandId = 'azureResourceGroups.revealResource';
+        this.commandId = 'azureResourceGroups.revealResource'; // TODO: is this still needed?
```
nturinski, line 36: I think we don't need the command for RG, but we need to provide a way for other extensions to use "revealResource".   A user may want to reveal a resource on the tree especially after creation.

I think we could let them use `revealResource` like so:
https://github.com/microsoft/vscode-azurestaticwebapps/commit/4ec265304fb94485d5389c820c5ba25e7e82265e#diff-a448840ab47b15ffc93675e7af136d99e1bc309a7ed03d64f126ea049aa54803R15

src/api.d.ts
```diff
@@ -157,16 +157,14 @@ export interface AbstractAzExtTreeItem {
     isAncestorOfImpl?(contextValue: string | RegExp): boolean;
 }
 
-export type ResolvedAppResourceTreeItemBase = Partial<{ [P in keyof SealedAzExtTreeItem]: never }> & AbstractAzExtTreeItem;
+export type ResolvedAppResourceBase = Partial<{ [P in keyof SealedAzExtTreeItem]: never }> & Partial<AbstractAzExtTreeItem> & { contextValue?: never, contextValues?: string[] };
```
nturinski, line 160: Should contextValues be a `string[]`?  In our implementation, we're using the `Set<string>`

src/resolvers/InstallableAppResourceResolver.ts
```diff
@@ -0,0 +1,31 @@
...
+ * This resolver acts as a "placeholder" resolver when an extension is known for the resource type but not installed (or not enabled).
+ * Upon the resolved node being clicked, it will open the extension page on the marketplace to allow them to easily install (or enable).
+ */
+class InstallableAppResourceResolver implements AppResourceResolver {
+    public resolveResource(_subContext: ISubscriptionContext, resource: AppResource): ResolvedAppResourceBase {
+        // We know the extension is known but uninstalled, or else it would not have passed the `isApplicable` check below
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        const extensionId = getAzureExtensions().find(azExt => azExt.matchesResourceType(resource))!.id;
+
+        return {
```
nturinski, line 20: I think we can explore using the vscode.command to just install the extension instead of taking them to the marketplace page.

nturinski (approved):

src/api.d.ts
bwateratmsft: That's a good idea. A way to be more intelligent than simple type checking. It also solves our type/kind dilemma.

alexweininger (approved):

src/resolvers/InstallableAppResourceResolver.ts
bwateratmsft: I'd like to hear from @fiveisprime on that too; personally I lean toward showing them the marketplace page rather than auto-installing as a sort of "good faith" way of making sure they know exactly what they're doing. Or you could call it C.Y.A. :smile:

src/api.d.ts
alexweininger: Hmm... I am normally a fan of modern ES features like Set but I been on the other side and dealt with having to polyfill these features with core-js and webpack is a nightmare. I'm not sure if this is a problem in the vscode extension ecosystem though.

I agree that renaming it to `contextValuesToAdd` (or something similar) would be good.

src/resolvers/InstallableAppResourceResolver.ts
fiveisprime: Yeah. As much as I'd love to auto-install, it just doesn't quite feel right.

src/tree/AppResourceTreeItem.ts
nturinski: Oh, my impression was that it revealed and highlighted it in the treeView even if the node was collapsed.

src/resolvers/InstallableAppResourceResolver.ts
alexweininger: Yeah it shows this modal after running the `workbench.extensions.installExtension` command.
image

fiveisprime: Oh nice, I dig that! :shipit:

bwateratmsft: That's neat! I'll make that change. I agree, long as we are giving them this prompt I'm happy.

src/tree/AppResourceTreeItem.ts
bwateratmsft: Ok, I'll need to review, this might be the wrong command ID. Ultimately the one we need is whichever is the one opening the raw JSON.

src/resolvers/InstallableAppResourceResolver.ts
bwateratmsft: @nturinski @fiveisprime @alexweininger I ran into one problem with the `workbench.extensions.installExtension`. If the extension is _installed_ but _disabled_, when you execute the command nothing happens. As far as I know it is not possible for us to tell the difference between uninstalled and disabled--either way the VSCode extension API returns `undefined`. I also do not see any commands for enabling extensions.

I don't know how common of a problem we expect this to be, but it is a point in favor of just opening the marketplace page.

src/tree/AppResourceTreeItem.ts
bwateratmsft: Ok, the correct command is `azureResourceGroups.viewProperties`, I have updated it.

src/api.d.ts
bwateratmsft: Ok, I renamed it to `contextValuesToAdd`. I also created an interface, `ContextValuesToAdd`, instead of the inline declaration--this lets us have documentation attached so intellisense can help users know what to do. I'm not sold on the interface name `ContextValuesToAdd` so if you have ideas I'm all ears.

src/api.d.ts
bwateratmsft: I did these changes, though I called it `matchesResource` to align with some similar functions already on our APIs.

src/resolvers/InstallableAppResourceResolver.ts
alexweininger: Ok, I think we should talk to Isidor about providing a better API for prompting to install extensions. But for now, taking users to the marketplace page seems like the safe bet.

PR #120, opened 2022-03-10
Expose TreeView and Tree

nturinski:
To be used with https://github.com/microsoft/vscode-azurestaticwebapps/tree/nat/removeTreeProvider

Each extension shouldn't have its own tree provider.  It'd make searching for nodes impossible, so the RG extension should expose its tree provider to the other extensions so they can call `revealTreeItem` and `findTreeItem`.


src/api.d.ts
```diff
@@ -6,6 +6,15 @@
 import { AzExtParentTreeItem, AzExtTreeDataProvider, AzExtTreeItem, IActionContext, ISubscriptionContext, TreeItemIconPath } from '@microsoft/vscode-azext-utils';
 import * as vscode from 'vscode';
 
+export interface AzureResourceGroupsExtensionApi {
+    tree: AzExtTreeDataProvider;
```
bwateratmsft, line 10: It probably doesn't make much difference in the grand scheme of things, but we could make all of these `readonly`.

bwateratmsft, line 10: It probably doesn't make much difference in the grand scheme of things, but we could make all of these `readonly`.

src/extension.ts
```diff
@@ -53,8 +56,11 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
     });
 
     return createApiProvider([
```
bwateratmsft, line 58: Along with the above change making the items `readonly` on the interface, we could make this a class instance and all these fields `get` methods only. The underlying instance variables could be `#private` members (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields).

I'm sure there are still ways around it but extra defense can't hurt.

bwateratmsft, line 58: Along with the above change making the items `readonly` on the interface, we could make this a class instance and all the fields `get` methods only.

I'm sure there are still ways around it but extra defense can't hurt.

bwateratmsft (approved):

src/api/AzureResourceGroupsExtensionApi.ts
```diff
@@ -0,0 +1,24 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzExtTreeDataProvider, AzExtTreeItem } from '@microsoft/vscode-azext-utils';
+import { Disposable, TreeView } from 'vscode';
+import { AppResourceResolver, AzureResourceGroupsExtensionApi } from '../api';
+
+export class InternalAzureResourceGroupsExtensionApi implements AzureResourceGroupsExtensionApi {
+    readonly tree: AzExtTreeDataProvider;
```
bwateratmsft, line 11: Isn't `readonly` on a class just TypeScript sugar? JavaScript has `#private` fields that we can wrap with `get`'ers only. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

nturinski: Oh, yeah you're right.  TypeScript will yell at you for trying to set it, but I'll do this implementation instead.

nturinski: I think the way I did it makes sense?

The interface in the api will still just have `readonly` properties.  That's what is being implemented and when we create the api, we are only enforcing those properties are passed in.  

Those are used to create the actual ExtensionApi which uses private variables that shouldn't be able to be overwritten, but since it's private, I don't think end-users would need to know about that implementation detail, which is why just leaving it as `readonly` makes sense in the api, right?

PR #124, opened 2022-03-14
Workspace tree view api

alexweininger:
Implements registerLocalResourceProvider and adds a new Workspace tree view to the extension. We may want to explore the value of adding a `priority?: number` parameter to registerLocalResourceProvider so that we can control the sorting of the tree items (if we don't want A-z).

I implemented the API in the Functions extension.
https://github.com/microsoft/vscode-azurefunctions/compare/alex/unified-workspace

Unrelated to the code in this PR, but I tried out moving the Create Function command to the Local Project tree item, since in the future the top ribbon could get super cluttered with commands.
image

### DIFF
--- a/src/api/regsiterLocalResourceProvider.ts
+++ b/src/api/regsiterLocalResourceProvider.ts
@@ -7,7 +7,6 @@ import { callWithTelemetryAndErrorHandlingSync } from "@microsoft/vscode-azext-u
 import { Disposable } from "vscode";
 import { LocalResourceProvider } from "../api";
 import { refreshWorkspace } from "../commands/workspace/refreshWorkspace";
-import { ext } from "../extensionVariables";
 
 export const localResourceProviders: Record<string, LocalResourceProvider> = {};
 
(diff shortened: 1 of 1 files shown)


src/commands/workspace/refreshWorkspace.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from "@microsoft/vscode-azext-utils";
+import { ext } from "../../extensionVariables";
+
+export async function refreshWorkspace(context: IActionContext): Promise<void> {
+    await ext.workspaceTree.refresh(context);
```
nturinski, line 10: Not sure if we really need a whole command for this line, but if we keep this, you should use it here https://github.com/microsoft/vscode-azureresourcegroups/pull/124/files?diff=split&w=1#diff-9cb9d4ec34a86167c68d8cb55f3d8de6b90eadf242b13a1f2e949608b1ea9401R21

alexweininger: Would you prefer putting the function body directly in the registerCommand arguments?

PR #134, opened 2022-03-22
Fix descriptions for leaf tree items

alexweininger:
Without this getter, descriptions present on the resolvedResult are never propagated up to the actual tree item. We may have to do this with some other properties in the future. Probably iconPath

### DIFF
--- a/src/tree/ResolvableTreeItemBase.ts
+++ b/src/tree/ResolvableTreeItemBase.ts
@@ -23,6 +23,10 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
         return Array.from(this.contextValues.values()).sort().join(';');
     }
 
+    public get description(): string | undefined {
+        return this.resolveResult?.description;
+    }
+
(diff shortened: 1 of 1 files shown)


src/tree/ResolvableTreeItemBase.ts
```diff
@@ -23,6 +23,10 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
         return Array.from(this.contextValues.values()).sort().join(';');
     }
 
+    public get description(): string | undefined {
```
bwateratmsft, line 26: The `Proxy` object isn't handling this automagically?

bwateratmsft (approved):

src/tree/ResolvableTreeItemBase.ts
alexweininger: I'm pretty sure it's due to how `effectiveDescription` works in AzExtTreeItem.

https://github.com/microsoft/vscode-azuretools/blob/main/utils/src/tree/AzExtTreeItem.ts#L45-L47

bwateratmsft: Gotcha. We should investigate if `iconPath` will also need it; seems like it might.

PR #140, opened 2022-03-24
Don't make GroupTreeItem ids lower case

alexweininger:
Was investigating an error I was getting trying to open storage blobs. Figured out that [`findTreeItem`](https://github.com/microsoft/vscode-azuretools/blob/d38498f0085deb912675e4d2cb376f973c12f31e/utils/src/tree/AzExtTreeDataProvider.ts#L261) is case sensitive and the fullIds were different only by the casing of `resourceGroups`.

`/subscriptions/843d979b-47f8-4bb9-9fbc-33e33962c304/resourcegroups/funcbasic1a9cbd88f7`
`/subscriptions/843d979b-47f8-4bb9-9fbc-33e33962c304/resourceGroups/funcbasic1ab151b770/...`

I'm not really sure how this is happening because I would think that if we `.toLowerCase()` it in one place, how could it be upper case when we search for it? But I must not be understanding something right because it works when I removed `.toLowerCase()`

### DIFF
--- a/src/utils/azureUtils.ts
+++ b/src/utils/azureUtils.ts
@@ -28,7 +28,7 @@ export function getResourceGroupFromId(id: string): string {
 export function createGroupConfigFromResource(resource: GenericResource, subscriptionId: string | undefined): GroupingConfig {
     const id = nonNullProp(resource, 'id');
     return {
-        resourceGroup: { keyLabel: 'Resource Groups', label: getResourceGroupFromId(id), id: id.substring(0, id.indexOf('/providers')).toLowerCase() },
+        resourceGroup: { keyLabel: 'Resource Groups', label: getResourceGroupFromId(id), id: id.substring(0, id.indexOf('/providers')) },
         resourceType: {
             keyLabel: 'Resource Types', label: resource.type?.toLowerCase() ?? 'unknown',
             id: `${subscriptionId}/${resource.type}` ?? 'unknown',


bwateratmsft (approved):

nturinski:
As a note, we should make all of the `fullId` and `id` properties lower case in the `findTreeItem` commands whenever it's compared to a tree item's id.

bwateratmsft:
> As a note, we should make all of the `fullId` and `id` properties lower case in the `findTreeItem` commands whenever it's compared to a tree item's id.
> 
> I think that [this](https://github.com/microsoft/vscode-azureresourcegroups/pull/141) would have also fixed this issue.

I wish TS had extension methods like C#, so we could do `'someString'.compare_no_case('another')`

alexweininger:
> I wish TS had extension methods like C#, so we could do `'someString'.compare_no_case('another')`

Technically you can, but it's usually [frowned upon](https://stackoverflow.com/a/48027169/5145807)

https://bobbyhadz.com/blog/typescript-extend-string-prototype
https://thewebdev.info/2022/03/19/how-to-extend-string-prototype-with-typescript/

bwateratmsft:
I can see why that would be frowned upon, since it's altering the original prototype. At least in C#, extension methods are just static methods. Doing `obj.extensionMethod()` is equivalent to `extensionMethod(obj)`; the former is just syntactic sugar that C# offers. TypeScript _could_ do the same thing, and just transpile `obj.extensionMethod()` to `extensionMethod(obj)`. Though, I suspect they've already been asked, and declined, and probably have good reasons.

alexweininger:
I think this caused a bug. Now I see why we were `.toLowerCase()`ing the ids. Now there are case sensitive duplicates in our tree.
image

PR #141, opened 2022-03-24
Fix `findTreeItem` by correctly linking `AppResourceTreeItem` to `GroupTreeItem`

nturinski:
For simplicity, I was just making the parent be the Subscription node since it was convenient for some tree actions.  We might have to be careful with some commands now (such as `deleteTreeItem`) because when deleted, it only removes it from the cache of the direct parent which means it likely won't be deleted from the subscription node's array of resources.  A refresh should hit the API again though, so it should be okay because we refresh whenever they change the grouping.


nturinski:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/142

alexweininger (approved):
We should remove this now https://github.com/microsoft/vscode-azureresourcegroups/blob/nat/treeItemPicker/src/tree/SubscriptionTreeItem.ts#L122-L137

PR #143, opened 2022-03-25
Use context menu for Group By setting and make some resolvers not collapsible

alexweininger:
I like this icon, but I'm open to suggestions.

image

Also made the Noop and Shallow resolver have `collapsibleState: TreeItemCollapsibleState.None`

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,12 +48,19 @@
     ],
     "main": "./main.js",
     "contributes": {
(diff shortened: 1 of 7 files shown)


bwateratmsft (approved):

bwateratmsft:
I like that icon too!

PR #146, opened 2022-03-29
Improve install extension experience

alexweininger:
Edited `InstallableAppResourceResolver` to improve the installation UX. Now it returns a node with a single child instructing users to install the associated extension.

image

To handle install, I'm calling `extension.open` and `workbench.extensions.installExtension` so that the marketplace page and the installation modal open together. I think opening both is better than opening one or the other.

installextension

We might need to add more logic later to handle if an extension is installed but disabled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -289,6 +289,10 @@
                 {
                     "command": "azureResourceGroups.revealResource",
                     "when": "never"
+                },
+                {
+                    "command": "azureResourceGroups.installExtension",
+                    "when": "never"
                 }
             ],
             "azureResourceGroups.groupBy": [
--- a/package.json
+++ b/package.json
@@ -63,6 +63,12 @@
                 "category": "Azure Resource Groups"
             },
             {
+                "command": "azureResourceGroups.installExtension",
+                "title": "Install Azure extension",
+                "icon": "$(extensions)",
+                "category": "Azure Resource Groups"
+            },
+            {
                 "command": "azureResourceGroups.createResource",
                 "title": "Create Resource...",
                 "category": "Azure Resource Groups",
--- a/src/commands/installExtension.ts
+++ b/src/commands/installExtension.ts
@@ -3,33 +3,10 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { DialogResponses, IActionContext } from "@microsoft/vscode-azext-utils";
-import { commands, MessageItem, window } from "vscode";
-import { getAzureExtensions } from "../AzExtWrapper";
-import { ext } from "../extensionVariables";
-import { localize } from "../utils/localize";
+import { IActionContext } from "@microsoft/vscode-azext-utils";
+import { commands } from "vscode";
 
 export async function installExtension(context: IActionContext, extensionId: string): Promise<void> {
     context.telemetry.properties.extensionId = extensionId;
     await commands.executeCommand('extension.open', extensionId);
-
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const extension = getAzureExtensions().find(azExt => azExt.id === extensionId)!;
-    const installMessageItem: MessageItem = {
-        title: localize('installExtension', 'Install extension')
-    }
-    // Use extension-specific message later
-    const result = await window.showInformationMessage(localize('installExtension', "Install the '{0}' extension to enable additional features?", extension.label), {
-        modal: true
-    }, installMessageItem, DialogResponses.cancel);
-
-    if (result === installMessageItem) {
-        context.errorHandling.suppressDisplay = true;
-        // Will prompt if settings sync is enabled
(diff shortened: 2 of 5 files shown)


bwateratmsft:
> We might need to add more logic later to handle if an extension is installed but disabled.

Can you check what happens in that case?

alexweininger:
> > We might need to add more logic later to handle if an extension is installed but disabled.
> 
> Can you check what happens in that case? IIRC if it was installed but disabled that install popup did nothing, so what might happen is **the extension page opens with no install popup**--which IMO is good enough.

Yes, that's exactly what happens.

src/resolvers/InstallableAppResourceResolver.ts
```diff
@@ -18,11 +20,20 @@ class InstallableAppResourceResolver implements AppResourceResolver, BuiltinReso
...
-        const extensionId = getAzureExtensions().find(azExt => azExt.matchesResourceType(resource))!.id;
+        const extension = getAzureExtensions().find(azExt => azExt.matchesResourceType(resource))!;
 
         return {
-            commandId: 'extension.open',
-            commandArgs: [extensionId],
+            loadMoreChildrenImpl: async () => {
+                const ti = new GenericTreeItem(undefined, {
+                    contextValue: 'installExtension',
+                    label: localize('installExtension', 'Install {0} extension...', extension.label),
```
nturinski, line 29: It might be too cluttered, but it would be good if we could include some text explaining why a user would want to install the extension.

Maybe something like:
`Install {0} extension for enhanced features...` or something like that.

alexweininger: We need Matt for this... 🪄✨

src/commands/installExtension.ts
```diff
@@ -0,0 +1,12 @@
...
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from "@microsoft/vscode-azext-utils";
+import { commands } from "vscode";
+
+export async function installExtension(_context: IActionContext, extensionId: string): Promise<void> {
+    void commands.executeCommand('extension.open', extensionId);
+    void commands.executeCommand('workbench.extensions.installExtension', extensionId);
```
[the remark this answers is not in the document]
bwateratmsft: It would already be happening via the `callWithTelemetryAndErrorHandling` implicitly wrapped around any command registered with `registerCommand`, so I think we're good.

package.json
```diff
@@ -62,6 +62,12 @@
                 "title": "%azureResourceGroups.createResourceGroup%",
                 "category": "Azure Resource Groups"
             },
+            {
+                "command": "azureResourceGroups.installExtension",
```
bwateratmsft, line 66: I'd configure this one to never show in the command palette, since it will not work if executed that way.

src/resolvers/InstallableAppResourceResolver.ts
alexweininger: Updated to this:
image

src/commands/installExtension.ts
```diff
@@ -3,10 +3,33 @@
...
+    }
+    // Use extension-specific message later
+    const result = await window.showInformationMessage(localize('installExtension', "Install the '{0}' extension to enable additional features?", extension.label), {
+        modal: true
+    }, installMessageItem, DialogResponses.cancel);
+
+    if (result === installMessageItem) {
+        context.errorHandling.suppressDisplay = true;
+        // Will prompt if settings sync is enabled
+        await commands.executeCommand('workbench.extensions.installExtension', extensionId);
```
bwateratmsft, line 29: I'm concerned about what happens when the extension is installed but disabled. AFAIK, VSCode offers no way to see if that is the case, nor a command to enable/disable extensions. If we just directly call the `workbench.extensions.installExtension` it's a noop, which is fine.

Additionally, if it's not installed, they'll be getting two modal popups in a row to install.

I wouldn't want to diminish the user experience to gain telemetry. Likely, VSCode has telemetry we can view for enablement/disablement and install/uninstall, so I'd say we capture the `extensionId` but otherwise just call `workbench.extensions.installExtension` unconditionally.

src/commands/installExtension.ts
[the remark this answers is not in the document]
nturinski: Yeah, I was thinking we wouldn't want to await either command, but I think it's okay to await the installExtension prompt.

src/commands/installExtension.ts
bwateratmsft: If we set up some way for the extension to activate immediately in this scenario, then the resolver flow will already take care of that--when the extension activates and the resolver registers, it will automatically re-resolve all those nodes. That way we don't need to await the install or refresh the tree.

Just need to change the activation manager to keep track of types it has already fetched/resolved, and if an extension comes in that requests activation based on a type that has already been seen, it gets immediately activated.

bwateratmsft (approved):

PR #148, opened 2022-03-30
Add display names and icons

alexweininger:
I copied names from the portal, but we can easily update any later if needed. Also there are a few that I couldn't find names in the portal for, like `microsoft.alertsmanagement/smartDetectorAlertRules`, not really sure what we want to do about those.

image


src/utils/azureUtils.ts
```diff
@@ -43,14 +44,23 @@ export function createGroupConfigFromResource(resource: GenericResource, subscri
     }
 }
 
-export function getIconPath(type?: string): TreeItemIconPath {
+function getId(subscriptionId?: string, type?: string, kind?: string): string {
+    if (type?.toLowerCase() === 'microsoft.web/sites') {
+        if (kind?.toLowerCase() === 'functionapp') {
+            return `${subscriptionId}/${kind}`;
```
bwateratmsft, line 50: I feel a little uneasy about reimagining the type as something else...can we make a supported type app that considers both type and optional kind as part of the key?

bwateratmsft (approved):

PR #149, opened 2022-03-30
Activate newly installed extensions immediately

bwateratmsft:
Based on https://github.com/microsoft/vscode-azureresourcegroups/pull/146#discussion_r838925631. If an extension is installed that would have already been activated by the activation manager (had it been installed), then we should immediately activate it.

### DIFF
--- a/src/api.d.ts
+++ b/src/api.d.ts
@@ -184,7 +184,7 @@ interface ContextValuesToAdd {
 
 export type ResolvedAppResourceBase = Partial<{ [P in keyof SealedAzExtTreeItem]: never }> & Partial<AbstractAzExtTreeItem> & ContextValuesToAdd;
 
-export type ResolvedAppResourceTreeItem<T extends ResolvedAppResourceBase> = AppResource & SealedAzExtTreeItem & Omit<T, keyof ResolvedAppResourceBase>;
+export type ResolvedAppResourceTreeItem<T extends ResolvedAppResourceBase> = SealedAzExtTreeItem & AbstractAzExtTreeItem & Omit<T, keyof ResolvedAppResourceBase>;
 
 export type LocalResource = AzExtTreeItem;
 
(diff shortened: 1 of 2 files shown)


src/api.d.ts
```diff
@@ -184,7 +184,7 @@ interface ContextValuesToAdd {
 
 export type ResolvedAppResourceBase = Partial<{ [P in keyof SealedAzExtTreeItem]: never }> & Partial<AbstractAzExtTreeItem> & ContextValuesToAdd;
 
-export type ResolvedAppResourceTreeItem<T extends ResolvedAppResourceBase> = AppResource & SealedAzExtTreeItem & Omit<T, keyof ResolvedAppResourceBase>;
+export type ResolvedAppResourceTreeItem<T extends ResolvedAppResourceBase> = SealedAzExtTreeItem & AbstractAzExtTreeItem & Omit<T, keyof ResolvedAppResourceBase>;
```
bwateratmsft, line 187: This is an unrelated fix to the API.

alexweininger:
How does this "listen" for newly installed extensions?

alexweininger:
How does this "listen" for newly installed extensions?

bwateratmsft:
> How does this "listen" for newly installed extensions?

It already was listening for new extension installs, but it wasn't previously immediately activating those that it would have activated had they been installed already. Line 34 (in the new PR version, line 37) adds a listener to `vscode.extensions.onDidChange`. Unfortunately VSCode doesn't offer any info on what changed (add, remove, upgrade...no data at all), so we just have to scan all the extensions again. Fortunately that takes very little time--2ms at most.

nturinski (approved):
Can't fully test this, but I see all the properties we're expecting when running it with the new extensions so 👍

bwateratmsft:
I was able to do a test by building a VSIX of the Docker extension with a fake `onResolve` listener. Everything worked 👍

PR #150, opened 2022-03-30
Fix openInPortal command

alexweininger:


bwateratmsft (approved):

src/commands/openInPortal.ts
```diff
@@ -14,5 +14,5 @@ export async function openInPortal(context: IActionContext, node?: ResourceGroup
         node = await ext.tree.showTreeItemPicker<ResourceGroupTreeItem>(ResourceGroupTreeItem.contextValue, context);
     }
 
-    await uiOpenInPortal(node, node.fullId);
+    await uiOpenInPortal(node, node.data?.id ?? node.fullId);
```
nturinski, line 17: Just an idea, but maybe we should have a property on the tree items called Azure id that we could use specifically for Azure related things.  I feel like it's kind of obtuse to know when to use id vs. fullId

bwateratmsft: I agree, that's a good idea. Easy thing to put on `AppResourceTreeItem`.

PR #155, opened 2022-04-06
Activity Log panel and API

alexweininger:
I'm currently exploring ways to integrate the Activity Log API directly with the AzureWizard, but wanted to get this PR out early to get preliminary feedback.

I plan to have most of this code live in the utils package, but for now I'll leave it here for the sake of reviewing.

Implemented in SWA: https://github.com/microsoft/vscode-azurestaticwebapps/pull/642


alexweininger:
I have a rough implementation hooked into the AzureWizard execute() method. 

Screen Recording 2022-04-06 at 3 05 19 PM

src/commands/revealResource.ts
```diff
@@ -4,45 +4,16 @@
...
-                context.telemetry.properties.revealError = parseError(error).message;
-            }
-        }
+export async function revealResource(context: IActionContext, resource: AppResource): Promise<void> {
+    context.telemetry.properties.resourceType = resource.type?.replace(/\//g, '|'); // Replace the slashes otherwise this gets redacted because it looks like a user file path
+    context.telemetry.properties.resourceKind = resource.kind;
+    try {
+        await revealTreeItem(resource.id);
+    } catch (error) {
+        context.telemetry.properties.revealError = parseError(error).message;
```
bwateratmsft, line 16: Should we maybe rethrow this?

src/api.d.ts
```diff
@@ -223,3 +224,33 @@ export declare function registerApplicationResourceProvider(id: string, provider
...
+    public done: boolean;
+    public error?: IParsedError;
+    public readonly task: () => Promise<R | undefined>;
+    public startedAtMs: number;
+
+    public constructor(task: () => Promise<R | undefined>);
+
+    public run(): Promise<R | undefined>;
+
+    public get state(): ActivityTreeItemOptions;
```
bwateratmsft, line 255: Maybe this to be extra clear?
```suggestion
    public get currentState(): ActivityTreeItemOptions;
```

src/activityLog/ActivityTreeItem.ts
```diff
@@ -0,0 +1,67 @@
...
+*--------------------------------------------------------------------------------------------*/
+
+import { AzExtParentTreeItem, AzExtTreeItem, IActionContext, TreeItemIconPath } from "@microsoft/vscode-azext-utils";
+import { ThemeColor, ThemeIcon, TreeItemCollapsibleState } from "vscode";
+import { localize } from "../utils/localize";
+import { ActivityBase } from "./Activity";
+
+export class ActivityTreeItem extends AzExtParentTreeItem {
+
+    public constructor(parent: AzExtParentTreeItem, public readonly activity: ActivityBase) {
```
bwateratmsft, line 13: Should take the interface instead of a specific class:
```suggestion
    public constructor(parent: AzExtParentTreeItem, public readonly activity: ActivityType) {
```

src/activityLog/Activity.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export interface ActivityTreeItemOptions {
+    label: string;
+    contextValuePostfix?: string;
+    collapsibleState?: TreeItemCollapsibleState;
+    children?: (parent: AzExtParentTreeItem) => AzExtTreeItem[];
+}
+
+interface ActivityType {
+    inital(): ActivityTreeItemOptions;
```
bwateratmsft, line 17: Nit: spelling (same misspelling appears elsewhere)
```suggestion
    initial(): ActivityTreeItemOptions;
```

src/activityLog/Activity.ts
```diff
@@ -0,0 +1,58 @@
...
+import { TreeItemCollapsibleState } from "vscode";
+
+export interface ActivityTreeItemOptions {
+    label: string;
+    contextValuePostfix?: string;
+    collapsibleState?: TreeItemCollapsibleState;
+    children?: (parent: AzExtParentTreeItem) => AzExtTreeItem[];
+}
+
+interface ActivityType {
```
bwateratmsft, line 16: Maybe just rename this `Activity`? (Renaming will affect some suggestions below)
```suggestion
interface Activity {
```
And export it so consumers of the API can use it instead of `ActivityBase`?

src/activityLog/Activity.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export type ActivityTask<R = void> = R extends void ? () => Promise<void> : () => Promise<R>;
+
+export abstract class ActivityBase<R = void> implements ActivityType {
+
+    abstract inital(): ActivityTreeItemOptions;
+    abstract onSuccess(): ActivityTreeItemOptions;
+    abstract onError(error: IParsedError): ActivityTreeItemOptions;
+
+    public done: boolean;
```
bwateratmsft, line 30: This is one of those times I sure wish TypeScript would take the `foo { get; private set; }` from C#. Still, I'd make `done`, `error`, and `startedAtMs` private with a public `get` method. I think that means you can also mark them `readonly` on the interface, TS seems to treat `readonly` on interfaces as read-only to externals, whereas on classes as read-only to everyone (including the class itself).

src/activityLog/ActivityTreeItem.ts
bwateratmsft: There may need to be a wrapper class built around `ActivityType` (or `Activity` if renamed), for tracking the done/error state, based on the calls to `onSuccess()` and `onError()`. This tree item can be that wrapper class.

src/activityLog/Activity.ts
```diff
@@ -0,0 +1,58 @@
...
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AzExtParentTreeItem, AzExtTreeItem, IParsedError, parseError } from "@microsoft/vscode-azext-utils";
+import { TreeItemCollapsibleState } from "vscode";
+
+export interface ActivityTreeItemOptions {
+    label: string;
+    contextValuePostfix?: string;
```
bwateratmsft, line 11: If possible I'd prefer to stick with the `ContextValuesToAdd` concept.

bwateratmsft (commented):
So it's a bit of a pivot from what you have, but I have another idea for an implementation. I'm concerned about having the API be attached to a class that the consumers are forced to use--same shortcomings as what we have with the tree API. (Having a class that they _can_ use is just fine, IMO, but I'm uneasy about one they _must_ use.)

My suggestion would be:
1. Define an `Activity` interface with three `Event<ActivityTreeItemOptions>`; one each for initial/success/error. (Or `EventLike` or something)
2. The `ActivityTreeItem`, when it is created, can manage the start time based on the call to `onInitial`, and can further register listeners to `onSuccess` / `onError`, to change its visible state and refresh, and hold state for done/error.
3. Implement a base class that takes in a simple `() => Promise<R | undefined>`, like you have, and builds the three `Event<ActivityTreeItemOptions>` based on the result of calls to that promise. Consumers could use this class but don't have to. This would be an easy way for consumers to use the API; they give us the thing to run and we handle the rest.
4. I like the `CreateResourceActivity` too, it can be a concrete child class of `ActivityBase`.

Issue #156, opened 2022-04-07
Unified: Error when expanding on Azure subscription


jinglouMSFT:
OS: Windows 10
VS Code:  Insider v1.67.0 & stable v1.66.0
Resource Group extension from: #20220405.3 Add display names and icons (#148)
Azure account extension: v0.10.0

Note:  I previously had the old release of the Resource Group extension installed, as well as the static web app extension installed. I uninstalled both. Then perform the following steps.

Repro:
1. In VS Code, F1 to bring up command prompt and search for Extension: install from VSIX.  Select the VSIX downloaded from the above build
2. Since my last Azure login session was from a long time ago, the resource extension couldn't be activated using the old authentication. I had to reauthenticate to sign into Azure which was successful, and my subscription was shown in the resource group
3. Try to expand the subscription node

Expected Results:
Show my Azure resources, or no error if I don't have any Azure resources

Actual Result:
"Error: Cannot read properties of undefined (reading 'id')

![image](https://user-images.githubusercontent.com/10081483/162097764-a4c1ed3b-d5d3-4a1b-93a0-95e32eb73368.png)

Account extension configuration:
![image](https://user-images.githubusercontent.com/10081483/162098052-5423c2c9-cdad-4495-87f9-ad94efdddf79.png)

Also FYI, this happened to all subscriptions I know I have access to or I am a owner of.

alexweininger:
Caused by the `groupBy` setting not respecting the default value somehow.

Repro steps:
1. Install extension from VSIX
2. Go to your settings.json file, and delete the `groupBy` setting
3. Reload VS Code or refresh the tree

We have a default value specified in the package.json, not sure why VS Code gives us undefined instead of this value...

nturinski:
Fixed by https://github.com/microsoft/vscode-azureresourcegroups/pull/159

PR #157, opened 2022-04-07
Add container apps to list of azureExtensions

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/82

PR #159, opened 2022-04-11
Group by ARM tag by selecting key

nturinski:
image

When selecting Group By ARM Tag, we allow the user to select a key.  That key is used to create groups by the value.  So, if the key is "defaultExperience", then the values would be Azure Table, Core SQL, etc. These values are coming from resources that have that key:value pair as an Azure tag.

Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/156 with line https://github.com/microsoft/vscode-azureresourcegroups/pull/159/files#diff-3b26e2c6749863538fbe756b9eb0e37aaa43b559e370bd6daf7aaa84a871a158R91

This was reproducing because we were only looking specifically at the globalValue which is not set until the value is in the workspace settings.  For a fresh install, we rely on the defaultValue.  By using `getWorkspaceSetting` we will look at the workspace setting, the global setting, and then the default setting in that order.


nturinski:
I added a change to have an "ungrouped" group if we can't find a group for the resource depending on the groupBy setting.  This is mostly relevant for "ARM tags," but would be useful for other configs in the future.

The main reason I wanted to do this is because I didn't want "groupBy" to be conflated with "filter" since that's effectively what it was doing.  We can add a "filter" to remove the ungrouped group if the user doesn't want all the clutter.

src/utils/azureUtils.ts
```diff
@@ -65,6 +76,18 @@ export function getIconPath(type?: string, kind?: string): TreeItemIconPath {
     return treeUtils.getIconPath(iconName);
 }
 
+export async function getArmTagKeys(context: IActionContext): Promise<Set<string>> {
+    const armTagKeys: Set<string> = new Set();
+    for (const sub of (await ext.rootAccountTreeItem.getCachedChildren(context))) {
```
nturinski, line 81: This can take a while if the user has multiple subscriptions opened, but not sure what we can do about that.

src/azureExtensions.ts
```diff
@@ -88,6 +88,14 @@ export const azureExtensions: IAzExtMetadata[] = [
...
         reportIssueCommandId: 'azureDatabases.reportIssue'
+    },
+    {
+        name: 'vscode-azurecontainerapps',
+        label: 'Container Apps',
+        resourceTypes: [
+            'microsoft.app/containerapps'
+        ],
+        reportIssueCommandId: 'containerApps.reportIssue'
     }
```
alexweininger, line 99: This was included in https://github.com/microsoft/vscode-azureresourcegroups/pull/157, do you need to merge origin/unified?

alexweininger:
Maybe use the tag codicon for the tag groups 
image

And then use 
image
for the ungrouped resources?

src/commands/explorer/groupBy.ts
```diff
@@ -5,13 +5,23 @@
...
     return (context: IActionContext): Promise<void> => groupBy(context, setting);
 }
 
 async function groupBy(context: IActionContext, setting: string): Promise<void> {
+    if (setting === 'armTag') {
+        const armTagKeys = await getArmTagKeys(context);
+        const tag = await context.ui.showQuickPick(Array.from(armTagKeys).map(key => { return { label: key } }), {
+            placeHolder: localize('groupByArmTagKey', 'Select the tag key to group by...')
+        });
+        setting += `-${tag.label}`;
```
alexweininger, line 22: Appending it to the groupBy setting string like this makes me a bit uneasy but I understand why we need to do it this way so it's ok.

src/utils/azureUtils.ts
alexweininger: Yeah I think it's unavoidable

alexweininger (approved):

src/commands/explorer/groupBy.ts
nturinski: Yeah, I was originally messing with groupBy setting that is only `armTag` and then other setting that was the key that is being used.  However, that created a lot of if/else clauses to special case when the groupBy setting was armTag which kind of defeated the simplicity of the group configuration code.

src/utils/azureUtils.ts
bwateratmsft: I don't know if this is even possible with an input/quickpick/whichever, but could we async load the keys, and while they're waiting let them paste or type something if they already know what they want?

bwateratmsft (approved):

nturinski:
Well, at least I can make the loading bar come up while it's loading the keys.  Visually, it's a lot better for the users to know that they're loading.

PR #163, opened 2022-04-12
Fix resource groups

alexweininger:
Fetches ResourceGroup data using the SDK to enable all the commands on resource group group tree items.

image

The implementation isn't the prettiest, but I made it as fast as we can until we add a something like a global resource cache. As of now, we fetch a list of all resource groups when the subscription is refreshed. Since I never await the promise, it seems to have a very minimal (if any), impact on the tree performance.

### DIFF
--- a/src/tree/AppResourceTreeItem.ts
+++ b/src/tree/AppResourceTreeItem.ts
@@ -95,11 +95,11 @@ export class AppResourceTreeItem extends ResolvableTreeItemBase implements Group
         ext.tagFS.fireSoon({ type: FileChangeType.Changed, item: this });
     }
 
-    public async mapSubGroupConfigTree(context: IActionContext, groupBySetting: string, resourceGroups: Promise<ResourceGroup[]>): Promise<void> {
+    public mapSubGroupConfigTree(context: IActionContext, groupBySetting: string, resourceGroups: Promise<ResourceGroup[]>): void {
         // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
         let subGroupTreeItem = (<SubscriptionTreeItem>this.rootGroupTreeItem).getSubConfigGroupTreeItem(this.groupConfig[groupBySetting].id)
         if (!subGroupTreeItem) {
-            subGroupTreeItem = await this.createSubGroupTreeItem(context, groupBySetting, resourceGroups);
+            subGroupTreeItem = this.createSubGroupTreeItem(context, groupBySetting, resourceGroups);
             (<SubscriptionTreeItem>this.rootGroupTreeItem).setSubConfigGroupTreeItem(this.groupConfig[groupBySetting].id, subGroupTreeItem)
         }
 
@@ -109,7 +109,7 @@ export class AppResourceTreeItem extends ResolvableTreeItemBase implements Group
         void subGroupTreeItem.refresh(context);
     }
 
-    public async createSubGroupTreeItem(_context: IActionContext, groupBySetting: string, resourceGroups: Promise<ResourceGroup[]>): Promise<GroupTreeItemBase> {
+    public createSubGroupTreeItem(_context: IActionContext, groupBySetting: string, resourceGroups: Promise<ResourceGroup[]>): GroupTreeItemBase {
         switch (groupBySetting) {
             // TODO: Use ResovableTreeItem here
             case 'resourceGroup':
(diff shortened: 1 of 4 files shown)


src/tree/AppResourceTreeItem.ts
```diff
@@ -107,12 +109,11 @@ export class AppResourceTreeItem extends ResolvableTreeItemBase implements Group
         void subGroupTreeItem.refresh(context);
     }
 
-    public createSubGroupTreeItem(_context: IActionContext, groupBySetting: string): GroupTreeItemBase {
-        // const client = await createResourceClient([context, this.rootGroupTreeItem.subscription]);
-
+    public async createSubGroupTreeItem(_context: IActionContext, groupBySetting: string, resourceGroups: Promise<ResourceGroup[]>): Promise<GroupTreeItemBase> {
```
alexweininger, line 112: Doesn't need to be async anymore

alexweininger, line 112: Doesn't need to be async anymore

src/tree/ResourceGroupTreeItem.ts
```diff
@@ -16,9 +16,19 @@ import { GroupTreeItemBase } from "./GroupTreeItemBase";
 export class ResourceGroupTreeItem extends GroupTreeItemBase {
     public static contextValue: string = 'azureResourceGroup';
 
-    constructor(parent: AzExtParentTreeItem, config: GroupNodeConfiguration, public data: ResourceGroup) {
+    public data?: ResourceGroup;
+
+    constructor(parent: AzExtParentTreeItem, config: GroupNodeConfiguration, data?: ResourceGroup | Promise<ResourceGroup[]>) {
```
nturinski, line 21: Discussed offline, but should pass in a `dataTask` into constructor that is always a `Promise<ResourceGroup[]>`.  Might be worth to just make an async constructor as well so we don't run into weird issues if people are trying to mess with the ResourceGroupTreeItem before it resolves the data.  Or make an async `getData` command that always gets called before anything that relies on data being there.

src/tree/ResourceGroupTreeItem.ts
```diff
@@ -16,18 +16,18 @@ import { GroupTreeItemBase } from "./GroupTreeItemBase";
 export class ResourceGroupTreeItem extends GroupTreeItemBase {
     public static contextValue: string = 'azureResourceGroup';
 
-    public data?: ResourceGroup;
+    public async getData(): Promise<ResourceGroup | undefined> {
+        return await this.getResourceGroup(this.name);
+    }
+
+    private data?: ResourceGroup;
```
bwateratmsft, line 23: This is being set but I don't see it being read anywhere. Were we meaning to use it as a cache so that `getData` only gets called once? If so we need a check+return in there.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -102,8 +104,15 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         this._treeMap = {};
         const groupBySetting = <string>settingUtils.getGlobalSetting<string>('groupBy');
 
-        for (const rgTree of this._items) {
-            (<AppResourceTreeItem>rgTree).mapSubGroupConfigTree(context, groupBySetting);
+        const client: ResourceManagementClient = await createResourceClient([context, this]);
+        const resourceGroups = uiUtils.listAllIterator(client.resourceGroups.list());
```
bwateratmsft, line 108: I'm only about 30% sure, but I think this is preemptively calling `client.resourceGroups.list()` (before the `await resourceGroups` below), and that this would make it not do that:
```suggestion
        const resourceGroups = uiUtils.listAllIterator(() => client.resourceGroups.list());
```

bwateratmsft (approved):

src/tree/ResourceGroupTreeItem.ts
alexweininger: It's being read in a `refreshImpl` and `get id()`

bwateratmsft: It's `private`, so it shouldn't be. :smile:

src/tree/SubscriptionTreeItem.ts
alexweininger: Wouldn't `uiUtils.listAllIterator(() => client.resourceGroups.list());` result in `client.resourceGroups.list()` being called every time the `resourceGroups` Promise is awaited?

Also I think we are ok with it calling `client.resourceGroups.list()` before the await, since we don't mind when it starts/finishes.

bwateratmsft: If we're starting it no matter what, then it's better as-written, because it will start the `client.resourceGroups.list()` immediately, rather than waiting until `await resourceGroups`.

PR #178, opened 2022-04-15
Use activity api for delete and create resource group

alexweininger:
Activity panel stress test 😄

https://user-images.githubusercontent.com/12476526/163603496-889fecaa-884c-4356-a28b-1807880eb931.mov

### DIFF
--- a/src/commands/deleteResourceGroup/deleteResourceGroup.ts
+++ b/src/commands/deleteResourceGroup/deleteResourceGroup.ts
@@ -4,10 +4,10 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { IActionContext, UserCancelledError } from '@microsoft/vscode-azext-utils';
-import { ext } from '../extensionVariables';
-import { ResourceGroupTreeItem } from '../tree/ResourceGroupTreeItem';
-import { localize } from '../utils/localize';
-import { settingUtils } from '../utils/settingUtils';
+import { ext } from '../../extensionVariables';
+import { ResourceGroupTreeItem } from '../../tree/ResourceGroupTreeItem';
+import { localize } from '../../utils/localize';
+import { settingUtils } from '../../utils/settingUtils';
 
 export async function deleteResourceGroup(context: IActionContext, primaryNode?: ResourceGroupTreeItem, selectedNodes?: ResourceGroupTreeItem[]): Promise<void> {
     if (!selectedNodes) {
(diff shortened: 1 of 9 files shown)


src/commands/deleteResourceGroup/deleteResourceGroup.ts
```diff
@@ -45,9 +44,6 @@ export async function deleteResourceGroup(context: IActionContext, primaryNode?:
         }
 
         void node.deleteTreeItem(context);
-        const message: string = localize('startedDelete', 'Started delete of resource group "{0}".', node.name);
-        void window.showInformationMessage(message);
-        ext.outputChannel.appendLog(message);
```
alexweininger, line 50: Removed this extra notification, and then moved the appendLog call into `DeleteResourceGroupStep`

src/tree/ResourceGroupTreeItem.ts
```diff
@@ -75,9 +77,17 @@ export class ResourceGroupTreeItem extends GroupTreeItemBase {
     }
 
     public async deleteTreeItemImpl(context: IActionContext): Promise<void> {
-        const client: ResourceManagementClient = await createResourceClient([context, this]);
-        await client.resourceGroups.beginDeleteAndWait(this.name);
-        ext.outputChannel.appendLog(localize('deletedRg', 'Successfully deleted resource group "{0}".', this.name));
+        const wizard = new AzureWizard<DeleteResourceGroupContext>({
```
bwateratmsft, line 80: It sorta seems like overkill to use the full wizard framework to do the delete. Is the experience significantly different if we just use a simple `ExecuteActivity` to wrap the create RG client+delete RG?

alexweininger: It may be overkill, but using a full AzureWizard automatically handles showing the progress, and takes into account the setting we added to suppress the progress notifications.

I could write a utility function to make creating azure wizards easier.

bwateratmsft: In a multi-step process I think it definitely makes sense to use the wizard and take advantage of its progress reporting, but this is just a single-step delete.

alexweininger: Ok, I overlooked this originally so I'll refactor `ExecuteActivity` a bit and also expose it in utils so we can use it outside of the AzureWizard.

alexweininger: Talked offline and decided it's ok to use a AzureWizard in this case

PR #182, opened 2022-04-18
Improve performance of changing groupBy setting

alexweininger:
~~Depends on https://github.com/microsoft/vscode-azuretools/pull/1122~~

Now if the refresh is triggered by a view setting change, we cache fetched azure resources. Any other refresh refetched the resources. This results in much smoother experience when changing the groupBy setting.

https://user-images.githubusercontent.com/12476526/163850585-2a4c60f6-811a-4a11-83eb-ace71976a8fb.mov

### DIFF
--- a/src/tree/GroupTreeItemBase.ts
+++ b/src/tree/GroupTreeItemBase.ts
@@ -34,8 +34,8 @@ export class GroupTreeItemBase extends AzExtParentTreeItem {
     }
 
     public get contextValue(): string {
-        const contextValues = [...this.config.contextValuesToAdd ?? [], ...this.internalContextValuesToAdd, 'group'];
-        return contextValues.sort().join(';');
+        const contextValues = new Set([...this.config.contextValuesToAdd ?? [], ...this.internalContextValuesToAdd, 'group']);
+        return Array.from(contextValues).sort().join(';');
     }
 
     public get description(): string | undefined {
--- a/src/tree/GroupTreeItemBase.ts
+++ b/src/tree/GroupTreeItemBase.ts
@@ -15,6 +15,8 @@ export class GroupTreeItemBase extends AzExtParentTreeItem {
     public treeMap: { [key: string]: ResolvableTreeItemBase } = {};
     public config: GroupNodeConfiguration;
 
+    protected internalContextValuesToAdd: string[] = []
+
     public readonly cTime: number = Date.now();
     public mTime: number = Date.now();
 
@@ -32,7 +34,8 @@ export class GroupTreeItemBase extends AzExtParentTreeItem {
     }
 
     public get contextValue(): string {
-        return this.config.contextValue || 'GroupTreeItem';
+        const contextValues = [...this.config.contextValuesToAdd ?? [], ...this.internalContextValuesToAdd, 'group'];
+        return contextValues.sort().join(';');
     }
 
     public get description(): string | undefined {
--- a/src/tree/ResourceGroupTreeItem.ts
+++ b/src/tree/ResourceGroupTreeItem.ts
@@ -4,8 +4,8 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { ResourceGroup } from "@azure/arm-resources";
-import { AzExtParentTreeItem, AzExtTreeItem, AzureWizard, IActionContext, nonNullProp, TreeItemIconPath } from "@microsoft/vscode-azext-utils";
-import { FileChangeType, TreeItemCollapsibleState } from "vscode";
+import { AzExtParentTreeItem, AzureWizard, IActionContext, nonNullProp, TreeItemIconPath } from "@microsoft/vscode-azext-utils";
+import { FileChangeType } from "vscode";
 import { GroupNodeConfiguration } from "../api";
 import { DeleteResourceGroupContext } from "../commands/deleteResourceGroup/DeleteResourceGroupContext";
 import { DeleteResourceGroupStep } from "../commands/deleteResourceGroup/DeleteResourceGroupStep";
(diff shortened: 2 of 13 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -21,46 +21,67 @@ import { GroupTreeItemBase } from './GroupTreeItemBase';
...
+        appResources: AppResource[];
+        groupableResources: GroupableResource[];
+    }
     private _treeMap: { [key: string]: GroupTreeItemBase } = {};
 
-    private rgsItem: AppResource[] = [];
+    private resetCache(): void {
+        this.cache = {
+            resourceGroups: [],
+            appResources: [],
```
bwateratmsft, line 46: `AppResource` and `GroupableResource` are meant to overlap; i.e. the relevant object implements both of them.

src/tree/ResourceGroupTreeItem.ts
```diff
@@ -90,15 +88,13 @@ export class ResourceGroupTreeItem extends GroupTreeItemBase {
         await wizard.execute();
     }
 
-    public compareChildrenImpl(item1: AzExtTreeItem, item2: AzExtTreeItem): number {
-        if ((item1 as AzExtParentTreeItem).loadMoreChildrenImpl && (item2 as AzExtParentTreeItem).loadMoreChildrenImpl) {
+    public compareChildrenImpl(item1: AzExtTreeItem & { collapsibleState: TreeItemCollapsibleState }, item2: AzExtTreeItem & { collapsibleState: TreeItemCollapsibleState }): number {
```
nturinski, line 91: This is kind of odd behavior to me.  I know that this is previous behavior, but I'm not really sure what the motivation behind it was.

alexweininger: Odd behavior to sort the nodes with children to the top?

nturinski: Yeah, rather than just alphabetical.  I mean, I guess it's pretty visually easy to tell that it's happening, but it doesn't intuitively make sense to me.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -21,46 +21,65 @@ import { GroupTreeItemBase } from './GroupTreeItemBase';
...
     public hasMoreChildrenImpl(): boolean {
-        return !!this._nextLink;
+        return !!this.cache.nextLink;
     }
 
     public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
-        if (clearCache) {
-            this._nextLink = undefined;
-            this.rgsItem = [];
+        if (clearCache && !this._keepCache) {
```
nturinski, line 62: I'm a little uncomfortable having two separate flags for this.  `clearCache` is handled internally by `AzExtParentTreeItem`.  I think you should be able to utilize `clearCache` in `loadMoreChildrenImpl` to do everything that you need.

If you really want to keep your own flag though, make sure you always put the state changes in a `try/finally`.  If there's ever an error (for example, if `this.createTreeMaps` ended with an error) we'd never change the state to false.

We could also consider exposing `treeProvider.refreshUIOnly` command as well.

bwateratmsft: > We could also consider exposing `treeProvider.refreshUIOnly` command as well.

I think that's a good idea. There's a lot of use cases for it.

src/tree/ResourceGroupTreeItem.ts
alexweininger: I'm not sure which one is better, what do you think @fiveisprime?

Tree items with children sorted on top:
image

Purely alphabetical:
image

bwateratmsft: My vote would be alphabetical. There is definitely precedent for have-children-first approach (file explorers typically show all folders in alphabetical order, then all files in alphabetical order), but at a high level I don't think of these as "folders" just because they have children. In my mind they are still resources, same as the ones that _don't_ have children.

nturinski: Yeah, I think the way that Brandon explained it is right on the money.  That's exactly why it felt off to me.

alexweininger: Sounds like a plan! :shipit:

src/tree/SubscriptionTreeItem.ts
alexweininger: I removed the additional flag, but at the cost of not caching as much as before. The clearCache flag is not aggressive enough (or at least we need to expose a way for us to easily call refresh without setting clearCache to true. 

I don't think `treeProvider.refreshUIOnly` will help in this case, because calling it does not result in `loadMoreChildrenImpl` getting called.

src/tree/ResourceGroupTreeItem.ts
fiveisprime: Agreed on alpha for now. I'm interested in experimenting with this further though for sorting by type so that the most commonly used services show up first. I think we can make it more helpful and consistent so that, for example, I always see function apps first and I can expect that.

nturinski (approved):

src/tree/GroupTreeItemBase.ts
```diff
@@ -32,7 +34,8 @@ export class GroupTreeItemBase extends AzExtParentTreeItem {
     }
 
     public get contextValue(): string {
-        return this.config.contextValue || 'GroupTreeItem';
+        const contextValues = new Set([...this.config.contextValuesToAdd ?? [], ...this.internalContextValuesToAdd, 'group']);
```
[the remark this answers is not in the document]
alexweininger: It makes it easier to implement focus group command since we know `group` will always be there. Instead of relying on having the config context value follow the `*Group` pattern.

PR #185, opened 2022-04-19
Change the resolving behavior

bwateratmsft:
This change makes it so that rather than having a `NoopResolver`, instead a wrapper promise is built around waiting for the extension to activate and then resolving through its resolver. This way, the delay caused by activation is transparent; calling `resolve` will always get to the final result.

Tested successfully. I tested by intentionally slowing down SWA's activation. What happens is the loading spinner persists until SWA finishes activating, registering the resolver, and resolving for the node(s).

If a user expands and then collapses a group before the extension has finished activating, the nodes that were revealed and then hidden will _still_ be resolved (which means a little wasted work), but I don't think that is a big problem.


bwateratmsft:
In theory this removes the need for https://github.com/microsoft/vscode-azuretools/pull/1122.

src/extension.ts
```diff
@@ -94,3 +94,13 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
 export function deactivateInternal(): void {
     ext.diagnosticWatcher?.dispose();
 }
+
+function setupEvents(context: vscode.ExtensionContext): void {
+    // Event emitter for when a group is focused/unfocused
```
bwateratmsft, line 99: @alexweininger I moved these down here, I could see this list getting long and having it in the main body of the activate function would clutter it a bit.

bwateratmsft: Also I followed the `ext.emitters` / `ext.events` pattern you established.

PR #187, opened 2022-04-20
Implement the new RG API

bwateratmsft:
This implements the new/slightly different RG API defined in https://github.com/microsoft/vscode-azuretools/pull/1126. Strictly speaking it wasn't necessary to change `ext.tree` -> `ext.appResourceTree`, but I thought it might save from confusion down the road.


bwateratmsft:
Whoops, wrong target branch

PR #197, opened 2022-04-21
Handle outdated extensions

alexweininger:
I was going to do something fancier with semver, but I ended up just checking for the existence of the `x-azResources` key in the extension manifest. IMO this is way simpler and more fool-proof, and we can easily update the `meetsMinVersion()` function later if needed.

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azureresourcegroups",
-    "version": "0.4.1-alpha.1",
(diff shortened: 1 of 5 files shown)


bwateratmsft (approved):
I think this is a good idea. It's probably more common than we'd like due to the recent issues requiring rollback.

It doesn't have to be part of this PR, but it might be a good idea to capture in telemetry when a builtin resolver gets used, and which (and for which resource type + kind).

Issue #191, opened 2022-04-21
Fail to execute command "Azure Resource Groups: Focus this Group" with an error from command palette


v-mengwe:
**OS:** All
**Build Version:** 20220420.34
**Regression:** Not a regression

**Repro Steps:**

1. F1 -> Azure Resource Groups: Focus this Group.
2. Check whether succeeds to execute command "Azure Resource Groups: Focus this Group".

**Expect:**
Succeed to execute command "Azure Resource Groups: Focus this Group".

**Actual:**
Fail to execute command "Azure Resource Groups: Focus this Group" with an error.
<img width="446" alt="Screen Shot 2022-04-21 at 15 08 56" src="https://user-images.githubusercontent.com/62053665/164397795-4f152a5f-8aae-4615-9788-4222925eedfb.png">

alexweininger:
Fixed by #200

PR #200, opened 2022-04-21
Fix focus this group command when used from palette

alexweininger:
Fix focus this group command when used from palette #191
Also fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/667

https://user-images.githubusercontent.com/12476526/164555132-e86d8dd5-c00b-4fa8-be88-755faa829159.mov

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -30,8 +30,6 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         return this.cache.resourceGroups;
     }
 
-    public readonly childTypeLabel: string = localize('resourceGroup', 'Resource Group');
-
     private cache: {
         resourceGroups: ResourceGroup[];
         nextLink?: string;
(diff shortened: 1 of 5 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -30,8 +30,6 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         return this.cache.resourceGroups;
     }
 
-    public readonly childTypeLabel: string = localize('resourceGroup', 'Resource Group');
```
nturinski, line 33: The .gif still shows "Resource Group" when you select it.  Was that screen grab made before this change?  I can't remember what it says by default.

nturinski (approved):

src/tree/SubscriptionTreeItem.ts
alexweininger: I fixed it after I noticed it in the gif 😄

PR #201, opened 2022-04-21
Tree mapping cache for all default groupings

nturinski:
We're limited to only relying on cache on the default three groupings.  Since the user can select which arm tag they would like to group by, there's no way to pre-emptively do that.

This should also help with any tree pickers problems where we need it to be in "resourceGroup" grouping.  We can use the `pickTreeItemImpl` to return a tree item that contains all the resource groupings.

The flag that gets set when the user changes the groupBy setting should _always_ be reset to false whenever the children are loaded.  We want to make sure that never ends up getting stuck in true.  This makes Alex's cache more persistent though so that we only refresh/reload resources when they actually hit "refresh".

I considered using a similar pattern for "focusOn" to make it a little bit snappier.  It's fast and snappy on my machine, but if you had a slow internet, I could see that being a bit sluggish.

cacheload


src/commands/explorer/groupBy.ts
```diff
@@ -31,6 +31,5 @@ async function getQuickPicks(context: IActionContext): Promise<QuickPickItem[]>
 export enum GroupBySettings {
     ResourceGroup = 'resourceGroup',
     ResourceType = 'resourceType',
-    Location = 'location',
-    ArmTag = 'armTag'
```
alexweininger, line 35: arm tag was removed because we don't want to generate the groups ahead of time?

src/tree/SubscriptionTreeItem.ts
```diff
@@ -113,62 +117,74 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
         registerEvent('treeView.onDidChangeConfiguration', workspace.onDidChangeConfiguration, async (context: IActionContext, e: ConfigurationChangeEvent) => {
             context.errorHandling.suppressDisplay = true;
             context.telemetry.suppressIfSuccessful = true;
             context.telemetry.properties.isActivationEvent = 'true';
 
-            if (e.affectsConfiguration(`${ext.prefix}.${key}`)) {
+            if (e.affectsConfiguration(`${ext.prefix}.groupBy`)) {
+                // we can generate the default groups ahead of time so we don't need to refresh the entire tree
+                this._triggeredByDefaultSetting = Object.values(GroupBySettings).includes(settingUtils.getWorkspaceSetting<string>('groupBy') as GroupBySettings);
                 await this.refresh(context);
```
alexweininger, line 136: Should we also wrap this in a try/finally to make sure we reset `this._triggeredByDefaultSetting`? Especially if we ever implement `refreshImpl`.

alexweininger (approved):

src/commands/explorer/groupBy.ts
nturinski: Yeah it being as a "default" didn't really make sense considering how we form the tag.

src/tree/SubscriptionTreeItem.ts
nturinski: Well, the thing is, even if you await the refresh, it still technically resolves before loadMoreChildren is actually called apparently.  I originally did try/catch the setting here, but it kept getting set before loadMoreChildren would actually start

Issue #202, opened 2022-04-22
There is no response after clicking "Click to view resource" in "Activity Log"


v-ruizh:
**OS:** All
**Azure Resource Extension Build Version:** 20220421.32 (unified)
**Azure Function Extension build Version:** 20220421.8 (unified)
**Regression:** Previous Build

**Repro Steps:**
1. Make sure the Azure Resources group by the **Resource Group**.
2. Create a Function App.
3. Click "Click to view resource" in "Activity Log".
4. Check whether the created Function App is focused or not.

**Expect:**
The created Function App is focused.

**Actual:**
There is no response after clicking "Click to view resource".

**More Info:**
This issue does not reproduce on the previous unified Azure Resource extension build 20220420.34.

alexweininger:
@nturinski Did something change with how app resource tree item ids are created? The reveal isn't working at all anymore

v-ruizh:
Verified this issue on the extension 20220524.1, it has been fixed.

PR #205, opened 2022-04-25
Fix reveal resource

nturinski:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/202

This fix relies on the fact that all of the `AppResourceTreeItem`s actually exist on the sub node.  Rather than using the resourceId to find the fullId (which we could do, but it'd be kind of cumbersome), we can just find the actual node with the resourceId and pass that to be revealed.

I kept the string parameter for `revealTreeItem` because being able to reveal by id could be useful for other extensions since they won't have access to the actual node like the RG extension does.

The other fix that this [PR](https://github.com/microsoft/vscode-azureresourcegroups/pull/204) extends off of fixes `findTreeItem`, so the command palette should be working again as well.


alexweininger:
I got this error right after switching my groupBy setting to location. Reveal worked the very next time I clicked it so I'm guessing something wasn't loaded yet.

image

nturinski:
Good catch, Alex.  Turns out that this is happening because the children weren't actually loaded yet (even though the mapping exists, the group nodes don't technically load until you expand them).  I only tested it after I had expanded the group node, so never saw this occur.

Made me think of another case-- when the sub node hasn't been expanded, the `AppResourceTreeItem` mapping hasn't been completed yet either.  So just load the children in both cases.

alexweininger (approved):
Works like a charm 💯  🚀

PR #206, opened 2022-04-27
Filter to only display resources supported by our extensions

nturinski:
image
image
image


alexweininger (commented):
Did you forget to add the command to the package.json?

nturinski:
Well, I don't think it should ever appear in the command palette, but I did forget to add it to the when === 'never' list.

PR #209, opened 2022-04-28
Ensure resource is resolved when picking children

alexweininger:
Needed to completely fix https://github.com/microsoft/vscode-azurefunctions/issues/3135

### DIFF
--- a/src/tree/ResolvableTreeItemBase.ts
+++ b/src/tree/ResolvableTreeItemBase.ts
@@ -63,6 +63,13 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
         });
     }
 
+    public async pickTreeItemImpl(expectedContextValues: (string | RegExp)[], context: IActionContext): Promise<AzExtTreeItem | undefined> {
+        if (!this.resolveResult) {
+            await this.resolve(true, context);
+        }
+        return super.pickTreeItemImpl?.(expectedContextValues, context);
+    }
+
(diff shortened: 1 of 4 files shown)


src/tree/ResolvableTreeItemBase.ts
```diff
@@ -63,6 +63,13 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
         });
     }
 
+    public async pickTreeItemImpl(expectedContextValues: (string | RegExp)[], context: IActionContext): Promise<AzExtTreeItem | undefined> {
+        if (!this.resolveResult) {
+            await this.resolve(true, context);
+        }
+        return super.pickTreeItemImpl?.(expectedContextValues, context);
+    }
```
alexweininger, line 71: Should we be resolving these in a `pickTreeItemImpl` in `GroupTreeItemBase`?

nturinski: Yeah, I think that makes more sense.

We can either have a loading bar while the children resolve or just display the list of current children and have them resolve in the background, and then await it if it hasn't finished resolving by the time the user selects it.

nturinski (approved):

alexweininger:
Unsure if this is still needed since we went a different direction #212, will leave it up for now.

nturinski:
It shouldn't be since the app resource is being resolved after it's picked now in #212

PR #212, opened 2022-04-29
Implement pickAppResource

alexweininger:
This is a big step forward in terms of helping us solve all of the issues surrounding running commands from the palette. The design of showTreeItemPicker doesn't fit well into the new model, and so this is just a slight variation that skips a step.

Depends on https://github.com/microsoft/vscode-azuretools/pull/1141

There are a few options we can take advantage of here:

```ts
export interface PickAppResourceOptions extends IAzureQuickPickOptions {
    type?: string; // ex: 'microsoft.web/sites'
    expectedChildContextValue?: string | RegExp | (string | RegExp)[]; // ex: EnvironmentTreeItem.contextValue
}
```

If you pass in `type`, the picks are filtered based on that type. Passing a value for `expectedChildContextValue` enables us to select children of app resources, ex: an app setting tree item.

Example of selecting an app resource:

https://user-images.githubusercontent.com/12476526/165870763-9de237f8-114a-4832-9701-52fc8b693e20.mov

### DIFF
--- a/src/api/pickAppResource.ts
+++ b/src/api/pickAppResource.ts
@@ -3,16 +3,19 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { AzExtTreeItem, IActionContext } from "@microsoft/vscode-azext-utils";
+import { AzExtTreeItem, ITreeItemPickerContext } from "@microsoft/vscode-azext-utils";
 import { PickAppResourceOptions } from "@microsoft/vscode-azext-utils/hostapi";
 import { ext } from "../extensionVariables";
 import { SubscriptionTreeItem } from "../tree/SubscriptionTreeItem";
 
-export async function pickAppResource<T extends AzExtTreeItem>(context: IActionContext, options?: PickAppResourceOptions): Promise<T> {
+export async function pickAppResource<T extends AzExtTreeItem>(context: ITreeItemPickerContext, options?: PickAppResourceOptions): Promise<T> {
     const subscription: SubscriptionTreeItem = await ext.appResourceTree.showTreeItemPicker(SubscriptionTreeItem.contextValue, context);
     const appResource = await subscription.pickAppResource(context, options);
 
     if (options?.expectedChildContextValue) {
+        if (!appResource.resolveResult) {
+            await appResource.resolve(true, context);
+        }
         return ext.appResourceTree.showTreeItemPicker(options.expectedChildContextValue, context, appResource);
     } else {
         return appResource as unknown as T;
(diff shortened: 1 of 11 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -38,6 +38,26 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         }
     }
 
+    public async pickAppResource(context: IActionContext, options?: PickAppResourceOptions): Promise<AppResourceTreeItem> {
+        await this.getCachedChildren(context);
+        let appResources = GroupTreeItemBase.filterResources(this.cache.appResources);
```
nturinski, line 43: If the setting to hide all resources is false, you shouldn't filter the appResources.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -38,6 +38,26 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+
+        if (options?.type) {
```
nturinski, line 45: I forget, do we actually change the type property on the app resource? Or do we need to check kind to distinguish between function and web apps?

alexweininger: I could change the options to this, then we can use the utils we already have setup to handle this.

```ts
export interface PickAppResourceOptions extends IAzureQuickPickOptions {
    /**
     * Options to filter the picks
     */
    filter?: {
        /**
         * App resource type to filter the app resource picks
         */
        type: string;
        /**
         * App resource kind to filter the app resource picks
         */
        kind?: string;
    }

    /**
     * Set this to pick a child of the selected app resource
     */
    expectedChildContextValue?: string | RegExp | (string | RegExp)[];
}
```

alexweininger: Or I was just going to pass in `microsoft.web/functionapp` for functions since it's a constant and that's what we convert function apps to anyway in resource groups using the utils in azureUtils.ts.

nturinski (approved):

src/api/pickAppResource.ts
```diff
@@ -0,0 +1,24 @@
...
+import { PickAppResourceOptions } from "@microsoft/vscode-azext-utils/hostapi";
+import { ext } from "../extensionVariables";
+import { SubscriptionTreeItem } from "../tree/SubscriptionTreeItem";
+
+export async function pickAppResource<T extends AzExtTreeItem>(context: ITreeItemPickerContext, options?: PickAppResourceOptions): Promise<T> {
+    const subscription: SubscriptionTreeItem = await ext.appResourceTree.showTreeItemPicker(SubscriptionTreeItem.contextValue, context);
+    const appResource = await subscription.pickAppResource(context, options);
+
+    if (!appResource.resolveResult) {
+        await appResource.resolve(true, context);
```
nturinski, line 16: Do we need to await the resolve here? Unless the options have `expectedChildContextValue`, we should be able to just return the `proxyTreeItem`.  

Haven't tested this, but could there be a noticeable delay if the resource isn't resolved?  Also, if the command we call is from Resource Groups (`openInPortal`) we'll have to wait for it to activate the extension and resolve as well.

I imagine that the extension will have to be activated since we're resolving the resource even if unneeded.  However, if it's that extensions registered command, it'll have to activate and should resolve all the resolvables.

alexweininger: Would you expect some commands to await the resolve after calling pickAppResource?

Issue #208, opened 2022-04-27
Odd loading behavior that blocks focus on group command


alexweininger:
Since the video doesn't show mouse clicks, I'll outline what I did in the video:

1. Clicked to expand the group node
2. Immediately after that, I clicked the "Focus this Group" button. 

Expected behavior:
Ideally the view would focus on the group as it's loading

Actual behavior:
The the view focuses on the group only after everything has loaded

https://user-images.githubusercontent.com/12476526/165635775-6aa7073a-e8e3-45df-a6e9-0e79160e3d97.mov

alexweininger:
I don't think this reproduces anymore. We'll re-open later if needed

Issue #213, opened 2022-05-02
Switching the group hangs when extension is in the middle of loading


wwlorey:
Repro steps:

1. Expand a tree item that requires a lengthy loading time

![Screen Shot 2022-05-02 at 1 19 57 PM](https://user-images.githubusercontent.com/22795803/166321384-e7c9bfa6-6a0c-4670-bc02-da5f5daf7d55.png)

2. Switch the grouping (in this case I chose "location")

**Expected**: Selecting a new grouping halts the loading and immediately switches to the new view.
**Actual**: The async call is fully awaited and the view switches after it completes (in this case it was 10 or 15sec).

nturinski:
We might consider having a cancellation token for loading children so that certain commands can cancel this async call.

PR #216, opened 2022-05-03
Use setCache to get rid of messy loadMoreChildrenImpl logic

nturinski:
Relies on https://github.com/microsoft/vscode-azuretools/pull/1142

I was using a bunch of weird flags/state to effectively use our cache.  I really didn't like that implementation because it seems like it could be very error prone.  If you try to refresh and change grouping or focus, it was having some weird behaviors.

`loadMoreChildren` tends to get called a lot too so there was a good chance that two separate calls could interfere with each other.  Changing to the setCache method should be safer since all calls are synchronous and it is never automatically called by the VS Code TreeProvider API.

I considered using the same approach for `GroupTreeItemBase.ts` but it seemed a bit overkill to register an event for it since the loadMoreChildrenImpl is already synchronous anyway.


src/tree/SubscriptionTreeItem.ts
```diff
@@ -246,6 +246,14 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         return this.cache.treeMaps[groupBy] ?? {};
     }
 
+    private _setCachedChildren(childrenToSet: AzExtTreeItem[]): void {
+        // To access the private cacheChildren, go buckwild and ignore typings!!
+        const thisUnknown = this as unknown as { _cachedChildren: AzExtTreeItem[] };
```
bwateratmsft, line 251: image

It's not my fault, I have young children.

nturinski: I'm pretty sure that this is Frozen 2... Deep cut.

alexweininger: 🤣

alexweininger:
Have you verified this fixes #208 and #213 ?

package.json
```diff
@@ -476,7 +476,7 @@
         "@azure/arm-resources": "5.0.0",
         "@azure/arm-resources-profile-2020-09-01-hybrid": "^1.0.0",
         "@microsoft/vscode-azext-azureutils": "^0.2.0",
-        "@microsoft/vscode-azext-utils": "^0.2.3",
+        "@microsoft/vscode-azext-utils": "0.2.4",
```
bwateratmsft, line 479:
```suggestion
        "@microsoft/vscode-azext-utils": "^0.2.4",
```

src/tree/SubscriptionTreeItem.ts
```diff
@@ -92,48 +92,29 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+            // await Promise.all(applicationResourceProviders.map((provider: ApplicationResourceProvider) => async () => this.rgsItem.push(...(await provider.provideResources(this.subscription) ?? []))));
+
+            resources.forEach(item => ext.activationManager.onNodeTypeFetched(item.type));
+            this.cache.appResources = resources.map((resource: AppResource) => AppResourceTreeItem.Create(this, resource));
+        }
+        await this.createTreeMaps(clearCache, context);
+
+        const focusedGroupId = await ext.context.workspaceState.get('focusedGroup') as string;
+        // on first load, check if there was persistent setting
+        const focusGroupTreeItem = await this.tryGetFocusGroupTreeItem(focusedGroupId);
```
bwateratmsft, line 112: Can we have `tryGetFocusGroupTreeItem` internally take care of looking up the setting value?

nturinski:
Closing in favor of https://github.com/microsoft/vscode-azureresourcegroups/pull/222

PR #217, opened 2022-05-03
Put filtered resources below toggle

nturinski:
filterResources

Looks much nicer.  Shout out to my boi @wwlorey for the suggestion


alexweininger:
i like it

alexweininger (approved):

src/tree/GroupTreeItemBase.ts
```diff
@@ -124,14 +129,14 @@ export class GroupTreeItemBase extends AzExtParentTreeItem {
             localize('showingResources', 'Showing {0} of {1} resources. Click to reveal all resources.', numOfResources, numOfTotalResources) :
             localize('hideAllResources', 'Click to hide filtered resources.');
```
bwateratmsft, line 130: Not directly related to this PR but I noticed this when it was originally added. I don't _love_ this string, but I'm not sure what else works better. My thinking is that if the resource was filtered, it would _already_ be hidden, not in need of hiding.

I thought of "Click to hide unsupported resources" but "unsupported" seems like it might be alarming if misunderstood, so I don't like that either.

bwateratmsft (approved):

src/tree/GroupTreeItemBase.ts
nturinski: Yeah, I'm waiting for @fiveisprime to sprinkle some PMagic

image

Because I definitely agree.

alexweininger: We're feeling the media today aren't we

everywhere

PR #221, opened 2022-05-05
Handle filtering by tag

alexweininger:
Depends on https://github.com/microsoft/vscode-azuretools/pull/1146

> Needed to support some database scenarios where the resources can only be differentiated based on tag values.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
     "name": "vscode-azureresourcegroups",
     "displayName": "Azure Resources",
     "description": "%azureResourceGroups.description%",
-    "version": "0.4.1-alpha.11",
+    "version": "0.4.1-alpha.12",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -476,7 +476,7 @@
         "@azure/arm-resources": "5.0.0",
         "@azure/arm-resources-profile-2020-09-01-hybrid": "^1.0.0",
         "@microsoft/vscode-azext-azureutils": "^0.2.0",
-        "@microsoft/vscode-azext-utils": "^0.2.3",
+        "@microsoft/vscode-azext-utils": "file:../vscode-azuretools/utils/microsoft-vscode-azext-utils-0.2.4.tgz",
```
nturinski, line 479: Probably realized, but your npm package version is `0.2.5`

PR #222, opened 2022-05-05
Implement ResourceCache that handles Azure resource caching

nturinski:
The purpose of this PR is to abstract all of the caching logic from the `SubscriptionTreeItem` to a `ResourceCache` that each `SubscriptionTreeItem` will have.  The `ResourceCache` handles caching the resources and creating the `AppResourceTreeItems` that it then caches so we aren't re-resolving the resources on each refresh.

It compares the cached resources to the providedResources and adds/removes accordingly.  This greatly improves refresh time because we don't have to resolve every resource after a create or delete, just the added ones.

We no longer cache the treeMaps because they are a cheap operation.  The result is more simple caching logic since we don't have to update each cached tree each time a resource is added/removed.  It also makes it easier to work with `AzExtParentTreeItem._cachedChildren` because `deleteNode` and `createNode` both contribute to cached children, which will be reflected on the tree view immediately.

The groupBy is marginally slower now that we are making an API call each time we change the grouping, but as pointed out by Brandon, it's unlikely that people are going to change that setting very often.  Focus, which users are more likely to use frequently,  still relies on a cache, so it remains very snappy.


nturinski:
Side note: I am going to hunt down where we create all of these ids and lower case them at the source.  Because the number of times there are id-mismatches due to casing is killing me 😒

src/tree/SubscriptionTreeItem.ts
```diff
@@ -63,77 +53,38 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+        // To support multiple app resource providers, need to use this pattern
+        // await Promise.all(applicationResourceProviders.map((provider: ApplicationResourceProvider) => async () => this.rgsItem.push(...(await provider.provideResources(this.subscription) ?? []))));
+        resources.forEach(item => ext.activationManager.onNodeTypeFetched(item.type));
+        this.cache.appResources = resources;
 
-                    resources.forEach(item => ext.activationManager.onNodeTypeFetched(item.type));
-                    this.cache.appResources = resources.map((resource: AppResource) => AppResourceTreeItem.Create(this, resource));
-                }
-                await this.createTreeMaps(clearCache, context);
+        await this.resolveResourceGroups(context);
```
bwateratmsft, line 78: If the grouping is not resource groups, do we still need to resolve info about them?

src/tree/ResourceCache.ts
```diff
@@ -0,0 +1,120 @@
...
+        this._appResources = [];
+        this._resourceGroups = [];
+    }
+
+    public get appResources(): AppResourceTreeItem[] {
+        return this._appResources;
+    }
+
+    public set appResources(resources: AppResource[]) {
+        const addedResources = this.resolveProvidedResources(this._appResources, resources);
```
bwateratmsft, line 40: I don't feel strongly about it, but it might be good to make `resolveProvidedResources` something like `diffProvidedResources`, and have it return a tuple of added and removed. And then, handle both the adding *and* the removing here, rather than adding here, and removing within `resolveProvidedResources`.

src/tree/ResourceCache.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    public set resourceGroups(resourceGroups: ResourceGroup[]) {
+        this._resourceGroups.push(...this.resolveProvidedResources(this._resourceGroups, resourceGroups));
+    }
+
+    public getTreeMap(context: IActionContext): GroupTreeMap {
+        const treeMap: GroupTreeMap = {};
+        // start this as early as possible
+        const getResourceGroupTask: (resourceGroup: string) => Promise<ResourceGroup | undefined> = async (resourceGroup: string) => {
```
bwateratmsft, line 51: I don't think this will start it, just create the promise (referring to the comment just above).

bwateratmsft:
> Side note: I am going to hunt down where we create all of these ids and lower case them at the source. Because the number of times there are id-mismatches due to casing is killing me 😒

C# has a very easy way to do this:
```csharp
var dictionary = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```

I'm beginning to think we should make a utility class that is like that. `CaselessMap<K extends string, V>` or something like that.

src/tree/ResourceCache.ts
nturinski: I originally  was going to do that, but having this duplicated kind of bothered me:
```
for (const deleted of deletedResources) {
            const index = cachedResources.findIndex(cachedResource => cachedResource.id === deleted.id);
            cachedResources.splice(index, 1);
        }
```

I could always just make that a util function as well, I suppose.

bwateratmsft: I like that idea

alexweininger:
> > Side note: I am going to hunt down where we create all of these ids and lower case them at the source. Because the number of times there are id-mismatches due to casing is killing me 😒
> 
> C# has a very easy way to do this:
> 
> ```cs
> var myDict = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
> ```
> 
> I'm beginning to think we should make a utility class that is like that. `CaselessMap<K extends string, V>` or something like that.

We're getting to the point we need an entire utility class for resource ids, types, and kinds that handles normalization, comparison, getting the icon, display name, etc.

nturinski:
Maybe just implement something like this? https://stackoverflow.com/questions/50019920/javascript-map-key-value-pairs-case-insensitive-search

alexweininger:
Tested the PR out with static web apps.

* Tested creating new SWA, and verified it appeared in the tree quickly after creation


Unsure if that's related to this PR, but can we cleanup the spinner animations? The spinners have gone rogue.


https://user-images.githubusercontent.com/12476526/167022869-faece7a3-07c6-4aa6-86ab-a6d841258720.mov

alexweininger (approved):

nturinski:
An issue with my ResourceCache is that if an AppResource gets modified, it won't get updated on refresh.  I figure that there's two ways we can handle this:
- The cache does a shallow compare between the JSON objects.  If it's different, delete the old tree and re-create it with the modified AppResource
- The ResolvedWebApps are responsible for doing an API call to retrieve their own resource on refresh

The downfall of approach 2 is everything is basically going to re-resolve when you refresh the entire tree view.  The benefit would be that if you refreshed just the tree item, it should update it.  I think we could be smart about it and only re-resolve it if the app resource was actually modified.

Approach 1 could probably have the ResolvedTreeItems look at the cache and compare it to the resource to see if it needs an update?  But it would still require the cache to check with Azure to see if there were any updates on that specific tree item.

bwateratmsft (approved):

PR #223, opened 2022-05-05
Remove old call to resolve all children

nturinski:
I think we used to manually just resolve all the children whenever that group was expanded.  Pretty sure this was added before Brandon had set up all that provider logic and we were mostly using it to just test that the Proxies were being set up properly.  I think it should fix that weird spinning bug Alex was seeing:  https://github.com/microsoft/vscode-azureresourcegroups/pull/222#issuecomment-1119033983

Relies on https://github.com/microsoft/vscode-azuretools/pull/1148


src/tree/GroupTreeItemBase.ts
```diff
@@ -58,13 +59,11 @@ export class GroupTreeItemBase extends AzExtParentTreeItem {
...
     public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
-        if (clearCache) {
-            // this.treeMap = {};
-        }
-
         for (const ti of Object.values(this.treeMap)) {
             if (ti instanceof ResolvableTreeItemBase) {
-                void ti.resolve(clearCache, context);
+                if (!(ti.getResolver() instanceof WrapperResolver)) {
+                    void ti.resolve(clearCache, context);
```
bwateratmsft, line 65: If we're resolving in the tree framework, we shouldn't need to resolve here, regardless of type.

bwateratmsft (changes requested):
Mentioned this offline but I think we need to target all of our `resolve`ing to happen within a callback to `TreeView.onDidExpandElement`, plus within `pickAppResource`.

PR #228, opened 2022-05-07
Change how function type/kind is handled

### LINKED ISSUE #203: An error occurs when executing "Edit Tags..." action for a Function App
**OS:** All
**Azure Resource Extension Build Version:** 20220421.32 (unified)
**Azure Function Extension build Version:** 20220421.8 (unified)
**Regression:** Previous Build

**Repro Steps:**
1. Create a Function App.
2. Right click this Function App and execute "Edit Tags..." action.
3. Check whether the "xxx-tags.jsonc" file opens or not.

**Expect:**
The "xxx-tags.jsonc" file opens.
image

**Actual:**
An error occurs.
image

**More Info:**
This issue does not reproduce on the Azure Resource extension previous unified build 20220420.34.

alexweininger:
Fixes #203 

Function apps will now have `microsoft.web/functionapp` as their type instead of `microsoft.web/sites/functionapp`.


nturinski (approved):
Unrelated to your PR, but the tags don't actually update when you try to "view properties".  Probably the app resource needs to actually get updated/re-retrieved.  If you try to group by tag, it shows up so probably not a big deal.

PR #230, opened 2022-05-10
Fix double spinners

alexweininger:
Here's how it looks:

https://user-images.githubusercontent.com/12476526/167732854-127a4349-cc0d-497d-b6ab-5ed42f40bb1d.mov

### DIFF
--- a/src/tree/ResolvableTreeItemBase.ts
+++ b/src/tree/ResolvableTreeItemBase.ts
@@ -21,12 +21,14 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
     protected readonly contextValues: Set<string> = new Set<string>();
     public abstract parent?: AzExtParentTreeItem | undefined;
 
+    private _tempDescription: string | undefined = 'Loading...';
+
     public get contextValue(): string {
         return Array.from(this.contextValues.values()).sort().join(';');
     }
 
     public get description(): string | undefined {
-        return this.resolveResult?.description;
+        return this._tempDescription ?? this.resolveResult?.description;
     }
 
     public override get collapsibleState(): TreeItemCollapsibleState | undefined {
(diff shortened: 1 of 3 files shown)


nturinski:
I think the heart of the problem is that `runWithTemporaryDescription` calls refresh on itself at the end.  That's what causing those spinners.  I'd we prefer we still this it, but fix it with either
- Don't always call a full refresh (maybe allow a boolean that defaults to a hard refresh?
- Don't have a re-resolve in refresh for the AppResourceTreeItem.  Maybe there's something that we can check in the context value

src/tree/ResolvableTreeItemBase.ts
```diff
@@ -61,9 +64,10 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
                 }
 
                 this.resolveResult.contextValuesToAdd?.forEach(cv => this.contextValues.add(cv));
-            });
-
-            // It is not needed to refresh at this point, because `runWithTemporaryDescription` already does that
+            } finally {
+                this._tempDescription = undefined;
+                this.treeDataProvider.refreshUIOnly(this.parent);
```
nturinski, line 69: Why is it refreshing the parent?

alexweininger: The two spinners that appear are:
1. VS Code spinner that only appears when `getChildren` is called (`loadMoreChildrenImpl`)
2. the tree item icon which we can set to `loading~spin`.

To fix this, we just need to figure out why the VS Code spinner is appearing, aka why is VS Code trying to load the children when these items aren't expanded?

I discovered that calling `refreshUIOnly(this)` where `this` is the child being resolved resulted in a call to `loadMoreChildrenImpl`, while calling `refreshUIOnly(this.parent)` doesn't. 

Did some searching and it's actually a VS Code bug (https://github.com/microsoft/vscode/issues/121567).

---

Side note:

Fixing this seems to speed up the resolving quite a bit, I think this bug might have been (and probably still is in other places) causing a lot of unnecessary calls to `loadMoreChildrenImpl`. We should ask VS Code to investigate fixing this.

src/tree/ResolvableTreeItemBase.ts
```diff
@@ -46,13 +48,14 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
...
     }
 
-    public async resolve(clearCache: boolean, context: IActionContext): Promise<void> {
+    public async resolve(clearCache: boolean, _context: IActionContext): Promise<void> {
         if (!this.resolveResult || clearCache) {
             ext.activationManager.onNodeTypeResolved(this.data.type);
 
             const resolver = this.getResolver();
 
-            await this.runWithTemporaryDescription(context, 'Loading...', async () => {
```
bwateratmsft, line 55: A couple of thoughts...

1. Should we change `runWithTemporaryDescription` to also refresh the parent instead?
    1. Relatedly...if VSCode fixes https://github.com/microsoft/vscode/issues/121567, what happens if we do that? 
1. As an alternative, we could provide an option to `runWithTemporaryDescription` that eliminates the spinner icon, instead of implementing a sorta-custom version here.

alexweininger:
Closed in favor of #238

PR #236, opened 2022-05-11
Fix focus/unfocus group

alexweininger:


src/tree/SubscriptionTreeItem.ts
```diff
@@ -187,9 +187,6 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
             if (focusedGroup) {
                 return focusedGroup;
             }
-
-            // if we can't find it, erase the current setting, return current cached tree
-            await ext.context.workspaceState.update('focusedGroup', '');
```
alexweininger, line 192: Clearing this on the first call to `tryGetFocusedGroup` breaks the feature. It behaves fine if we never clear it.

bwateratmsft (approved):

PR #238, opened 2022-05-11
Fix double spinners

alexweininger:
Depends on https://github.com/microsoft/vscode-azuretools/pull/1156
Alternative to #230 

Here's how it looks:

https://user-images.githubusercontent.com/12476526/167943950-06bb1765-9062-45ac-beac-43d2b18fe972.mov

### DIFF
--- a/src/tree/ResolvableTreeItemBase.ts
+++ b/src/tree/ResolvableTreeItemBase.ts
@@ -6,6 +6,7 @@
 import { AzExtParentTreeItem, AzExtTreeItem, IActionContext } from "@microsoft/vscode-azext-utils";
 import { AppResource, AppResourceResolver, GroupableResource, GroupingConfig, ResolvedAppResourceBase } from "@microsoft/vscode-azext-utils/hostapi";
 import { TreeItemCollapsibleState } from "vscode";
+import { localize } from "vscode-nls";
 import { applicationResourceResolvers } from "../api/registerApplicationResourceResolver";
 import { ext } from "../extensionVariables";
 import { isBuiltinResolver } from "../resolvers/BuiltinResolver";
@@ -13,6 +14,8 @@ import { installableAppResourceResolver } from "../resolvers/InstallableAppResou
 import { outdatedAppResourceResolver } from "../resolvers/OutdatedAppResourceResolver";
 import { shallowResourceResolver } from "../resolvers/ShallowResourceResolver";
 import { wrapperResolver } from "../resolvers/WrapperResolver";
+
+const loading = localize('loading', "Loading...");
 
 export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem implements GroupableResource {
     public groupConfig: GroupingConfig;
@@ -25,7 +28,7 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
     // Setting this forces the tree item to always start out with a spinner icon, and have a "Loading..." description
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    private _temporaryDescription: string | undefined = 'Loading...';
+    private _temporaryDescription: string | undefined = loading;
 
     public get contextValue(): string {
         return Array.from(this.contextValues.values()).sort().join(';');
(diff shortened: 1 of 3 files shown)


src/tree/ResolvableTreeItemBase.ts
```diff
@@ -22,6 +22,11 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
     public abstract parent?: AzExtParentTreeItem | undefined;
     public abstract fullId: string;
 
+    // Setting this forces the tree item to always start out with a spinner icon, and have a "Loading..." description
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    private _temporaryDescription: string | undefined = 'Loading...';
```
alexweininger, line 28: I know this is a hack, but if we don't make the loading state the default, then it results in some flashing.

nturinski, line 28: Why is this necessary?  Because of the refresh?

src/tree/ResolvableTreeItemBase.ts
```diff
@@ -53,7 +58,10 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
 
             const resolver = this.getResolver();
 
-            await this.runWithTemporaryDescription(context, 'Loading...', async () => {
+            await this.runWithTemporaryDescription(context, {
+                description: 'Loading...',
```
nturinski, line 62: Should probably go ahead and just localize this.

src/tree/ResolvableTreeItemBase.ts
alexweininger: See above comment

src/tree/ResolvableTreeItemBase.ts
```diff
@@ -22,6 +25,11 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
     public abstract parent?: AzExtParentTreeItem | undefined;
     public abstract fullId: string;
 
+    // Setting this forces the tree item to always start out with a spinner icon, and have a "Loading..." description
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    private _temporaryDescription: string | undefined = loading;
```
bwateratmsft, line 31: image

src/tree/ResolvableTreeItemBase.ts
```diff
@@ -53,7 +61,10 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
 
             const resolver = this.getResolver();
 
-            await this.runWithTemporaryDescription(context, 'Loading...', async () => {
+            await this.runWithTemporaryDescription(context, {
+                description: loading,
+                skipRefresh: true
```
bwateratmsft, line 66:
```suggestion
                softRefresh: true
```

In the utils package it's `softRefresh` (which I think is a better name as it's still doing _some_ refresh)

src/tree/ResolvableTreeItemBase.ts
```diff
@@ -53,7 +61,10 @@ export abstract class ResolvableTreeItemBase extends AzExtParentTreeItem impleme
...
+            }, async () => {
                 this.resolveResult = await resolver.resolveResource(this.subscription, this.data);
 
                 // Debug only?
```
bwateratmsft, line 70: There's a comment below (but GitHub won't let me add it in context):
> `// It is not needed to refresh at this point, because runWithTemporaryDescription already does that`

Can we make sure that resolving can update the label/description? I _think_ so but I'm not totally sure. Stopping a function app makes resolving update the description so IMO that's a sufficient test--if that "Stopped" description still appears I'm happy.

bwateratmsft (approved):

src/tree/ResolvableTreeItemBase.ts
bwateratmsft: :chef's kiss:
Perfect! Also, love how fast SWA resolves.

src/tree/ResolvableTreeItemBase.ts
alexweininger: Was there an action Item here or are we just celebrating Frozen 2?

bwateratmsft: The second one (just celebrating Frozen 2)

alexweininger: okbye

PR #251, opened 2022-05-13
Remove container from the supported resourceTypes

nturinski:
Not planning to release container apps with the unified experience, so just removing the resource type so that we're not pulling for it.


alexweininger:
Did you mean to make this PR against main instead of unified?

PR #256, opened 2022-05-16
Make pickAppResource case insensitive

alexweininger:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3170

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -41,7 +41,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         if (options?.filter) {
             const filters: AppResourceFilter[] = Array.isArray(options.filter) ? options.filter : [options.filter];
             appResources = appResources.filter((appResource) => filters.some(filter => {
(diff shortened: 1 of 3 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -41,7 +41,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         if (options?.filter) {
             const filters: AppResourceFilter[] = Array.isArray(options.filter) ? options.filter : [options.filter];
             appResources = appResources.filter((appResource) => filters.some(filter => {
-                if (getResourceType(filter.type, filter.kind) === getResourceType(appResource.data.type, appResource.data.kind)) {
+                if (getResourceType(filter.type, filter.kind).toLowerCase() === getResourceType(appResource.data.type, appResource.data.kind).toLowerCase()) {
```
alexweininger, line 44: `.toLowerCase()`'ing it here rather than inside `getResourceType` because I know doing it here won't break anything else. I plan to revisit this whole id/type/kind mess in the future.

bwateratmsft (approved):

PR #257, opened 2022-05-16
Add walkthrough

alexweininger:
Added a new command to show the Group By options
image

Also added a welcome view to the Activity Log panel
image


src/commands/registerCommands.ts
```diff
@@ -57,4 +58,5 @@ export function registerCommands(): void {
 
     registerCommand('azureResourceGroups.clearActivities', clearActivities);
     registerCommand('azureResourceGroups.toggleShowAllResources', toggleShowAllResources);
+    registerCommand('azureResourceGroups.showGroupOptions', showGroupOptions);
```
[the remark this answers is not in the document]
bwateratmsft: It might be best to just include it in the package manifest along with an activation event, and suppress it there from the command palette. If the walkthrough opens and the user very quickly clicks that button _before_ the RG extension has time to activate, it will throw an error.

package.json
```diff
@@ -442,6 +449,64 @@
...
+            {
+                "id": "azure-get-started",
+                "title": "Get Started with Azure in VS Code",
+                "description": "Setup your account and get to know the Azure extension view.",
+                "steps": [
+                    {
+                        "id": "sign-in",
+                        "title": "Sign into your Azure account",
+                        "description": "Sign in or create an Azure Account to see all of the resources within your Azure subscription.\n[Sign In](command:azure-account.login)\n",
+                        "when": "!azure-account.signedIn",
```
bwateratmsft, line 463: This step appears briefly and then disappears once the extension is activated. It might look nicer if we flip around the logic; i.e. the step is hidden and the context is updated to show it. However, I don't know if the focus will jump around in a weird way if this happens.

alexweininger:
Sign in step disappearing:

https://user-images.githubusercontent.com/12476526/168925305-7b79e464-c384-427b-9962-71bcf3a19047.mov

package.json
alexweininger: Matt says this behavior is fine

package.json
```diff
@@ -188,6 +189,11 @@
                 "command": "azureResourceGroups.toggleShowAllResources",
                 "title": "%azureResourceGroups.toggleShowAllResources%",
                 "category": "Azure"
+            },
+            {
+                "command": "azureResourceGroups.showGroupOptions",
```
bwateratmsft, line 194: This command should be suppressed from the command palette

bwateratmsft (dismissed):

bwateratmsft (dismissed):

alexweininger:
Brandon I'm so sorry lol

bwateratmsft (approved):

PR #258, opened 2022-05-17
Remove container apps

alexweininger:


src/azureExtensions.ts
```diff
@@ -81,14 +81,14 @@ export const azureExtensions: IAzExtMetadata[] = [
...
-        resourceTypes: [
-            'microsoft.app/containerapps'
-        ],
-        reportIssueCommandId: 'containerApps.reportIssue'
-    }
+    // {
+    //     name: 'vscode-azurecontainerapps',
+    //     label: 'Container Apps',
+    //     resourceTypes: [
+    //         'microsoft.app/containerapps'
```
nturinski, line 88: Just comment out this line.  We still want the reportIssueCommandId to get registered otherwise it won't show up when users try to report a problem.

PR #266, opened 2022-05-17
Don't suppress resolve telemetry or error messages

### LINKED ISSUE #235: App centric telemetry
* Make sure we're not suppressing resolve telemetry

alexweininger:
Fixes #235

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -58,6 +58,8 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
 
         // Hook up the resolve handler
         registerEvent('treeItem.expanded', ext.appResourceTree.onDidExpandOrRefreshExpandedTreeItem, async (context: IActionContext, treeItem: AzExtTreeItem) => {
+            context.errorHandling.suppressDisplay = true;
+
             if (treeItem instanceof GroupTreeItemBase) {
                 await treeItem.resolveAllChildrenOnExpanded(context);
             }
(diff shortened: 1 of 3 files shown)


nturinski (commented):
Did we want to display the error though? The issue specifically mentions suppressing for telemetry, but not sure if the user needs/wants to know about these errors.

alexweininger:
> Did we want to display the error though? The issue specifically mentions suppressing for telemetry, but not sure if the user needs/wants to know about these errors.

I was wondering the same thing. Was hoping you would have some epic wisdom on what we should do. I guess we shouldn't show it but I can't give a good reason why. Maybe we wrap it in try/catch and show a nicer error message?

src/extension.ts
```diff
@@ -87,6 +89,8 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
         registerApplicationResourceResolver('vscode-azureresourcegroups.wrapperResolver', wrapperResolver);
         registerApplicationResourceResolver('vscode-azureresourcegroups.installableAppResourceResolver', installableAppResourceResolver);
         registerApplicationResourceResolver('vscode-azureresourcegroups.shallowResourceResolver', shallowResourceResolver);
+
+        await vscode.commands.executeCommand('setContext', 'azure-account.signedIn', await ext.rootAccountTreeItem.getIsLoggedIn());
```
nturinski, line 93: Is this related to the walkthrough?

nturinski:
> > Did we want to display the error though? The issue specifically mentions suppressing for telemetry, but not sure if the user needs/wants to know about these errors.
> 
> I was wondering the same thing. Was hoping you would have some epic wisdom on what we should do. I guess we shouldn't show it but I can't give a good reason why. Maybe we wrap it in try/catch and show a nicer error message?

I'm actually not even sure what it looks like when it does error.  I'm worried that one error might prevent the whole group from being displayed.  I'll try forcing an error for the resolve-- might have a better insight after that.  But I think it's safe to suppress for now.

src/extension.ts
alexweininger: Damn, how did that get there

nturinski (approved):

PR #268, opened 2022-05-18
Refresh UI only to prevent extra calls to refreshImpl

alexweininger:
This prevents `GroupTreeItem.refreshImpl` from being called every time it gets created. I don't believe this is expected behavior.

I'm also unsure if the `this should actually be "resolve"` comment still holds up


nturinski (dismissed):
It does not-- feel free to remove that comment.

nturinski:
bump

nturinski (approved):

PR #269, opened 2022-05-18
Add and use pickResourceGroup method

### LINKED ISSUE #260: There is an extra step "Select Resource" when executing "Azure: Delete..." from command palette
**OS:** Win10
**Azure Resources Build Version:** 20220516.8

**Repro Steps:**

1. Make sure the resource groups by Resource Type.
2. F1 -> Azure: Delete... -> Select a subscription -> Select a resource group, e.g. "Azure Cosmos DB".
3. Check whether executes the command "Azure: Delete.." directly.

**Expect:**
Execute the command "Azure: Delete.." directly with an error.

**Actual:**
There is an extra step "Select Resource".
image

alexweininger:
Fixes #260

### DIFF
--- a/src/tree/ResourceCache.ts
+++ b/src/tree/ResourceCache.ts
@@ -49,7 +49,7 @@ export class ResourceCache {
         this._resourceGroups.push(...addedResources);
     }
 
-    public getTreeMap(context: IActionContext): GroupTreeMap {
+    public getTreeMap(context: IActionContext, groupBySetting?: string): GroupTreeMap {
         const treeMap: GroupTreeMap = {};
         const getResourceGroupTask: (resourceGroup: string) => Promise<ResourceGroup | undefined> = async (resourceGroup: string) => {
             return this._resourceGroups.find((rg) => rg.name === resourceGroup);
@@ -62,7 +62,7 @@ export class ResourceCache {
         });
 
         treeMap[ungroupedTreeItem.id] = ungroupedTreeItem;
-        const groupBySetting = <string>settingUtils.getWorkspaceSetting<string>('groupBy');
+        groupBySetting ||= <string>settingUtils.getWorkspaceSetting<string>('groupBy');
         for (const rgTree of this._appResources) {
             (<AppResourceTreeItem>rgTree).mapSubGroupConfigTree(context, groupBySetting, treeMap, getResourceGroupTask);
         }
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -5,7 +5,7 @@
 
 import { ResourceGroup, ResourceManagementClient } from '@azure/arm-resources';
 import { IResourceGroupWizardContext, LocationListStep, ResourceGroupCreateStep, ResourceGroupNameStep, SubscriptionTreeItemBase, uiUtils } from '@microsoft/vscode-azext-azureutils';
-import { AzExtParentTreeItem, AzExtTreeItem, AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, ExecuteActivityContext, IActionContext, IAzureQuickPickOptions, ICreateChildImplContext, ISubscriptionContext, nonNullOrEmptyValue, nonNullProp, registerEvent } from '@microsoft/vscode-azext-utils';
+import { AzExtParentTreeItem, AzExtTreeItem, AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, ExecuteActivityContext, IActionContext, IAzureQuickPickItem, IAzureQuickPickOptions, ICreateChildImplContext, ISubscriptionContext, nonNullOrEmptyValue, nonNullProp, registerEvent } from '@microsoft/vscode-azext-utils';
 import { AppResourceFilter, PickAppResourceOptions } from '@microsoft/vscode-azext-utils/hostapi';
 import { ConfigurationChangeEvent, workspace } from 'vscode';
 import { applicationResourceProviders } from '../api/registerApplicationResourceProvider';
@@ -21,6 +21,11 @@ import { AppResourceTreeItem } from './AppResourceTreeItem';
 import { GroupTreeItemBase } from './GroupTreeItemBase';
 import { GroupTreeMap, ResourceCache } from './ResourceCache';
 import { ResourceGroupTreeItem } from './ResourceGroupTreeItem';
+
+interface PickResourceGroupOptions {
+    canPickMany?: boolean;
+    placeholder?: string;
+}
 
 export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
     public readonly childTypeLabel: string = localize('resourceGroup', 'Resource Group');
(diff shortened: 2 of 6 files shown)


src/tree/SubscriptionTreeItem.ts
```diff
@@ -76,6 +81,26 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         return result.data;
     }
 
+    public async pickResourceGroup(context: IActionContext, options: PickResourceGroupOptions & { canPickMany: true }): Promise<ResourceGroupTreeItem[]>;
+    public async pickResourceGroup(context: IActionContext, options: PickResourceGroupOptions & { canPickMany: false }): Promise<ResourceGroupTreeItem>;
+    public async pickResourceGroup(context: IActionContext, options: PickResourceGroupOptions): Promise<ResourceGroupTreeItem | ResourceGroupTreeItem[]> {
+        await this.getCachedChildren(context);
+        const rgsTreeMap = this.cache.getTreeMap(context, 'resourceGroup');
```
nturinski, line 88: I think rather than making the tree map, we can leverage the ResourceGroups on the ResourceCache.  

We can just make sure that it is resolved whenever this command is called.  I don't think you would even need the tree item for deleting the ResourceGroup.

This method certainly works, but it makes the getTreeMap function a little hacky and feels like a bit of a workaround.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -76,6 +81,26 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
+        const tis = (await context.ui.showQuickPick((Object.values(rgsTreeMap).filter((groupTreeItem) => groupTreeItem instanceof ResourceGroupTreeItem) as ResourceGroupTreeItem[]).map((rg: ResourceGroupTreeItem): IAzureQuickPickItem<ResourceGroupTreeItem> => ({
```
nturinski, line 89: Might be worth to include this implementation in the azext-utils package.  I know in the interest of time, might not be feasible now, but I could see this being useful in other commands.

src/tree/SubscriptionTreeItem.ts
nturinski: I think it would make sense for the SubscriptionTreeItem to have a `getResourceGroupsTask` promise that we can just check and re-resolve on refreshes

alexweininger: > I think rather than making the tree map, we can leverage the ResourceGroups on the ResourceCache.
>
> We can just make sure that it is resolved whenever this command is called. I don't think you would even need the tree item for deleting the ResourceGroup.

Although we can delete the RG without the `ResourceGroupTreeItem`, in order to show a temporary description and a spinner on the resource group during delete, we need to call `.deleteTreeItem`.

nturinski: That would only be in the case if the groupBy setting was already set to ResourceGroup, right?  Otherwise it wouldn't be displayed in UI anyway.

In that case, you could take the RG and if you find the `ResourceGroupTreeItem`, do `.deleteTreeItem`; otherwise just delete it via the API call.

I don't have a very strong opinion about this, but it doesn't really feel like how `getTreeMap` should be used especially since you're calling `getCachedChildren` which would also call `getTreeMap`.  I'm worried that this code might mess with the `AppResourceTreeItems`

```
public mapSubGroupConfigTree(context: IActionContext,
        groupBySetting: string,
        treeMap: GroupTreeMap,
        getResourceGroup: (resourceGroup: string) => Promise<ResourceGroup | undefined>): void {

        const configId: string | undefined = this.groupConfig[groupBySetting]?.id.toLowerCase() ?? `${this.rootGroupConfig.id}/${ungroupedId}`;
        let subGroupTreeItem = treeMap[configId];
        if (!subGroupTreeItem) {
            subGroupTreeItem = this.createSubGroupTreeItem(context, groupBySetting, getResourceGroup);
            treeMap[configId] = subGroupTreeItem;
        }

        subGroupTreeItem.treeMap[this.id] = this;
        void ext.appResourceTree.refreshUIOnly(subGroupTreeItem);
    }
```

nturinski: I don't think it does since it doesn't actually set anything on the `AppResourceTreeItem` and treeMap is something is passed in, but that could definitely change in the future.

alexweininger: I don't love the fact that different things would be happening based on the group by setting. And also if they switched to grouping by resource group after initiating a delete, the deleting... state wouldn't be reflected. That's definitely a nit at this point, but I can see how that isn't the best UX.

I could make a new method that is responsible for creating the RG tree items and caching them... But I want to avoid a large change like that so close to release.

nturinski: What if we change 
```
public set resourceGroups(resourceGroups: ResourceGroup[]) {
        const { addedResources, deletedResources } = this.diffProvidedResources(this._resourceGroups, resourceGroups);
        this.deleteResourcesFromCache(this._resourceGroups, deletedResources);
        this._resourceGroups.push(...addedResources);
    }
```

to store the `ResourceGroupTreeItems` rather than the `ResourceGroup` exactly like `appResources` does?

src/tree/ResourceCache.ts
```diff
@@ -66,7 +66,7 @@ export class ResourceCache {
         });
 
         treeMap[ungroupedTreeItem.id] = ungroupedTreeItem;
-        const groupBySetting = <string>settingUtils.getWorkspaceSetting<string>('groupBy');
+        groupBySetting ||= <string>settingUtils.getWorkspaceSetting<string>('groupBy');
```
nturinski, line 69: I think you can get rid of this logic now?

nturinski (approved):

src/tree/SubscriptionTreeItem.ts
```diff
@@ -76,6 +81,29 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
 
+    public async pickResourceGroup(context: IActionContext, options: PickResourceGroupOptions & { canPickMany: true }): Promise<ResourceGroupTreeItem[]>;
+    public async pickResourceGroup(context: IActionContext, options: PickResourceGroupOptions & { canPickMany: false }): Promise<ResourceGroupTreeItem>;
+    public async pickResourceGroup(context: IActionContext, options: PickResourceGroupOptions): Promise<ResourceGroupTreeItem | ResourceGroupTreeItem[]> {
+        if (this.cache.resourceGroups.length === 0) {
+            const client: ResourceManagementClient = await createResourceClient([context, this.subscription]);
+            this.cache.resourceGroups = await uiUtils.listAllIterator(client.resourceGroups.list());
+        }
+
+        const tis = (await context.ui.showQuickPick(this.cache.resourceGroups.sort((a, b) => a.name.localeCompare(b.name)).map((rg: ResourceGroupTreeItem): IAzureQuickPickItem<ResourceGroupTreeItem> => ({
```
bwateratmsft, line 92: Just to double check, if there are zero resource groups in the subscription, this will throw the `NoResourcesFoundError`, correct? (And thus, not go on to the logic below)

alexweininger: It does now!

bwateratmsft (approved):

PR #272, opened 2022-05-18
Cache resource group tree items rather than the resource groups

nturinski:


bwateratmsft (commented):
Which issue was this PR fixing? I'm a little concerned about caching tree items themselves rather than the underlying data.

PR #278, opened 2022-05-23
Maintain focus on refresh

### LINKED ISSUE #277: All resource groups show in the "Resources" tree view after refreshing it
**OS:** M1 & Win10
**Azure Resource Extension Build Version:** 20220522.1

**Repro Steps:**

1. Make sure the resource groups by Resource Group.
2. Click the "Focus this Group" icon to focus a resource group -> Refresh "Resources" tree view.
3. Check whether the focused resource group shows in the "Resources" tree view.

**Expect:**
Only the focused resource group shows in the "Resources" tree view.
image

**Actual:**
1. All resource groups show in the "Resources" tree view.
2. The icon of the previous focused resource group still is "Unfocus Group".
image

alexweininger:
Fixes #277

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azureresourcegroups",
-    "version": "0.4.1-alpha.28",
+    "version": "0.4.1-alpha.29",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-azureresourcegroups",
-            "version": "0.4.1-alpha.28",
+            "version": "0.4.1-alpha.29",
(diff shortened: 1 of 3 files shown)


bwateratmsft (approved):
If you haven't already, double check that focus/unfocus works on all grouping modes and has a reasonable result if the focused group goes away (e.g. a resource group is deleted).

alexweininger:
> If you haven't already, double check that focus/unfocus works on all grouping modes and has a reasonable result if the focused group goes away (e.g. a resource group is deleted).

If the focused resource group is deleted, no tree items are shown until the view is refreshed. I think that's ok for now.

Issue #264, opened 2022-05-17
[Unstable] An error occurs when expanding a resource group


v-mengwe:
**OS:** Win10
**Build Version:** 20220516.8
**Regression:** Not a regression

**Repro Steps:**

1. Make sure the resource groups by Resource Group -> Expand a resource group.
2. Check whether no error occurs.

**Expect:**
No error occurs.

**Actual:**
An error occurs.
![image](https://user-images.githubusercontent.com/62053665/168799288-4aae07dc-e846-4876-acfc-b1840dff1e24.png)

**More Info:**
This error pops up irregularly and the frequency is chaotic.

alexweininger:
I've seen this very rarely with SWA and Web Apps. Seems hard to reproduce.

alexweininger:
We've seen reports of this also happening if a group is focused before it is expanded.

alexweininger:
Created https://github.com/microsoft/vscode-azureresourcegroups/issues/284 because although it's the same error message, it's likely a different bug than this one.

alexweininger:
The only way I've been able to reproduce this is with these steps:
1. Focus on a resource group
2. Unfocus resource group
3. Go and expand other resource groups

PR #291, opened 2022-05-31
Fix registering duplicate tree items

alexweininger:
Fixes https://github.com/microsoft/vscode-azureresourcegroups/issues/264

Depends on https://github.com/microsoft/vscode-azuretools/pull/1170

The bug is definitely caused by [setting the cached children](https://github.com/microsoft/vscode-azureresourcegroups/blob/main/src/tree/SubscriptionTreeItem.ts#L180) in `SubscriptionTreeItem`. I've tried to fix this bug by no longer setting the cached children, but haven't been able to replicate the nice behavior that it gives. So I think this is an acceptable fix.

The changes in this PR and in the tools package allow us to properly clear VS Code's internal cache before it makes calls to `getChildren`.


src/extension.ts
```diff
@@ -57,11 +57,14 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
 
         // Hook up the resolve handler
-        registerEvent('treeItem.expanded', ext.appResourceTree.onDidExpandOrRefreshExpandedTreeItem, async (context: IActionContext, treeItem: AzExtTreeItem) => {
+        registerEvent('treeItem.expanded', ext.appResourceTree.onDidExpandOrRefreshExpandedTreeItem, async (context: IActionContext, data: OnDidExpandOrRefreshExpandedEmitterData) => {
             context.errorHandling.suppressDisplay = true;
-
-            if (treeItem instanceof GroupTreeItemBase) {
-                await treeItem.resolveAllChildrenOnExpanded(context);
+            if (data.item instanceof GroupTreeItemBase) {
+                if (data.source !== 'refresh') {
```
alexweininger, line 63: Don't call `refreshUIOnly` if this event was triggered from a refresh to prevent an infinite loop.

src/extension.ts
```diff
@@ -57,11 +57,14 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
+                    // clear VS Code's internal cache
+                    ext.appResourceTree.refreshUIOnly(undefined);
```
bwateratmsft, line 65: I don't think I understand why we should clear VSCode's cache when expanding a node.

alexweininger: Focusing, then unfocusing a group puts the tree view in a state in which expanding group nodes causes the "Tree item with id X is already registered" error. 

My theory is: on focus the app resource tree items are not removed from VS Code's tree item cache, but the group tree items are removed. So VS Code's cache contains the children, but not the parents. On unfocus, we give VS Code the parents (group tree items). 

Then, when a group tree item is expanded VS Code doesn't associate the app resources that were never removed from the cache on focus with the new group tree items, and throws an error when `getChildren` returns items which already exist in VS Code's cache.

The best fix is to make sure the app resources are cleared from the cache on focus/unfocus. I tried a bunch of things but was never able to get this to happen. `refreshUIOnly` is already [called upon every focus/unfocus command](https://github.com/microsoft/vscode-azureresourcegroups/blob/a1290599aa1e7ea771de36c0d830e7b33cdd6cc0/src/tree/SubscriptionTreeItem.ts#L217). To me this should completely clear VS Code's cache on every focus/unfocus. However it doesn't seem to be doing that.

So I ended up on this "band aid" solution.

alexweininger:
Closing in favor of https://github.com/microsoft/vscode-azuretools/pull/1174

PR #294, opened 2022-06-01
Resolve on refresh

### LINKED ISSUE #280: Resources are not re-resolved after installing extensions
Shallow resources that say "Install extension to enable additional features..." are not re-resolved and refreshed upon installing an extension. The workaround is to reload VS Code 😞

alexweininger:
Fixes #280

https://user-images.githubusercontent.com/12476526/171518566-0819bc36-8971-40ae-8e55-692d387a69f3.mov


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-azureresourcegroups",
     "displayName": "Azure Resources",
     "description": "%azureResourceGroups.description%",
-    "version": "0.5.2",
+    "version": "0.5.3-0",
```
bwateratmsft, line 5: Versioning did it weird again

src/tree/AppResourceTreeItem.ts
```diff
@@ -128,9 +128,10 @@ export class AppResourceTreeItem extends ResolvableTreeItemBase implements Group
         // do nothing as we only want to return parent dynamically
     }
 
-    public async refreshImpl(): Promise<void> {
+    public async refreshImpl(context: IActionContext): Promise<void> {
         this.mTime = Date.now();
         ext.tagFS.fireSoon({ type: FileChangeType.Changed, item: this });
+        await super.refreshImpl(context);
```
bwateratmsft, line 134: Shouldn't refresh be dumping all of these nodes entirely?

alexweininger: This refresh implementation is for a single app resource, which nodes are you referring to? The children are automatically cleared from cache on refresh by utils.

bwateratmsft: The bug is that when we refresh, we aren't re-resolving, right? I think that we should entirely clear the cache (i.e. the `ResourceCache`), including the tree nodes, resolve results, etc., when we refresh. This issue would be fixed if we can trigger a refresh when these extensions install.

alexweininger: It sounds like you want to refresh and re-resolve all resources when an extension is installed. I think installing the Functions extension should only refresh and re-resolve Function app resources instead of all app resources.

bwateratmsft: That's definitely reasonable, but is it worth the extra work?

alexweininger: I also believe refreshing a group should re-resolve all children. And refreshing an app resource should re-resolve it. The code I added also adds this, so the extra work is probably needed anyway.

bwateratmsft:
Is this affected at all by #303 / https://github.com/microsoft/vscode-azuretools/pull/1174?

alexweininger:
> Is this affected at all by #303 / [microsoft/vscode-azuretools#1174](https://github.com/microsoft/vscode-azuretools/pull/1174)?

The behavior in main nor the behavior in this PR seem to be affected by #303

bwateratmsft (approved):

PR #307, opened 2022-06-09
Enable canSelectMany for workspace tree view

### LINKED ISSUE #300: Cannot select multiple files and folders at once for an attached storage account
**OS:** Win10/M1
**Azure Resource Extension Build Version:** 20220523.7
**Azure Storage Extension Build Version:** 20220523.2

**Repro Steps:**
1. Expand an attached storage account -> Create a blob -> Upload some files and folders to it.
2. Check whether select multiple files and folders at once.

**Expect:**
Select multiple files and folders at once.
image

**Actual:**
Cannot select multiple files and folders at once.

**More Info:**
1. This issue also reproduces for the file share.
2. This issue also reproduces for the Local Emulator.
3. This issue doesn't reproduce for a normal storage account.

alexweininger:
Fixes #300

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-azureresourcegroups",
-    "version": "0.5.3",
+    "version": "0.5.4-alpha.1",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "vscode-azureresourcegroups",
-            "version": "0.5.3",
+            "version": "0.5.4-alpha.1",
             "license": "SEE LICENSE IN LICENSE.md",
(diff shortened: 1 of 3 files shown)


bwateratmsft (approved):
It's probably a near-certainty that this will have some unintended effects...for example, the user selects a handful of things across totally different extensions and then right-clicks a local function. VSCode will show the context menu options for the function but pass _all_ of the selected items to the command, which could result in some odd behavior.

There's no solving the oddities of what can be selected vs. what can be executed, but we should add some guards in the command code to ignore tree items that were selected but not applicable.

nturinski (approved):

PR #311, opened 2022-06-10
Fetch resources to get number of resources in RG

### LINKED ISSUE #308: The confirmation message is inconsistent when deleting a resource group from context menu and command palette
**OS:** Win10
**Azure Resource Extension Build Version:** 20220609.5

**Repro Steps:**
1. Right click a resource group -> Click "Delete Resource Group..." -> Check the confirmation message in the dialog.
2. F1 -> Click "Azure: Delete Resource Group..." -> Select the above selected resource group -> Check the confirmation message in the dialog.
3. Check whether the confirmation message is consistent between in the step1 and step 2.

**Expect:**
The confirmation message is consistent between in the step1 and step 2.

**Actual:**
The confirmation message is inconsistent between in the step1 and step 2.
image

alexweininger:
Fixes #308 

Deriving the number of resources in a resource group from the cached children (or a similar source like the treeMap) results in the number being wrong unless view is grouped by resource group and the subscription tree item has been expanded.

It's much simpler to just fetch the list of resources and use that.

### DIFF
--- a/src/tree/ResourceGroupTreeItem.ts
+++ b/src/tree/ResourceGroupTreeItem.ts
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { ResourceGroup, ResourceManagementClient } from "@azure/arm-resources";
-import { uiUtils } from "@microsoft/vscode-azext-azureutils";
 import { AzExtParentTreeItem, AzureWizard, IActionContext, nonNullProp, TreeItemIconPath } from "@microsoft/vscode-azext-utils";
 import { GroupNodeConfiguration } from "@microsoft/vscode-azext-utils/hostapi";
 import { FileChangeType } from "vscode";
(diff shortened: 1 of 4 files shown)


nturinski (dismissed):

src/tree/ResourceGroupTreeItem.ts
```diff
@@ -64,13 +65,8 @@ export class ResourceGroupTreeItem extends GroupTreeItemBase {
     }
 
     public async getNumOfResources(context: IActionContext): Promise<number> {
```
bwateratmsft, line 67: Since the deletion is the only place to use this, we should move this code there, to ensure it's scoped to _only_ that scenario. We wouldn't want live updates coming from the web all the time.

bwateratmsft (changes requested):

bwateratmsft (approved):

PR #326, opened 2022-07-06
Remove container app resource types from resolvers

nturinski:
The version bot better shut up on this one 😠 

Just commenting out the azureExtension part of it.  The icons/labels will still be in there. Shouldn't be an issue, makes it look better but it's also only seen if the hiddenStacks is enabled anyway.


alexweininger (approved):

alexweininger:
The version bot is dead #322

PR #328, opened 2022-07-07
V2 Quick Pick API Discussion

philliphoff:
# Do not merge! #

A draft PR for feedback and discussion regarding the V2 Quick Pick API.


src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -50,8 +50,8 @@ export interface ApplicationResourceProvider extends ResourceProviderBase<Applic
 }
 
 export interface ResourceQuickPickOptions {
-    readonly contexts?: string[];
-    readonly isParent?: boolean;
+    readonly contexts: string[];
+    readonly isLeaf: boolean;
```
philliphoff, line 54: `hasChildren`?  VS Code tends to prefer parent/child terminology within the tree; while we're intending to use "branch", we may want to make sure we don't take the metaphor too far.

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -148,7 +148,7 @@ export interface V2AzureResourcesApi extends AzureResourcesApiBase {
     /**
      * Show a quick picker of app resources. Set `options.type` to filter the picks.
      */
-    pickResource<TModel>(options?: ResourcePickOptions): vscode.ProviderResult<TModel>
+    pickResource<TModel extends ResourceModelBase>(actionContext: IActionContext, options?: ResourcePickOptions): vscode.ProviderResult<TModel | ApplicationResource>;
```
philliphoff, line 151: If at all possible, I would really, really, *really* prefer to keep `IActionContext` out of the API.  For the most part, that interface represents implementation details of our extensions (e.g. a conveyance for telemetry properties) that I don't think should be forced upon all extensions.

philliphoff, line 151: I was thinking that there would be an expectation that callers of `pickResource<TModel>()` would have specific knowledge of the returned model type.  That is, an extension exposing a command that picks one of its resources would always expect to get back one of its model types, even if the thing being picked was an application resource (i.e. otherwise an `ApplicationResource`).  Hence, returning `TModel | ApplicationResource` seems a little odd to me.  Even in the case where the Resource Groups extension is doing the picking, it'd still (I would think) expect to get back its own model type, which isn't an `ApplicationResource` but, rather, a type that wraps an `ApplicationResource`.

src/api/v2/quickPickWizard/QuickPickAppResourceStep.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    private matchesAppResource(resource: ApplicationResource): boolean {
+        if (!this.options.filter) {
+            return true;
+        }
+
+        const filterArray = Array.isArray(this.options.filter) ? this.options.filter : [this.options.filter];
+
+        return filterArray.some(filter => {
+            if (filter.type.toLowerCase() !== resource.type.toLowerCase()) {
```
philliphoff, line 67: Was thinking about this earlier and not specific to the quick pick; we should be normalizing the `type` and `kind` properties of `ApplicationResource` on creation.  I'm finding `.toLowerCase()` creeping into a lot of places.

src/api/v2/v2AzureResourcesApi.ts
bwateratmsft: I want it to be a fixed property that does not depend on whether or not the node actually has children, otherwise, we have to start querying deeper into the tree than we ought. I worried that "isParent" had the same ambiguity. However, "isLeaf" is pretty clear IMO.

We could also do something like "canHaveChildren".

src/api/v2/quickPickWizard/CreateQuickPickStep.ts
```diff
@@ -0,0 +1,40 @@
...
+import { localize } from '../../../utils/localize';
+import { ResourceModelBase } from '../v2AzureResourcesApi';
+import { ContextValueFilter } from './ContextValueFilter';
+import { GenericQuickPickStep } from './GenericQuickPickStep';
+import { QuickPickWizardContext } from './QuickPickWizardContext';
+
+type CreateCallback = () => vscode.ProviderResult<never>;
+
+export class CreateQuickPickStep<TModel extends ResourceModelBase> extends GenericQuickPickStep<TModel> {
+    public constructor(treeDataProvider: vscode.TreeDataProvider<TModel>, contextValueFilter: ContextValueFilter, private readonly createCallback: CreateCallback) {
```
nturinski, line 17: Also might be worth having a contextExcluder.  The perfect example is for Logic Apps vs. Function Apps.  Both have the same kind, but Logic Apps has an extra kind, `workflowapp`.  However, we wouldn't want to have _exact_ matching because Function apps can have both `functionapp` and `functionapp,linux`

src/api/v2/quickPickWizard/GenericQuickPickStep.ts
```diff
@@ -0,0 +1,65 @@
...
+        super();
+    }
+
+    public async prompt(wizardContext: QuickPickWizardContext<TModel>): Promise<void> {
+        const selected = await wizardContext.ui.showQuickPick(await this.getPicks(wizardContext), { /* TODO: options */ });
+        wizardContext.currentNode = selected.data;
+    }
+
+    public shouldPrompt(_wizardContext: QuickPickWizardContext<TModel>): boolean {
+        return true;
```
nturinski, line 28: I was thinking we may be able to skip the prompt step if:

1. This isn't a matching context value
2. AND There is only one available option
3. OR the user has specified what should be selected as the currentNode at this point

src/api/v2/quickPickWizard/GenericQuickPickStep.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    protected async getPicks(wizardContext: QuickPickWizardContext<TModel>): Promise<IAzureQuickPickItem<TModel>[]> {
+        const children = (await this.treeDataProvider.getChildren(wizardContext.currentNode)) || [];
+
+        const matchingChildren = children.filter(c => matchesContextValueFilter(c, this.contextValueFilter));
+        const nonLeafChildren = children.filter(c => c.quickPickOptions?.isLeaf === false);
```
nturinski, line 35: nit: Don't need to run this until after check of `matchingChildren.length === 0`

src/api/v2/quickPickWizard/GenericQuickPickStep.ts
```diff
@@ -0,0 +1,65 @@
...
+                throw new NoResourceFoundError();
+            } else {
+                promptChoices = nonLeafChildren;
+            }
+        } else {
+            promptChoices = matchingChildren;
+        }
+
+        const picks: IAzureQuickPickItem<TModel>[] = [];
+        for (const choice of promptChoices) {
```
nturinski, line 49: Rather than doing a somewhat recursive loop like this, I was thinking we might be able to leverage the subwizard instead to generate more steps as needed.  The key benefit to this would be users should be able to use the back button to return to a previous step.

src/api/v2/quickPickWizard/QuickPickAppResourceStep.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    private getQuickPickItem(resource: ApplicationResource): IAzureQuickPickItem<ApplicationResource> {
+        return {
+            label: resource.name,
+            data: resource,
+        };
+    }
+
+    private matchesAppResource(resource: ApplicationResource): boolean {
```
nturinski, line 67: Could we leverage some of the code in [`SubscriptionTreeItem.pickAppResource`](https://github.com/microsoft/vscode-azureresourcegroups/blob/main/src/tree/SubscriptionTreeItem.ts#L37-L83) or maybe extract it out as a util function?

I think the only reason it lives on the Sub treeItem is because it directly accesses the appResources cache, but definitely doesn't need to be that way.

src/api/v2/quickPickWizard/RecursiveQuickPickStep.ts
```diff
@@ -0,0 +1,24 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IWizardOptions } from '@microsoft/vscode-azext-utils';
+import { ResourceModelBase } from '../v2AzureResourcesApi';
+import { GenericQuickPickStep } from './GenericQuickPickStep';
+import { QuickPickWizardContext } from './QuickPickWizardContext';
+
+export class RecursiveQuickPickStep<TModel extends ResourceModelBase> extends GenericQuickPickStep<TModel> {
```
nturinski, line 11: Since this inherits `getPicks` and `prompt` from `GenericQuickPickStep`, would the `getSubWizard` interfere with this [block of code?](https://github.com/microsoft/vscode-azureresourcegroups/pull/328/files#diff-fafee4fc40399ada1bd6f49e86a9eff708bdf1230ada0e0b5614c76a03f72ac5R48-R54)

src/api/v2/quickPickWizard/SkipIfOneQuickPickStep.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { ResourceModelBase } from '../v2AzureResourcesApi';
+import { QuickPickWizardContext } from './QuickPickWizardContext';
+import { RecursiveQuickPickStep } from './RecursiveQuickPickStep';
+
+export class SkipIfOneQuickPickStep<TModel extends ResourceModelBase> extends RecursiveQuickPickStep<TModel> {
+    public override async prompt(wizardContext: QuickPickWizardContext<TModel>): Promise<void> {
+        const picks = await this.getPicks(wizardContext);
+
+        if (picks.length === 1) {
```
nturinski, line 14: I mentioned some other criterion for skipping in a prior comment

src/api/v2/quickPickWizard/ContextValueFilter.ts
```diff
@@ -0,0 +1,27 @@
...
+        return false;
+    }
+
+    const filterArray = Array.isArray(filter) ? filter : [filter];
+
+    return filterArray.some(filter => {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        return resource.quickPickOptions!.contexts.some(contextValue => {
+            if (typeof filter === 'string') {
+                return filter === contextValue;
```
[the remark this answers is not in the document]
alexweininger: I thought the idea was that `contexts` was an array of the context values split up. Like `['azResource', 'contextValue1', 'contextValue2']`. So if you wanted to select resources with a context of `azResource;contextValue1` you'd pass in `['azResource', 'contextValue1']`. This works since the ordering doesn't really matter (we just sort it alphabetically).

src/api/v2/quickPickWizard/CreateQuickPickStep.ts
alexweininger: We keep on running into this issue with type and kind all over the place. 

We should try to build on top of https://github.com/microsoft/vscode-azuretools/pull/1191 so that we don't have to solve the problem again here.

src/api/v2/quickPickWizard/QuickPickAppResourceStep.ts
nturinski: I guess that [this ](https://github.com/microsoft/vscode-azuretools/pull/1191/files?diff=split&w=1)is for?

src/api/v2/quickPickWizard/ContextValueFilter.ts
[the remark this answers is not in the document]
nturinski: Oh, yeah that's true. I forgot that this is its own model and is not using the `AzExtTreeItems` to get the context values.

src/api/v2/quickPickWizard/GenericQuickPickStep.ts
bwateratmsft: I think it always has to depend on the use case. For example, when choosing a resource to delete, we can _never_ skip a step.

src/api/v2/quickPickWizard/GenericQuickPickStep.ts
bwateratmsft: `RecursiveQuickPickStep` is that.

src/api/v2/quickPickWizard/QuickPickAppResourceStep.ts
bwateratmsft: Phil's code is different enough that I thought it made the most sense to just write from scratch.

src/api/v2/quickPickWizard/QuickPickAppResourceStep.ts
```diff
@@ -0,0 +1,94 @@
...
+        const bdp = this.branchDataProviderManager.getApplicationResourceBranchDataProvider(wizardContext.applicationResource);
+        wizardContext.currentNode = await Promise.resolve(bdp.getResourceItem(wizardContext.applicationResource)) as TModel;
+    }
+
+    public shouldPrompt(_wizardContext: QuickPickAppResourceWizardContext<TModel>): boolean {
+        return true;
+    }
+
+    public async getSubWizard(wizardContext: QuickPickAppResourceWizardContext<TModel>): Promise<IWizardOptions<QuickPickAppResourceWizardContext<TModel>> | undefined> {
+        if (this.options.expectedChildContextValue) {
```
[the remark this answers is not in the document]
bwateratmsft: It does make sense. My motivation for this was to keep the workload of changing existing code lower--i.e. rather than needing to construct a wizard, they can just continue using resource type filter + optional expected child context value.

src/api/v2/quickPickWizard/SkipIfOneQuickPickStep.ts
bwateratmsft: I think it needs to be determined by the consumer and vary depending on the scenario.

alexweininger:
I did some refactoring of the filters, and made a draft PR against this draft PR...

https://github.com/microsoft/vscode-azureresourcegroups/pull/333

src/api/v2/quickPickWizard/GenericQuickPickStep.ts
alexweininger: Yeah I think I'd want to keep the skipping logic here to a minimum for the sake of simplicity too.

> This isn't a matching context value

This doesn't sound like a case where we could skip. More like give them the option to create whatever we're looking for, or throw a "not found" error, right?

> AND There is only one available option

If there is only one available option, we still may want to show the "+ Create X..." item. I'm also not a huge fan of the "skip if only 1" idea because it could confuse users.

> OR the user has specified what should be selected as the currentNode at this point

Not sure what you mean by this one.

bwateratmsft:
Closing in favor of https://github.com/microsoft/vscode-azuretools/pull/1229

Issue #298, opened 2022-06-08
Creating a Function App throws "The api-version '2021-01-01' is invalid." error.


bpsims:
Windows 10
Azure Stack

After signing, click "Create Resource" icon, select "Create Function App in Azure...", enter name and runtime stack.  
Immediately throws:

**The api-version '2021-01-01' is invalid. The supported versions are '2020-06-01,2020-05-01,2020-01-01,2019-11-01,2019-10-01,2019-09-01,2019-08-01,2019-07-01,2019-06-01,2019-05-10,2019-05-01,2019-03-01,2018-11-01,2018-09-01,2018-08-01,2018-07-01,2018-06-01,2018-05-01,2018-02-01,2018-01-01,2017-12-01,2017-08-01,2017-06-01,2017-05-10,2017-05-01,2017-03-01,2016-09-01,2016-07-01,2016-06-01,2016-02-01,2015-11-01,2015-01-01,2014-04-01-preview,2014-04-01,2014-01-01,2013-03-01,2014-02-26,2014-04,2015-11-01,2015-11-01'.**

alexsmithMSFT:
For App Services on Azure Stack Hub, the extension should use **2020-09-01-hybrid** API profile. For more details onsupported api-versions, see- https://docs.microsoft.com/azure-stack/user/azure-stack-version-profiles#azure-api-profiles-and-azure-stack-hub-compatibility
Specifically for App Services (Microsoft.Web provider)- 
https://docs.microsoft.com/azure-stack/user/azure-stack-profiles-azure-resource-manager-versions#microsoftweb
And implementation in .NET- https://docs.microsoft.com/azure-stack/user/azure-stack-version-profiles-net#profiles

alexsmithMSFT:
This issue might be better handled over in https://github.com/microsoft/vscode-azureappservice/issues

Either way, the API profile for AzureCustomCloud option should be locked into one of the hybrid profiles that is supported on Azure Stack Hub, like in https://github.com/microsoft/vscode-azureresourcegroups/pull/67

PR #329, opened 2022-07-11
Upgrade 2020-09-01-hybrid sdk to T2

### LINKED ISSUE #296: g[Symbol.iterator] is not a function error when expanding Subscription
After signing into Azure, expanding the Subscription name will throw a "g[Symbol.iterator] is not a function" error and no resources will load under the headings.  I believe this started happening within the past week. 

Windows 10
Azure Stack tenant

Another error shows up in the console:
ERR [Extension Host] (node:45396) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.

alexweininger:
Fixes #296

[View changelog](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/resources/arm-resources-profile-2020-09-01-hybrid/CHANGELOG.md#200-2022-03-08)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
             "license": "SEE LICENSE IN LICENSE.md",
             "dependencies": {
                 "@azure/arm-resources": "5.0.0",
-                "@azure/arm-resources-profile-2020-09-01-hybrid": "^1.0.0",
+                "@azure/arm-resources-profile-2020-09-01-hybrid": "^2.0.0",
                 "@microsoft/vscode-azext-azureutils": "^0.3.1",
                 "@microsoft/vscode-azext-utils": "^0.3.7",
                 "fs-extra": "^8.1.0",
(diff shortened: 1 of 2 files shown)


bwateratmsft (approved):

alexweininger:
> We may also need to do [microsoft/vscode-azuretools#1180](https://github.com/microsoft/vscode-azuretools/pull/1180) as well. Though I'm not entirely sure these `-profile-2020-09-01-hybrid` packages are even needed anymore...

We need these packages to support custom clouds, which includes Azure Stack. We actually need to add these packages to the Functions and App Service extensions in the future so those extensions support Azure Stack too. 

Related:
https://github.com/microsoft/vscode-azurefunctions/issues/2259 
https://github.com/microsoft/vscode-azureresourcegroups/issues/298

alexsmithMSFT:
VSCode extensions for Azure Account, Resources, Storage, and VM are all officially supported- https://docs.microsoft.com/azure-stack/user/azure-stack-dev-start-vscode-azure
Using a supported version of "hybrid" API profiles- https://docs.microsoft.com/azure-stack/user/azure-stack-profiles-azure-resource-manager-versions
As already mentioned by @alexweininger it would also be great to have working extensions for other resources available on Stack Hub, including App Services and Functions.

PR #332, opened 2022-07-12
Fix displaying Logic Apps as Function Apps

### LINKED ISSUE #246: [Suggestion] Show Logic Apps using Logic App Icon rather than Function App Icon
This would match the Azure Logic Apps extensions too

alexweininger:
Fixes #246 

* Adds logic app icon and display name
* Won't show up in App Service or Functions resource type groups


src/utils/azureUtils.ts
```diff
@@ -227,22 +223,40 @@ const supportedTypes: SupportedTypeMap = {
...
-        if (kind?.toLowerCase().includes('functionapp')) {
+        if (kind?.toLowerCase().includes('functionapp') && kind?.toLowerCase().includes('workflowapp')) {
             return true;
         }
     }
     return false;
 }
 
+export function isAppServiceApp(resource: GenericResource): boolean {
+    return !isFunctionApp(resource) && !isLogicApp(resource);
```
bwateratmsft, line 250: This should include a resource type check, or else it would return true for everything that is not a function or logic app (e.g., true for a storage account).

Something like:
```suggestion
    return resource.type.toLowerCase() === 'microsoft.web/sites' &&
        !isFunctionApp(resource) &&
        !isLogicApp(resource);
```

Also, doesn't need to be in this PR but it's increasingly clear we need to define our own resource type enumeration--which has separate entries for App Service, Functions, and Logic Apps--and have a single function in utils to normalize `GenericResource` -> that type. Then we can stop needing all these `toLowerCase()` and type/kind checks all over.

src/utils/azureUtils.ts
```diff
@@ -227,22 +223,40 @@ const supportedTypes: SupportedTypeMap = {
     'microsoft.web/serverfarms': { displayName: localize('serverFarms', 'App Service plans') },
     'microsoft.web/kubeenvironments': { displayName: localize('containerService', 'App Service Kubernetes Environment') },
     'microsoft.app/managedenvironments': { displayName: localize('containerAppsEnv', 'Container Apps Environment') },
-    'microsoft.app/containerapps': { displayName: localize('containerApp', 'Container Apps') },
+    'microsoft.app/containerapps': { displayName: localize('containerApp', 'Container Apps') }
```
bwateratmsft, line 226: Dunno if this comma was removed intentionally but I like to leave them in arrays and objects, since in the future it will mean one less line changed when adding new rows.
```suggestion
    'microsoft.app/containerapps': { displayName: localize('containerApp', 'Container Apps') },
```

alexweininger: I prefer trailing commas too, I think my formatter/ES Lint removed it. I'll have to try to turn that off

PR #337, opened 2022-07-14
Use `AzExtResourceType` to simplify `azureUtils.ts`

alexweininger:
Depends on https://github.com/microsoft/vscode-azuretools/pull/1191

Each icon now maps to an `AzExtResourceType`.

edit:

Now this depends on https://github.com/microsoft/vscode-azuretools/pull/1222


src/azureExtensions.ts
```diff
@@ -3,19 +3,16 @@
...
+import { AzExtResourceType } from "@microsoft/vscode-azext-utils/azExtResourceType";
+import { AppResource } from "@microsoft/vscode-azext-utils/hostapi";
 import { localize } from "./utils/localize";
 
 export const azureExtensions: IAzExtMetadata[] = [
     {
         name: 'vscode-azurefunctions',
         label: 'Functions',
         resourceTypes: [
-            {
```
bwateratmsft, line 15: These changes will affect the extension activation manager, since the Azure resource type is used as the qualifier for activation. I agree that the `resourceType` field should be changed here as you have it, but I think we need to also include a separate `azureResourceType` (or some name along those lines), that the extension activation manager can use, _in addition_ to using the new `AzExtResourceType`--so the activation manager would activate _both_ (for now). Eventually we can remove the Azure resource type from here and the activation manager when all of the extensions have shipped.

Unless we're absolutely sure we're going to synchronously ship all of them again...

src/utils/azureUtils.ts
```diff
@@ -115,115 +100,43 @@ export async function getArmTagKeys(context: IActionContext): Promise<Set<string
...
-    'microsoft.documentdb/databaseaccounts': { displayName: localize('documentDB', 'Azure Cosmos DB') },
-    'microsoft.operationalinsights/workspaces': { displayName: localize('operationalInsightsWorkspaces', 'Operational Insights workspaces') },
-    'microsoft.operationsmanagement/solutions': { displayName: localize('operationsManagementSolutions', 'Operations management solutions') },
-    'microsoft.insights/components': { displayName: localize('insightsComponents', 'Application Insights') },
-    'microsoft.web/serverfarms': { displayName: localize('serverFarms', 'App Service plans') },
-    'microsoft.web/kubeenvironments': { displayName: localize('containerService', 'App Service Kubernetes Environment') },
-    'microsoft.app/managedenvironments': { displayName: localize('containerAppsEnv', 'Container Apps Environment') },
-    'microsoft.app/containerapps': { displayName: localize('containerApp', 'Container Apps') },
+const azExtDisplayInfo: Partial<Record<AzExtResourceType, AzExtResourceTypeDisplayInfo>> = {
+    'ApplicationInsights': { displayName: localize('insightsComponents', 'Application Insights') },
```
bwateratmsft, line 108: Does TS not like this?
```typescript
AzExtResourceType.ApplicationInsights: { displayName: /* ... */ },
```

src/utils/azureUtils.ts
```diff
@@ -115,115 +100,43 @@ export async function getArmTagKeys(context: IActionContext): Promise<Set<string
...
-    'microsoft.sql/servers': { displayName: localize('sqlServers', 'SQL servers') },
-    'microsoft.documentdb/databaseaccounts': { displayName: localize('documentDB', 'Azure Cosmos DB') },
-    'microsoft.operationalinsights/workspaces': { displayName: localize('operationalInsightsWorkspaces', 'Operational Insights workspaces') },
-    'microsoft.operationsmanagement/solutions': { displayName: localize('operationsManagementSolutions', 'Operations management solutions') },
-    'microsoft.insights/components': { displayName: localize('insightsComponents', 'Application Insights') },
-    'microsoft.web/serverfarms': { displayName: localize('serverFarms', 'App Service plans') },
-    'microsoft.web/kubeenvironments': { displayName: localize('containerService', 'App Service Kubernetes Environment') },
-    'microsoft.app/managedenvironments': { displayName: localize('containerAppsEnv', 'Container Apps Environment') },
-    'microsoft.app/containerapps': { displayName: localize('containerApp', 'Container Apps') },
+const azExtDisplayInfo: Partial<Record<AzExtResourceType, AzExtResourceTypeDisplayInfo>> = {
```
bwateratmsft, line 107: Is `Partial` needed? Or is that because not all `AzExtResourceType`s are included?

alexweininger: > because not all AzExtResourceTypes are included?

Correct

src/azureExtensions.ts
alexweininger: We should discuss this in standup.

There are a few more breaking changes I'd like to make based on `AzExtResourceType`. So I was hoping we could sim-ship the extensions.

I'll look into making the changes backwards compatible too.

alexweininger: > These changes will affect the extension activation manager, since the Azure resource type is used as the qualifier for activation.

I agree we should change the extension activation manager (EAM) to use `AzExtResourceType`. However, these changes don't actually break the existing EAM implementation because each extension [declares the Azure type in the the manifest](https://github.com/microsoft/vscode-azureresourcegroups/blob/2a19e3397229f75e95eb39fcbfa5d032c7b32446/src/utils/ExtensionActivationManager.ts#L56-L57).

bwateratmsft: *smack* Oh duh! Yeah. we should try to set up EAM to try waking up both the old `microsoft.blah/blah` types and the new `AzExtResourceType`, over time we can convert the extensions to have the new items.

src/AzureResourceProvider.ts
```diff
@@ -29,6 +30,7 @@ export class AzureResourceProvider implements AppResourceProvider {
             type: nonNullProp(resource, 'type'),
             kind: resource.kind,
             location: resource.location,
+            azExtResourceType: getAzExtResourceType({ type: nonNullProp(resource, 'type'), kind: resource.kind }) ?? resource.type as AzExtResourceType,
```
bwateratmsft, line 33: This bit is kinda concerning:

> `?? resource.type as AzExtResourceType`

I brought that up yesterday as an option but I can see it coming back to bite us.

alexweininger:
I wasn't able to get the import from `@microsoft/vscode-azext-utils/getAzExtResourceType` to work at runtime so I had to make some changes to utils. I also fixed a bad bug with `getAzExtResourceType`. This PR now depends on https://github.com/microsoft/vscode-azuretools/pull/1222.

src/AzureResourceProvider.ts
alexweininger: I agree that it could cause issues, but I'm not sure what alternatives there are. We could add an `Unknown` `AzExtResourceType`.

bwateratmsft: Is it really so bad if it returns undefined? We could alter the API to note that it's not always defined in the `AppResource`.

alexweininger: No, I think making `azExtResourceType` an optional property on `AppResource` is a good idea

alexweininger: Ok, `azExtResourceType` is now optional.

bwateratmsft (approved):

bwateratmsft (approved):

PR #339, opened 2022-07-21
When sub or group is refreshed, refresh the children as well

nturinski:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/1096

When "refresh" is triggered on the root or a parent, it will only call its refresh, and then call `loadMoreChildren` on itself.  However, because `loadMoreChildrenImpl` doesn't actually create new tree items (since we are caching all of the `AppResourceTreeItems`), refreshing the root, the subscription, and the groups didn't actually do anything.

By adding `refreshImpl` to `SubscriptionTreeItem` and `GroupTreeItemBase`, the `AppResourceTreeItems` will actually get refreshed now and it'll be a meaningful refresh.

I can't fix the root level refresh without affecting performance because if I call refresh in `SubscriptionTreeItem.loadMoreChildrenImpl`, it will activate all of the extensions at once. 

The root level refresh only works for subscription filtering so I would almost opt to get rid of the refresh button on the ribbon and add it to the groups themselves. It would looks like this:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -332,14 +332,19 @@
                     "group": "9@1"
                 },
                 {
+                    "command": "azureResourceGroups.refresh",
+                    "when": "view == azureResourceGroups && viewItem =~ /group/",
+                    "group": "inline@1"
+                },
+                {
                     "command": "azureResourceGroups.focusGroup",
                     "when": "view == azureResourceGroups && viewItem =~ /(group).*(unfocused)/",
-                    "group": "inline"
+                    "group": "inline@2"
                 },
                 {
                     "command": "azureResourceGroups.unfocusGroup",
                     "when": "view == azureResourceGroups && viewItem =~ /(focused).*(group)/",
-                    "group": "inline"
+                    "group": "inline@2"
                 }
             ],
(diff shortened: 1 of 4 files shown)


bwateratmsft:
@alexweininger didn't #294 do this?

alexweininger:
> didn't #294 do this?

The logic added in 294 only re-resolves if a new resolver has been registered that matches the resource

bwateratmsft (approved):

alexweininger:
> The root level refresh only works for subscription filtering so I would almost opt to get rid of the refresh button on the ribbon

I don't think I'm ready to get rid of the refresh button on the ribbon... but you make a good point here. It really doesn't do much currently

nturinski:
> > The root level refresh only works for subscription filtering so I would almost opt to get rid of the refresh button on the ribbon
> 
> I don't think I'm ready to get rid of the refresh button on the ribbon... but you make a good point here. It really doesn't do much currently

I think when we move to v2, we could probably fix it not refreshing anything so probably good to just leave it for now.

PR #341, opened 2022-07-22
Make revealResource accept an Azure Id

alexweininger:
See [this comment](https://github.com/microsoft/vscode-azuretools/pull/1191#discussion_r927181841) for context on this change. The summary is that revealResource really only needs the Azure Id and not the full `AppResource` interface.

Depends on https://github.com/microsoft/vscode-azureresourcegroups/pull/340

### DIFF
--- a/src/commands/revealResource.ts
+++ b/src/commands/revealResource.ts
@@ -3,19 +3,19 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { parseAzureResourceId } from '@microsoft/vscode-azext-azureutils';
 import { IActionContext, parseError } from '@microsoft/vscode-azext-utils';
 import { AppResource } from '@microsoft/vscode-azext-utils/hostapi';
 import { revealTreeItem } from '../api/revealTreeItem';
 import { ext } from '../extensionVariables';
 import { AppResourceTreeItem } from '../tree/AppResourceTreeItem';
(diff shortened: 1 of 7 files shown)


bwateratmsft (approved):

src/commands/revealResource.ts
```diff
@@ -11,12 +11,16 @@ import { ext } from '../extensionVariables';
 import { AppResourceTreeItem } from '../tree/AppResourceTreeItem';
 import { SubscriptionTreeItem } from '../tree/SubscriptionTreeItem';
 
-export async function revealResource(context: IActionContext, resource: AppResource): Promise<void> {
-    context.telemetry.properties.resourceType = resource.type?.replace(/\//g, '|'); // Replace the slashes otherwise this gets redacted because it looks like a user file path
-    context.telemetry.properties.resourceKind = resource.kind;
+export async function revealResource(context: IActionContext, resourceId: string): Promise<void>;
+export async function revealResource(context: IActionContext, resource: AppResource): Promise<void>;
+export async function revealResource(context: IActionContext, arg: AppResource | string): Promise<void> {
```
nturinski, line 16: One thing I don't really understand is what's the point of exporting 3 different function definitions when `AppResource | string` covers the other cases? Is it just to see the arg names (`resourceId` and `resource`?)

alexweininger: Yeah that way you get the two declarations, instead of one with just `AppResource | string`.

PR #348, opened 2022-07-27
v0.5.5 release prep

alexweininger:


CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 ## Unreleased
```
nturinski, line 3: Have we not been removing this between releases? Should we?

nturinski (approved):

CHANGELOG.md
alexweininger: We could remove it since we don't use it, but I also wish we used it...

alexweininger: I'll try out using it for the next release cycle and see how it goes. If we don't like it we'll remove it.

It's nice to keep a log of the changes as we make them so that we don't miss anything important when it comes to releasing. Also makes it easier on the person piloting the release 😄

nturinski: > We could remove it since we don't use it, but I also wish we used it...

How hard would it be to write two scripts:
- Adds ##Unreleased to CHANGELOG at the same time the package.json is bumped
- Add the commit message as a bullet point under ##Unreleased after each merge

alexweininger: > Adds ##Unreleased to CHANGELOG at the same time the package.json is bumped
Add the commit message as a bullet point under ##Unreleased after each merge

Both are possible and can be easily accomplished using existing tools. There's even one built and maintained by [Microsoft called beachball](https://microsoft.github.io/beachball/)

nturinski (approved):

PR #357, opened 2022-08-10
Fix Workspace view loadMore

alexweininger:
Fixes https://github.com/microsoft/vscode-azurestorage/issues/1046

### DIFF
--- a/src/commands/registerCommands.ts
+++ b/src/commands/registerCommands.ts
@@ -47,6 +47,8 @@ export function registerCommands(): void {
     registerReportIssueCommand('azureResourceGroups.reportIssue');
     registerCommand('azureResourceGroups.createResource', createResource);
     registerCommand('azureResourceGroups.refreshWorkspace', refreshWorkspace);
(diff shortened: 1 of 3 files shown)


bwateratmsft (dismissed):

src/commands/registerCommands.ts
```diff
@@ -47,6 +47,8 @@ export function registerCommands(): void {
     registerReportIssueCommand('azureResourceGroups.reportIssue');
     registerCommand('azureResourceGroups.createResource', createResource);
     registerCommand('azureResourceGroups.refreshWorkspace', refreshWorkspace);
+    registerCommand('azureWorkspace.loadMore', async (context: IActionContext, node: AzExtTreeItem) => await ext.workspaceTree.loadMore(node, context));
```
nturinski, line 50: None of the other workspace commands use the pattern of the prefix being `azureWorkspace`. I'm fine with it, but we should be consistent.

alexweininger: Good point. I changed the other workspace command (refresh) to follow the same convention.

bwateratmsft (approved):

PR #358, opened 2022-08-10
Start implementation of compat shim

bwateratmsft:
Making as a draft PR to gather feedback.


src/tree/v2/providers/CompatibleBranchDataProvider.ts
```diff
@@ -0,0 +1,69 @@
...
+        super(undefined as unknown as AzExtParentTreeItem, loadMoreCommandId);
+    }
+
+    //#region TreeDataProvider
+
+    public override readonly onDidChangeTreeData = this.overrideOnDidChangeTreeDataEmitter.event;
+
+    public override getParent(treeItem: TModel): Promise<TModel> {
+        // This method is redeclared to make TypeScript happier, but it's no more than a super call with extra casts
+        return super.getParent(treeItem) as Promise<TModel>;
```
bwateratmsft, line 24: This is operating on the assumption that if the parent of the AppResource node itself is requested, that RG will trap and answer that--because it _is_ technically the owner of that node.

src/tree/v2/providers/CompatibleBranchDataProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+    //#region BranchDataProvider
+
+    public async getResourceItem(element: TResource): Promise<TModel> {
+        const oldAppResource: AppResource = {
+            ...element,
+            type: element.type.type,
+            kind: element.type.kinds?.join(';'),
+        };
+
+        return this.resolver.resolveResource(/* TODO */'foo', oldAppResource) as Promise<TModel>;
```
alexweininger, line 43: Do we have a plan for how subscriptions will be passed down yet?

bwateratmsft: @philliphoff has been playing around with the VSCode auth API, he could tell you more.

philliphoff: Right now I'm working toward having application resources contain a subscription property (like they do today), where the subscription object contains a `getSession()` method that returns a VS Code authentication session.  Whether the session is an *actual* VS Code authentication session or just a wrapper around the existing Azure Account API is to be seen. It'll likely start as the latter, but this allows us to eventually move to VS Code's authentication system under the covers (e.g. when ADAL/MSAL limitations get resolved, etc.).  This also insulates our API from T1, T2, ..., Tn credentials changes.  That is, the intent is that the Azure Utils library would have convenience methods to convert a VS Code `AuthenticationSession` to a `TokenCredential` or whatever the Azure SDKs need in the future, without requring us to update our API.

```typescript
export interface ApplicationAuthentication {
    getSession(scopes?: string[]): vscode.ProviderResult<vscode.AuthenticationSession>;
}

/**
 * Information specific to the Subscription
 */
export interface ApplicationSubscription {
    readonly authentication: ApplicationAuthentication;
    readonly displayName: string;
    readonly subscriptionId: string;
    readonly environment: Environment;
    readonly isCustomCloud: boolean;
}

/**
 * Represents an individual resource in Azure.
 * @remarks The `id` property is expected to be the Azure resource ID.
 */
export interface ApplicationResource extends ResourceBase {
    readonly subscription: ApplicationSubscription;
    readonly type: ApplicationResourceType;
    readonly location?: string;
    readonly resourceGroup?: string;
    /** Resource tags */
    readonly tags?: {
    };
    /* add more properties from GenericResource if needed */
}
```

Note that we don't just expose the `AuthenticationSession`, even though that should generally work, instead having an intermediate `getSession()` function.  This is because of the way `TokenCredential.getToken()` works where new scopes may be injected, whereas VS Code tends to expect that scopes are all defined up front.  That said, we *generally* know what scopes will be needed, so we can "pre-authenticate" with VS Code with those scopes and keep prompts to a minimum.

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -82,6 +82,21 @@ export interface WrappedResourceModel {
  * The interface that resource resolvers must implement
  */
 export interface BranchDataProvider<TResource extends ResourceBase, TModel extends ResourceModelBase> extends vscode.TreeDataProvider<TModel> {
+    /**
+     * @deprecated This is implemented by the Resources API and does not need to be implemented by the
+     * {@link BranchDataProvider}.
+     */
+    getParent?: never;
```
bwateratmsft, line 89: In line with my suggestion, we can override with this `getParent?: never` to indicate that `BranchDataProvider`s do not need to implement `getParent` (and, _should_ not, to avoid confusion).

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -82,6 +82,21 @@ export interface WrappedResourceModel {
...
+
+    /**
+     * Get the children of `element`. This differs from standard {@link vscode.TreeDataProvider}
+     * in that the `element` parameter will never be undefined.
+     *
+     * @param element The element from which the provider gets children. Cannot be `undefined`.
+     * @return Children of `element`.
+     */
+    getChildren(element: TModel): TModel[] | Thenable<TModel[]>;
```
bwateratmsft, line 98: In line with my suggestion, we can override with this `getChildren(element: TModel)` to indicate that `element` will never be `undefined`.

src/tree/v2/providers/CompatibleBranchDataProvider.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    public async getResourceItem(element: TResource): Promise<TModel> {
+        const oldAppResource: AppResource = {
+            ...element,
+            type: element.type.type,
+            kind: element.type.kinds?.join(';'),
+        };
+        const subscriptionContext: ISubscriptionContext = createSubscriptionContext(element.subscription);
+
+        return await this.resolver.resolveResource(subscriptionContext, oldAppResource) as Promise<TModel>;
```
alexweininger, line 45: This is going to return a `ResolvedAppResourceBase` which is not an `AzExtTreeItem` by itself. Currently we "combine" a `ResolvedAppResourceBase` with an `AppResourceTreeItem` using a `Proxy`.

I think we'll need to make a compat version of this and return it here, because just returning a `ResolvedAppResourceBase` doesn't have the necessary properties fro the `AzExtTreeDataProvider` to turn it into a `vscode.TreeItem`.

bwateratmsft: Yeah. I think we'll have to evaluate what parts of `AzExtTreeItem` are needed and include them.

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -71,19 +73,34 @@ export interface ResourceModelBase {
 /**
  * Represents a branch data provider resource model as returned by a context menu command.
  */
-export interface WrappedResourceModel {
+export interface Box {
```
bwateratmsft, line 76: Ultimately this will be in `@microsoft/vscode-azext-utils`.

alexweininger:
We're currently storing branch data providers in a map using the Id passed during registration as the key. Currently, that Id wasn't used for anything, but now we are using that Id to lookup the BDP for a given application resource.

[BranchDataProviderManager.ts where we find a branch data provider for a given application resource](https://github.com/microsoft/vscode-azureresourcegroups/blob/bmw/philliphoff-resource-api-compatshim/src/tree/v2/providers/BranchDataProviderManager.ts#L45-L58)

Two things:
1. We should be mapping them by `AzExtResourceType` rather than `type` since `type` has some overlap
2. We'll have to update the Ids we're passing to `registerAppResourceResolver` in the client extensions, since they are random (some pass in the extensionId, others pass in resource type). We'll update them to use `AzExtResourceType`.

src/api/v2/compatibility/CompatibleAzExtTreeDataProvider.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export class CompatibleAzExtTreeDataProvider extends IntermediateCompatibleAzExtTreeDataProvider {
+    public constructor(private readonly tdp: ResourceTreeDataProviderBase) {
+        super({} as unknown as AzExtParentTreeItem, undefined as unknown as string);
+    }
+
+    //#region Things that should not be called
+    public override trackTreeItemCollapsibleState(_treeView: TreeView<AzExtTreeItem>): Disposable {
+        throw new Error('This method should never be called');
```
nturinski, line 29: I know that these are supposed to never be called, but would it be worth outputting what method or accessor was called to throw the error, if an internal error ever does happen?

nturinski (approved):
I didn't notice anything weird or out of place (besides some awful things we had to do to make `CompatiableBranchDataProviderBase` work).

I left some comments, but more to just gain more context. I pointed them out in-case they were mistakes, but I doubt that they were.

package.json
```diff
@@ -600,7 +565,7 @@
         "@azure/arm-resources": "5.0.0",
         "@azure/arm-resources-profile-2020-09-01-hybrid": "^2.0.0",
         "@microsoft/vscode-azext-azureutils": "^0.3.4",
-        "@microsoft/vscode-azext-utils": "^0.3.14",
+        "@microsoft/vscode-azext-utils": "file:../vscode-azuretools/utils/microsoft-vscode-azext-utils-0.3.15.tgz",
```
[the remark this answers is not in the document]
bwateratmsft: We should release it, merging with a local package produces a nonfunctional build which is not something I like to have in any important branches

src/api/AzureResourceGroupsExtensionApi.ts
```diff
@@ -8,13 +8,13 @@ import { Activity, AppResourceResolver, AzureHostExtensionApi, AzureResourceGrou
 import { Disposable, TreeView } from 'vscode';
 
 export class InternalAzureResourceGroupsExtensionApi implements AzureHostExtensionApi, AzureResourceGroupsExtensionApi {
-    public static apiVersion = '0.0.1';
+    public static apiVersion: '0.0.1' = '0.0.1';
```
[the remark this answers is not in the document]
bwateratmsft: The RG extension needs to expose two API versions; one that has the same major version as the old ones--so compat extensions continue to work as they do now--and one with a new major version. The implementation under the hood of the both of those APIs will be (mostly) the same, but the client extensions do not need to know that.

src/tree/v2/BranchDataProviderItem.ts
```diff
@@ -48,11 +39,11 @@ export class BranchDataProviderItem implements ResourceGroupsItem, WrappedResour
 
         return {
             ...this.options?.defaults ?? {},
-            ...treeItem
+            ...treeItem,
```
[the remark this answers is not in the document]
bwateratmsft: I like trailing commas on objects. That way if you have to add a line there's only one line edit.

src/api/v2/compatibility/CompatibleAzExtTreeDataProvider.ts
bwateratmsft: I think that's a good idea. Call stacks frequently end up being useless thanks to webpack.

src/api/pickAppResource.ts
```diff
@@ -3,18 +3,43 @@
...
-import { AzExtTreeItem, ITreeItemPickerContext } from "@microsoft/vscode-azext-utils";
+import { AzExtResourceType, AzExtTreeItem, compatibilityPickAppResourceExperience, ContextValueFilter, getAzExtResourceType, ITreeItemPickerContext } from "@microsoft/vscode-azext-utils";
 import { PickAppResourceOptions } from "@microsoft/vscode-azext-utils/hostapi";
 import { ext } from "../extensionVariables";
-import { SubscriptionTreeItem } from "../tree/SubscriptionTreeItem";
 
 export async function pickAppResource<T extends AzExtTreeItem>(context: ITreeItemPickerContext, options?: PickAppResourceOptions): Promise<T> {
-    const subscription: SubscriptionTreeItem = await ext.appResourceTree.showTreeItemPicker(SubscriptionTreeItem.contextValue, context);
-    const appResource = await subscription.pickAppResource(context, options);
+    return await compatibilityPickAppResourceExperience<T>(context, ext.v2.api.applicationResourceTreeDataProvider, {
```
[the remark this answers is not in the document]
alexweininger: When we upgrade to v2, we can use the non-compat `pickAppResource` experience. I've also been working on some cooler more customizable APIs, but they aren't completely finalized yet.

src/api/pickAppResource.ts
```diff
@@ -3,18 +3,43 @@
...
+
+function filterMap<T, TMapped>(
+    source: T[],
+    predicateMapper: (item: T, index: number) => TMapped | null | undefined,
+): TMapped[] {
+    let index = 0;
+    return source.reduce<TMapped[]>((accumulator, current) => {
+        const mapped = predicateMapper(current, index++);
+        // handles null or undefined
+        // eslint-disable-next-line eqeqeq
```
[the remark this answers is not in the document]
alexweininger: I know it's weird. But not a typo :D

src/api/registerWorkspaceResourceProvider.ts
```diff
@@ -3,23 +3,31 @@
...
-        void refreshWorkspace(context);
+        void refreshWorkspace();
+
+        ext.v2.api.registerWorkspaceResourceProvider(new CompatibilityWorkspaceResourceProvider(resourceType, provider));
+        ext.v2.api.registerWorkspaceResourceBranchDataProvider(resourceType, new CompatibleWorkspaceResourceBranchDataProvider('azureWorkspace.loadMore') as unknown as BranchDataProvider<WorkspaceResource, AzExtTreeItem>)
 
         return new Disposable(() => {
             delete workspaceResourceProviders[resourceType];
-            void refreshWorkspace(context);
+            refreshWorkspace();
```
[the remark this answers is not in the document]
alexweininger: It's sync. Just fires an event emitter.

src/tree/AppResourceTreeItem.ts
```diff
@@ -39,11 +39,14 @@ export class AppResourceTreeItem extends ResolvableTreeItemBase implements Group
 
     private constructor(root: AzExtParentTreeItem, resource: AppResource) {
         super(root);
-        this.rootGroupTreeItem = root;
-        this.rootGroupConfig = <GroupNodeConfiguration><unknown>root;
+
+        if (root) {
```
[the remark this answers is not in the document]
alexweininger: I actually think these changes aren't needed anymore since this class shouldn't be used anymore. I will double check, but I think I'll revert these changes and then delete all these old classes in a follow up PR.

src/tree/v2/BranchDataProviderItem.ts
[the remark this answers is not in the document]
alexweininger: Me too!

src/api/registerWorkspaceResourceProvider.ts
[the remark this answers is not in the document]
nturinski: Then we should be able to remove the void above as well?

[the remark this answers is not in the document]
alexweininger: Yeah, I'll remove.

package.json
```diff
@@ -328,14 +308,9 @@
                     "when": "view == azureResourceGroups && viewItem == azureextensionui.azureSubscription",
                     "group": "9@2"
                 },
-                {
-                    "command": "azureResourceGroups.deleteResourceGroup",
-                    "when": "view == azureResourceGroups && viewItem =~ /azureResourceGroup/",
-                    "group": "1@1"
-                },
                 {
                     "command": "azureResourceGroups.deleteResourceGroupV2",
```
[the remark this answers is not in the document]
alexweininger: I'll remove the old delete command, and rename the new command in a future PR.

package.json
```diff
@@ -367,16 +342,6 @@
                     "command": "azureResourceGroups.refresh",
                     "when": "view == azureResourceGroups && viewItem =~ /group/",
                     "group": "inline@1"
-                },
```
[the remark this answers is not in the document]
alexweininger: 😭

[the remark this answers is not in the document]
bwateratmsft: 😈

src/api/AzureResourceGroupsExtensionApi.ts
```diff
@@ -41,19 +41,19 @@ export class InternalAzureResourceGroupsExtensionApi implements AzureHostExtensi
         return this.#appResourceTree;
     }
 
-    public get appResourceTreeView(): TreeView<AzExtTreeItem> {
+    public get appResourceTreeView(): TreeView<unknown> {
```
philliphoff, line 44: Won't users of `0.0.1` still expect `AzExtTreeItem`-derived views?  Or, is the compat shim replacing them with something that just looks like an `AzExTreeItem` but isn't?

philliphoff (commented):
Overall looks good; added a few comments.

src/api/v2/compatibility/CompatibleBranchDataProviderBase.ts
```diff
@@ -0,0 +1,71 @@
...
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AzExtParentTreeItem, AzExtTreeDataProvider, AzExtTreeItem, IFindTreeItemContext, ITreeItemPickerContext } from "@microsoft/vscode-azext-utils";
+import * as vscode from 'vscode';
+import { BranchDataProvider, ResourceBase, ResourceModelBase } from "../v2AzureResourcesApi";
+
+export class CompatibleBranchDataProviderBase<TResource extends ResourceBase, TModel extends AzExtTreeItem & ResourceModelBase> extends AzExtTreeDataProvider implements BranchDataProvider<TResource, TModel> {
+    protected readonly overrideOnDidChangeTreeDataEmitter = new vscode.EventEmitter<TModel | undefined>();
```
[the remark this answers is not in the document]
bwateratmsft: Actually I think this same change should be made in the base `AzExtTreeDataProvider`, with an override here to dispose of the additional `EventEmitter`.

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -260,6 +261,16 @@ export type WorkspaceResourceBranchDataProvider<TModel extends WorkspaceResource
  * The current (v2) Azure Resources extension API.
  */
 export interface V2AzureResourcesApi extends AzureResourcesApiBase {
+
+    applicationResourceTreeDataProvider: vscode.TreeDataProvider<ResourceGroupsItem>;
```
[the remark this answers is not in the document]
bwateratmsft: They'll be used for the quick pick stuff (https://github.com/microsoft/vscode-azuretools/pull/1229). In https://github.com/microsoft/vscode-azuretools/pull/1229/files#diff-694f3a35fbb1b75da1d7fae941610ecb074e247c59186fb8c7c85b5ecf69386bR64 we treat them as `<unknown>` so I think we can do that here.

src/api/registerApplicationResourceResolver.ts
```diff
@@ -3,21 +3,28 @@
...
 *--------------------------------------------------------------------------------------------*/
 
+import { AzExtResourceType, AzExtTreeItem } from "@microsoft/vscode-azext-utils";
 import { AppResourceResolver } from "@microsoft/vscode-azext-utils/hostapi";
 import { Disposable } from "vscode";
 import { ext } from "../extensionVariables";
+import { CompatibleApplicationResourceBranchDataProvider } from "./v2/compatibility/application/CompatibleApplicationResourceBranchDataProvider";
+import { ApplicationResource, BranchDataProvider } from "./v2/v2AzureResourcesApi";
 
 export const applicationResourceResolvers: Record<string, AppResourceResolver> = {};
```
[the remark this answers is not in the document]
alexweininger: Yeah we can improve the types here, but it has to be `Partial<Record<AzExtResourceType, AppResourceResolver>>`.

src/tree/v2/application/GroupingItem.ts
```diff
@@ -31,7 +31,13 @@ export class GroupingItem implements ResourceGroupsItem {
         private readonly contextValues: string[] | undefined,
         private readonly iconPath: TreeItemIconPath | undefined,
         public readonly label: string,
-        public readonly resources: ApplicationResource[]) {
+        public readonly resources: ApplicationResource[],
+        public readonly resourceType: string | undefined
+    ) {
+    }
+
+    public get subscription(): ISubscriptionContext {
```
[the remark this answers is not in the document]
alexweininger: This is for compatibility with the commands that are attached to the grouping tree items. They rely on this property.
image

src/tree/v2/BranchDataProviderItem.ts
```diff
@@ -48,6 +48,10 @@ export class BranchDataProviderItem implements ResourceGroupsItem, WrappedResour
 
     readonly portalUrl: vscode.Uri | undefined = this.options?.portalUrl;
 
+    get resource(): ApplicationResource {
```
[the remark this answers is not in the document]
alexweininger: This is only used by the QuickPickAppResource step in utils. Where it attaches the `resource` and `resource.resourceGroup` to the wizard context after an item is picked by the user.

The property only needs to be defined on application resource items. I've proposed changes to make this less clunky. https://github.com/microsoft/vscode-azureresourcegroups/pull/432

src/commands/createResource.ts
```diff
@@ -7,14 +7,14 @@
...
 interface AzExtCreateResourceCommand extends Command {
     title: string;
     detail?: string;
 }
 
-export async function createResource(context: IActionContext, node?: SubscriptionTreeItem): Promise<void> {
+export async function createResource(context: IActionContext, node?: SubscriptionItem): Promise<void> {
     const all = extensions.all;
 
     const extCommands = all.map((azExt) => azExt.packageJSON?.contributes?.[contributesKey]?.commands as unknown).filter((value) => value !== undefined);
```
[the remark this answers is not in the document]
alexweininger: I'm going to revert the changes made to this file and update the create command properly in a separate PR.

src/api/registerApplicationResourceResolver.ts
```diff
@@ -3,21 +3,30 @@
...
 
-    applicationResourceResolvers[id] = resolver;
-    ext.emitters.onDidRegisterResolver.fire(resolver);
+        applicationResourceResolvers[type] = resolver;
+        ext.emitters.onDidRegisterResolver.fire(resolver);
 
-    return new Disposable(() => {
-        delete applicationResourceResolvers[id];
-    });
+        const compat = new CompatibleApplicationResourceBranchDataProvider(resolver, 'azureResourceGroups.loadMore' /** TODO: what is the correct value for this? */);
```
nturinski, line 24: I think it's the load more of the resource providers, right? So it may have to be a command that's passed in from the resource extensions.

src/api/v2/compatibility/application/CompatibleApplicationResourceTreeItem.ts
```diff
@@ -0,0 +1,108 @@
...
+
+import { AzExtParentTreeItem, AzExtResourceType, AzExtTreeDataProvider, AzExtTreeItem, IActionContext, ISubscriptionContext, nonNullProp, TreeItemIconPath } from "@microsoft/vscode-azext-utils";
+import type { ResolvedAppResourceBase } from "@microsoft/vscode-azext-utils/hostapi";
+import { TreeItemCollapsibleState } from "vscode";
+import { createResolvableProxy } from "../../../../tree/AppResourceTreeItem";
+import { getIconPath } from "../../../../utils/azureUtils";
+import { AzureResource } from "../../v2AzureResourcesApi";
+
+/**
+ * Must immitate the behavior of AppResourceTreeItem
```
nturinski, line 14:
```suggestion
 * Must imitate the behavior of AppResourceTreeItem
```

src/api/v2/compatibility/application/CompatibleApplicationResourceTreeItem.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+export class CompatibleResolvedApplicationResourceTreeItem extends AzExtParentTreeItem {
```
nturinski, line 16: If the intent of this class is to imitate the behavior of the `AppResourceTreeItem`, should this also implement `GroupableResource, AppResource` like `AppResourceTreeItem` does? 

In doing so, it reveals that `groupConfig` isn't defined. Though I wouldn't think that should be necessary to have since the original `AppResourceTreeItem` before it is resolved contains that.

Following that line of thinking, isn't this imitating the `ResolvedAppResourceBase` more than an `AppResourceTreeItem`?

src/api/v2/compatibility/application/CompatibleApplicationResourceTreeItem.ts
alexweininger: 😨

src/api/registerApplicationResourceResolver.ts
```diff
@@ -3,21 +3,30 @@
...
+export const applicationResourceResolvers: Partial<Record<AzExtResourceType, AppResourceResolver>> = {};
 
-export function registerApplicationResourceResolver(id: string, resolver: AppResourceResolver): Disposable {
-    if (applicationResourceResolvers[id]) {
-        throw new Error(`Application resource resolver with id '${id}' has already been registered.`);
-    }
+export function registerApplicationResourceResolver(type: AzExtResourceType, resolver: AppResourceResolver): Disposable {
+    return callWithTelemetryAndErrorHandlingSync('registerApplicationResourceProvider', () => {
+        if (applicationResourceResolvers[type]) {
+            throw new Error(`Application resource resolver with id '${type}' has already been registered.`);
```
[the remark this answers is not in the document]
bwateratmsft: I'm not sure it's necessary to localize. If this error is being thrown, it's 100% a bug on our end. I would never expect a user to see this message.

src/api/registerApplicationResourceResolver.ts
bwateratmsft: It would be to load more of the resources, not the providers. But I'm not sure if we need this here or not since we aren't truncating resource lists.

src/api/v2/compatibility/application/CompatibleApplicationResourceTreeItem.ts
alexweininger: I'm not sure I follow exactly, but

> Following that line of thinking, isn't this imitating the ResolvedAppResourceBase more than an AppResourceTreeItem?

That is exactly correct, this is imitating a `ResolvedAppResourceBase`. I will update the comment.

src/api/registerApplicationResourceResolver.ts
```diff
@@ -3,21 +3,30 @@
...
-    ext.emitters.onDidRegisterResolver.fire(resolver);
+        applicationResourceResolvers[type] = resolver;
+        ext.emitters.onDidRegisterResolver.fire(resolver);
 
-    return new Disposable(() => {
-        delete applicationResourceResolvers[id];
-    });
+        const compat = new CompatibleApplicationResourceBranchDataProvider(resolver, 'azureResourceGroups.loadMore' /** TODO: what is the correct value for this? */);
+
+        ext.v2.api.registerApplicationResourceBranchDataProvider(type, compat as unknown as AzureResourceBranchDataProvider<AzExtTreeItem>);
```
bwateratmsft, line 26: Need to combine this disposable with the one below to dispose of both at the same time

src/api/v2/compatibility/workspace/CompatibilityWorkspaceResourceProvider.ts
```diff
@@ -0,0 +1,44 @@
...
+import { AzExtParentTreeItem, AzExtTreeItem } from "@microsoft/vscode-azext-utils";
+import { WorkspaceResourceProvider } from "@microsoft/vscode-azext-utils/hostapi";
+import { Event, WorkspaceFolder } from "vscode";
+import { WorkspaceResource, WorkspaceResourceProvider as V2WorkspaceResourceProvider } from "../../v2AzureResourcesApi";
+import { CompatibleWorkspaceResourceBranchDataProvider } from "./CompatibleWorkspaceResourceBranchDataProvider";
+
+export class CompatibilityWorkspaceResourceProvider implements V2WorkspaceResourceProvider {
+    constructor(private readonly resourceType: string, private readonly provider: WorkspaceResourceProvider) { }
+
+    // TODO: not sure what to do with this?
```
bwateratmsft, line 15: This can stay undefined; v1 didn't have a comparable mechanism

src/tree/v2/BranchDataProviderItem.ts
```diff
@@ -43,9 +44,15 @@ export class BranchDataProviderItem implements ResourceGroupsItem, WrappedResour
         private readonly itemCache: BranchDataItemCache,
         private readonly options?: BranchDataItemOptions) {
         itemCache.addBranchItem(this.branchItem, this);
+
+        if (isAzExtTreeItem(this.branchItem)) {
```
bwateratmsft, line 48: Add a comment about the compat?

src/tree/v2/azure/GroupingItem.ts
```diff
@@ -33,7 +33,12 @@ export class GroupingItem implements ResourceGroupsItem {
         private readonly iconPath: TreeItemIconPath | undefined,
         public readonly label: string,
         public readonly resources: AzureResource[],
-        public readonly parent?: ResourceGroupsItem) {
+        public readonly resourceType: AzExtResourceType | undefined,
+        public readonly parent?: ResourceGroupsItem
+    ) { }
+
+    public get subscription(): ISubscriptionContext {
```
bwateratmsft, line 40: Add a comment about why this is needed? (commands on the groups)

src/tree/v2/azure/GroupingItem.ts
```diff
@@ -67,7 +72,7 @@ export class GroupingItem implements ResourceGroupsItem {
     async getTreeItem(): Promise<vscode.TreeItem> {
         const treeItem = new vscode.TreeItem(this.label, vscode.TreeItemCollapsibleState.Collapsed);
 
-        treeItem.contextValue = this.contextValues?.join(' ');
+        treeItem.contextValue = this.contextValues?.sort().join(';');
```
bwateratmsft, line 75: Use a common utility

src/commands/workspace/refreshWorkspace.ts
```diff
@@ -3,9 +3,8 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { IActionContext } from "@microsoft/vscode-azext-utils";
 import { ext } from "../../extensionVariables";
 
-export async function refreshWorkspace(context: IActionContext): Promise<void> {
-    await ext.workspaceTree.refresh(context);
+export function refreshWorkspace(): void {
```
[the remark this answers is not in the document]
alexweininger: Probably ok to leave this in this PR. It's just fixing up the workspace refresh command, which is broken otherwise.

src/tree/AppResourceTreeItem.ts
```diff
@@ -66,32 +66,7 @@ export class AppResourceTreeItem extends ResolvableTreeItemBase implements Group
      */
     public static Create(parent: AzExtParentTreeItem, resource: AppResource): AppResourceTreeItem {
```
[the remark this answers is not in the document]
alexweininger: I vote for separately 😄

src/api/registerWorkspaceResourceProvider.ts
```diff
@@ -3,23 +3,35 @@
...
+        disposables.push(ext.v2.api.registerWorkspaceResourceProvider(new CompatibilityWorkspaceResourceProvider(resourceType, provider)));
+        disposables.push(ext.v2.api.registerWorkspaceResourceBranchDataProvider(resourceType, new CompatibleWorkspaceResourceBranchDataProvider('azureWorkspace.loadMore') as unknown as BranchDataProvider<WorkspaceResource, AzExtTreeItem>));
 
         return new Disposable(() => {
+            for (const disposable of disposables) {
+                disposable.dispose();
+            }
             delete workspaceResourceProviders[resourceType];
-            void refreshWorkspace(context);
+            refreshWorkspace();
```
bwateratmsft, line 33: How come this one is `refreshWorkspace()` but the other just fires the emitter directly?

alexweininger: The other is only fired here, whereas refreshWorkspace is called other places too. If the other one needed to be fired elsewhere we'd make a util for that one too.

src/api/registerWorkspaceResourceProvider.ts
```diff
@@ -3,23 +3,31 @@
...
     workspaceResourceProviders[resourceType] = provider;
 
-    return callWithTelemetryAndErrorHandlingSync('registerWorkspaceResourceProvider', (context) => {
+    return callWithTelemetryAndErrorHandlingSync('registerWorkspaceResourceProvider', () => {
 
-        void refreshWorkspace(context);
+        refreshWorkspace();
+
+        ext.v2.api.registerWorkspaceResourceProvider(new CompatibilityWorkspaceResourceProvider(resourceType, provider));
+        ext.v2.api.registerWorkspaceResourceBranchDataProvider(resourceType, new CompatibleWorkspaceResourceBranchDataProvider('azureWorkspace.loadMore') as unknown as BranchDataProvider<WorkspaceResource, AzExtTreeItem>)
```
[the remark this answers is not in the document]
alexweininger: There was just register provider. Since workspace resources didn't need resolving.

src/api/v2/compatibility/application/CompatibleApplicationResourceTreeItem.ts
alexweininger: Comment updated

bwateratmsft:
> I believe all comments have been resolved.
> 
>  [ 68747470733a2f2f6d65646961312e67697068792e636f6d2f6d656469612f7134554a6c4264664c445769412f67697068792e676966 ](https://camo.githubusercontent.com/b82fa4544c1ed8e1df2a1fb9c49b381c3d21962794881ef8307484bd8f95f2e4/68747470733a2f2f6d65646961312e67697068792e636f6d2f6d656469612f7134554a6c4264664c445769412f67697068792e676966) [ ](https://camo.githubusercontent.com/b82fa4544c1ed8e1df2a1fb9c49b381c3d21962794881ef8307484bd8f95f2e4/68747470733a2f2f6d65646961312e67697068792e636f6d2f6d656469612f7134554a6c4264664c445769412f67697068792e676966)

:shipit:

PR #365, opened 2022-08-18
Switch to new test package

alexweininger:


package.json
```diff
@@ -575,5 +575,11 @@
     },
     "extensionDependencies": [
         "ms-vscode.azure-account"
-    ]
+    ],
+    "__metadata": {
```
nturinski, line 579: Where did this come from??

package-lock.json
```diff
@@ -13249,6 +13249,18 @@
                 "@microsoft/1ds-post-js": "^3.2.3"
             }
         },
+        "@vscode/test-electron": {
```
nturinski, line 13252: Weird, it seems like the `package-lock.json` updated but I don't see these changes reflected in the `package.json` itself.

package.json
alexweininger: Added automatically by the new test package.

package-lock.json
alexweininger: Fixed. I slipped up when reverting the package.json changes that the new test package automatically makes (it annoyingly changes all the whitespace, etc).

package.json
nturinski: It's not in the Functions repo [though](https://github.com/microsoft/vscode-azurefunctions/blob/main/package.json#L1117). Did we explicitly remove it?

bwateratmsft: Yeah, I'd prefer to remove it. It's puzzling why it's being added at all, I think we should raise it with the package authors.

alexweininger: I removed it, but according to https://github.com/microsoft/vscode/issues/148975#issuecomment-1124930941 it's intentionally added and "can be used later for other features like updating etc.,"

bwateratmsft: I think it makes sense for VSCode to update package.json in the installed extensions folder, but I don't think it should be doing it in source (like it currently is here).

nturinski (approved):

PR #366, opened 2022-08-18
Support legacy resource types

alexweininger:
Needed to make sure extensions that haven't updated to use the `AzExtResourceType` context value yet will still be able to add the proper context menus to the group tree items.

We made this change in https://github.com/microsoft/vscode-azureresourcegroups/pull/337, but forgot to consider this breaking change.

### DIFF
--- a/src/azureExtensions.ts
+++ b/src/azureExtensions.ts
@@ -90,7 +90,6 @@ export const legacyTypeMap: Partial<Record<AzExtResourceType, string>> = {
     FunctionApp: 'microsoft.web/functionapp',
     AppServices: 'microsoft.web/sites',
     StaticWebApps: 'microsoft.web/staticsites',
-    StorageAccounts: 'microsoft.storage/storageaccounts',
     VirtualMachines: 'microsoft.compute/virtualmachines',
     AzureCosmosDb: 'microsoft.documentdb/databaseaccounts',
     PostgresqlServersStandard: 'microsoft.dbforpostgresql/servers',
--- a/src/azureExtensions.ts
+++ b/src/azureExtensions.ts
@@ -87,6 +87,17 @@ export const azureExtensions: IAzExtMetadata[] = [
     }
 ];
 
+export const legacyTypeMap: Partial<Record<AzExtResourceType, string>> = {
+    FunctionApp: 'microsoft.web/functionapp',
+    AppServices: 'microsoft.web/sites',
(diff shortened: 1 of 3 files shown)


src/azureExtensions.ts
```diff
@@ -86,6 +86,17 @@ export const azureExtensions: IAzExtMetadata[] = [
     }
 ];
 
+export const legacyTypeMap: Partial<Record<AzExtResourceType, string>> = {
```
nturinski, line 89: Is the plan to start deleting these as we make the other extensions compatible? If so, then okie dokes, but if not, we're missing Container Apps in here.

src/utils/azureUtils.ts
```diff
@@ -34,7 +34,7 @@ export function createGroupConfigFromResource(resource: AppResource, subscriptio
             label: resource.azExtResourceType ? azExtDisplayInfo[resource.azExtResourceType ?? '']?.displayName ?? unknown : unknown,
             id: `${subscriptionId}/${resource.azExtResourceType}`,
             iconPath,
-            contextValuesToAdd: ['azureResourceTypeGroup', ...(resource.azExtResourceType ? [resource.azExtResourceType] : [])]
+            contextValuesToAdd: ['azureResourceTypeGroup', ...(resource.azExtResourceType ? [resource.azExtResourceType] : []), ...(resource.azExtResourceType ? [legacyTypeMap[resource.azExtResourceType] ?? ''] : [])]
```
nturinski, line 37: As I mentioned to Matt, I don't think you need to do the legacy stuff in here because all of our Azure Extension groups are already created. Technically, this is only for resource types that aren't part of our extension list and to have the proper id to compare it to the existing `GroupTreeItem`.

I only bring it up because this ternary is particularly nasty looking and this is a temporary workaround until we ship all of the extensions with the updated context values.

nturinski (approved):

src/utils/azureUtils.ts
alexweininger: Thanks, wasn't sure I needed it here.

> this ternary is particularly nasty looking

Ya I wish there was an easier way to conditionally add an item to an array inline. Maybe we make a util for this lol

src/azureExtensions.ts
nturinski: Probably should go ahead and remove StorageAccounts and update the context values appropriately before we forget.

nturinski (approved):

PR #369, opened 2022-08-22
Update 'PickAppResource' to allow retrieval of local workspace resources

MicroFish91:
I added an update to `pickAppResource` to support this [Storage commit](https://github.com/microsoft/vscode-azurestorage/pull/1126) which also involved changes in [AzureTools](https://github.com/microsoft/vscode-azuretools/pull/1228).  The idea was to add two additional fields under `PickAppResourceOptions` which would allow for a way to choose between traversing the workspace vs. the remote tree.

I decided to make the change this way because I felt it was one of the least invasive routes in that we would not have to refactor any of our existing code and adding the support to any future command would involve only changing one or two fields.  I could add it as a completely separate command if that seems like a better representation, let me know!

### DIFF
--- a/src/api/pickAppResource.ts
+++ b/src/api/pickAppResource.ts
@@ -5,16 +5,48 @@
 
 import { AzExtTreeItem, ITreeItemPickerContext } from "@microsoft/vscode-azext-utils";
 import { PickAppResourceOptions } from "@microsoft/vscode-azext-utils/hostapi";
+import { QuickPickItem } from "vscode";
+import { ResourceLocation } from "../constants";
 import { ext } from "../extensionVariables";
 import { SubscriptionTreeItem } from "../tree/SubscriptionTreeItem";
+import { WorkspaceTreeItem } from "../tree/WorkspaceTreeItem";
+import { localize } from "../utils/localize";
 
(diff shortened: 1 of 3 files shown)


nturinski:
I think that ultimately, the approach that we would want to take would allow us to mix in context values of both local and remote resources because a lot of the time, we need to display a list of both.

For example, a user would want a list of storage accounts which could include attached/emulated ones. So ideally we could generate a list of quickpicks that would display all of these options.

If we don't want to wait for the treeV2 API to fix this, I think I'd prefer an approach where the `pickAppResources` command either always checks both trees which would require changing the behavior of `showTreeItemPicker` a decent amount or a way to pass in custom nodes to the top level quickpick selector. 

@alexweininger Your thoughts?

MicroFish91:
> I think I'd prefer an approach where the `pickAppResources` command either always checks both trees which would require changing the behavior of `showTreeItemPicker` a decent amount or a way to pass in custom nodes to the top level quickpick selector.

Yeah I was considering doing it this way, it seems like a better experience but it felt like I'd have to change up a lot of the core functionality of things so I decided to just propose the simple way first to get the conversation going.  Looking forward to seeing which direction you guys prefer for this :)

MicroFish91:
Leaving this for v2 of the API.

PR #379, opened 2022-09-13
Add legacy type to resource type tree items

alexweininger:


nturinski (approved):
Maybe you don't need it in the other config then? But whatever 🤷🏻‍♂️

alexweininger:
> Maybe you don't need it in the other config then? But whatever 🤷🏻‍♂️

Well I think the other config matters if you don't have any resources of that type.

PR #381, opened 2022-09-15
Add `CompatibleBranchDataItem`

alexweininger:
**PR 1 of 3** - should merge/review this PR before the others

Changes provide compatibility with V1 `registerAppResourceResolver` method by registering a `CompatibleBranchDataProvider` wrapped around the given resolver.

This means that with minimal changes ([view changes required for Functions](https://github.com/microsoft/vscode-azurefunctions/compare/main...alex/api-v2-compat)), "trees" (resolvers) client extensions provide for V1, work with V2.

image

### DIFF
--- a/src/api/registerApplicationResourceResolver.ts
+++ b/src/api/registerApplicationResourceResolver.ts
@@ -3,6 +3,7 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
+import { AzExtResourceType } from "@microsoft/vscode-azext-utils";
 import { AppResourceResolver } from "@microsoft/vscode-azext-utils/hostapi";
 import { Disposable } from "vscode";
 import { ext } from "../extensionVariables";
@@ -12,19 +13,19 @@ import { ApplicationResource, BranchDataProvider } from "./v2/v2AzureResourcesAp
 
 export const applicationResourceResolvers: Record<string, AppResourceResolver> = {};
 
-export function registerApplicationResourceResolver(id: string, resolver: AppResourceResolver): Disposable {
-    if (applicationResourceResolvers[id]) {
-        throw new Error(`Application resource resolver with id '${id}' has already been registered.`);
+export function registerApplicationResourceResolver(type: AzExtResourceType, resolver: AppResourceResolver): Disposable {
+    if (applicationResourceResolvers[type]) {
+        throw new Error(`Application resource resolver with id '${type}' has already been registered.`);
     }
 
-    applicationResourceResolvers[id] = resolver;
+    applicationResourceResolvers[type] = resolver;
     ext.emitters.onDidRegisterResolver.fire(resolver);
 
     const compat = new CompatibleBranchDataProvider(resolver, 'azureResourceGroups.loadMore' /** TODO: what is the correct value for this? */);
 
-    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat as unknown as BranchDataProvider<ApplicationResource, CompatibleBranchDataItem>);
+    ext.v2.api.registerApplicationResourceBranchDataProvider(type, compat as unknown as BranchDataProvider<ApplicationResource, CompatibleBranchDataItem>);
 
     return new Disposable(() => {
-        delete applicationResourceResolvers[id];
+        delete applicationResourceResolvers[type];
     });
 }
--- a/src/api/registerApplicationResourceResolver.ts
+++ b/src/api/registerApplicationResourceResolver.ts
@@ -6,6 +6,8 @@
 import { AppResourceResolver } from "@microsoft/vscode-azext-utils/hostapi";
 import { Disposable } from "vscode";
 import { ext } from "../extensionVariables";
+import { CompatibleBranchDataProvider } from "./v2/compatibility/CompatibleBranchDataProvider";
+import { ApplicationResource, BranchDataProvider, ResourceModelBase } from "./v2/v2AzureResourcesApi";
 
 export const applicationResourceResolvers: Record<string, AppResourceResolver> = {};
 
@@ -16,6 +18,10 @@ export function registerApplicationResourceResolver(id: string, resolver: AppRes
 
     applicationResourceResolvers[id] = resolver;
     ext.emitters.onDidRegisterResolver.fire(resolver);
+
+    const compat = new CompatibleBranchDataProvider(resolver, 'azureResourceGroups.loadMore' /** TODO: what is the correct value for this? */);
+
+    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat as unknown as BranchDataProvider<ApplicationResource, ResourceModelBase>);
 
     return new Disposable(() => {
         delete applicationResourceResolvers[id];
--- a/src/api/registerApplicationResourceResolver.ts
+++ b/src/api/registerApplicationResourceResolver.ts
@@ -7,7 +7,6 @@ import { AppResourceResolver } from "@microsoft/vscode-azext-utils/hostapi";
 import { Disposable } from "vscode";
 import { ext } from "../extensionVariables";
 import { CompatibleBranchDataProvider } from "./v2/compatibility/CompatibleBranchDataProvider";
-import { ApplicationResource, BranchDataProvider, ResourceModelBase } from "./v2/v2AzureResourcesApi";
 
 export const applicationResourceResolvers: Record<string, AppResourceResolver> = {};
 
@@ -21,7 +20,7 @@ export function registerApplicationResourceResolver(id: string, resolver: AppRes
 
     const compat = new CompatibleBranchDataProvider(resolver, 'azureResourceGroups.loadMore' /** TODO: what is the correct value for this? */);
 
-    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat as unknown as BranchDataProvider<ApplicationResource, ResourceModelBase>);
+    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat);
 
     return new Disposable(() => {
         delete applicationResourceResolvers[id];
--- a/src/api/registerApplicationResourceResolver.ts
+++ b/src/api/registerApplicationResourceResolver.ts
@@ -21,7 +21,7 @@ export function registerApplicationResourceResolver(id: string, resolver: AppRes
 
     const compat = new CompatibleBranchDataProvider(resolver, 'azureResourceGroups.loadMore' /** TODO: what is the correct value for this? */);
 
-    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat as unknown as BranchDataProvider<ApplicationResource, ResourceModelBase>);
+    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat as unknown as BranchDataProvider<ApplicationResource, CompatibleBranchDataItem>);
 
     return new Disposable(() => {
         delete applicationResourceResolvers[id];
--- a/src/api/registerApplicationResourceResolver.ts
+++ b/src/api/registerApplicationResourceResolver.ts
@@ -7,6 +7,7 @@ import { AppResourceResolver } from "@microsoft/vscode-azext-utils/hostapi";
 import { Disposable } from "vscode";
 import { ext } from "../extensionVariables";
 import { CompatibleBranchDataProvider } from "./v2/compatibility/CompatibleBranchDataProvider";
+import { ApplicationResource, BranchDataProvider, ResourceModelBase } from "./v2/v2AzureResourcesApi";
 
 export const applicationResourceResolvers: Record<string, AppResourceResolver> = {};
 
@@ -20,7 +21,7 @@ export function registerApplicationResourceResolver(id: string, resolver: AppRes
 
     const compat = new CompatibleBranchDataProvider(resolver, 'azureResourceGroups.loadMore' /** TODO: what is the correct value for this? */);
 
-    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat);
+    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat as unknown as BranchDataProvider<ApplicationResource, ResourceModelBase>);
 
     return new Disposable(() => {
         delete applicationResourceResolvers[id];
--- a/src/api/registerApplicationResourceResolver.ts
+++ b/src/api/registerApplicationResourceResolver.ts
@@ -6,8 +6,9 @@
 import { AppResourceResolver } from "@microsoft/vscode-azext-utils/hostapi";
 import { Disposable } from "vscode";
 import { ext } from "../extensionVariables";
+import { CompatibleBranchDataItem } from "../tree/v2/CompatibleBranchDataItem";
 import { CompatibleBranchDataProvider } from "./v2/compatibility/CompatibleBranchDataProvider";
-import { ApplicationResource, BranchDataProvider, ResourceModelBase } from "./v2/v2AzureResourcesApi";
+import { ApplicationResource, BranchDataProvider } from "./v2/v2AzureResourcesApi";
 
 export const applicationResourceResolvers: Record<string, AppResourceResolver> = {};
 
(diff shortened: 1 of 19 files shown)


src/api/registerApplicationResourceResolver.ts
```diff
@@ -17,6 +19,10 @@ export function registerApplicationResourceResolver(id: string, resolver: AppRes
     applicationResourceResolvers[id] = resolver;
     ext.emitters.onDidRegisterResolver.fire(resolver);
 
+    const compat = new CompatibleBranchDataProvider(resolver, 'azureResourceGroups.loadMore' /** TODO: what is the correct value for this? */);
+
+    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat as unknown as BranchDataProvider<ApplicationResource, ResourceModelBase>);
```
bwateratmsft, line 24: What makes this `unknown` cast necessary? `CompatibleBranchDataProvider` does implement `BranchDataProvider`.

src/api/v2/compatibility/CompatibleAzureResourceGroupsExtensionApi.ts
```diff
@@ -6,14 +6,13 @@
...
-import { CompatibleAzExtTreeDataProvider } from './CompatibleAzExtTreeDataProvider';
 
 export class CompatibleAzureResourceGroupsExtensionApi implements AzureHostExtensionApi, AzureResourceGroupsExtensionApi {
     public static apiVersion = '0.1.0';
 
-    #appResourceTree: CompatibleAzExtTreeDataProvider;
+    #appResourceTree: AzExtTreeDataProvider;
     #appResourceTreeView: TreeView<unknown>;
-    #workspaceResourceTree: CompatibleAzExtTreeDataProvider;
+    #workspaceResourceTree: AzExtTreeDataProvider;
```
bwateratmsft, line 15: Since these are the private members I think it's fine for them to refer to the true type.

src/api/v2/compatibility/CompatibleBranchDataProvider.ts
```diff
@@ -4,21 +4,29 @@
...
-        super(undefined as unknown as AzExtParentTreeItem, loadMoreCommandId);
+        // Using `{}` here so property assignment doesn't throw
+        super({} as unknown as AzExtParentTreeItem, loadMoreCommandId);
     }
 
     //#region TreeDataProvider
 
-    public override readonly onDidChangeTreeData = this.overrideOnDidChangeTreeDataEmitter.event;
+    // @ts-expect-error `getParent` is not meant to be defined by `BranchDataProvider`s but is already defined by `AzExtTreeDataProvider`
+    public override get onDidChangeTreeData(): Event<AzExtTreeItem | undefined> {
```
bwateratmsft, line 27: What made this change necessary?

src/api/registerApplicationResourceResolver.ts
bwateratmsft, line 24: Can this use `CompatibleBranchDataItem` instead of `ResourceModelBase`?
```suggestion
    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat as unknown as BranchDataProvider<ApplicationResource, CompatibleBranchDataItem>);
```

src/tree/v2/BranchDataItem.ts
```diff
@@ -50,8 +51,10 @@ export class BranchDataItem implements ResourceGroupsItem, Wrapper {
     type: string;
 }
 
-export type BranchDataItemFactory = (branchItem: ResourceModelBase, branchDataProvider: BranchDataProvider<ApplicationResource, ResourceModelBase>, options?: BranchDataItemOptions) => BranchDataItem;
+export type BranchDataItemFactory = (branchItem: ResourceModelBase, branchDataProvider: BranchDataProvider<ApplicationResource, ResourceModelBase>, options?: BranchDataItemOptions) => BranchDataItem | CompatibleBranchDataItem;
```
bwateratmsft, line 54: IMO we should make `CompatibleBranchDataItem` extend `BranchDataItem` instead of making this `BranchDataItem | CompatibleBranchDataItem`

src/tree/v2/CompatibleBranchDataItem.ts
```diff
@@ -0,0 +1,57 @@
...
+    async getChildren(): Promise<ResourceGroupsItem[] | undefined> {
+        const children = await this.branchDataProvider.getChildren(this.branchItem);
+        const factory = createBranchDataItemFactory(this.itemCache);
+        return children?.map(child => factory(child, this.branchDataProvider));
+    }
+
+    public get quickPickOptions(): CompatibleQuickPickOptions {
+        const ti = this.branchItem as AzExtTreeItem;
+
+        const maybeParent = ti as AzExtParentTreeItem;
```
bwateratmsft, line 27: As long as we're opening up the `isAzExtTreeItem` method, I think it would make sense to have a corresponding `isAzExtParentTreeItem` method to go with it. There's one implemented but it's currently internal-only.

src/tree/v2/providers/BuiltInApplicationResourceItem.ts
```diff
@@ -5,6 +5,8 @@ import { BuiltInResourceModelBase } from './BuiltInResourceModelBase';
 
 export class BuiltInApplicationResourceItem implements BuiltInResourceModelBase {
     constructor(private readonly resource: ApplicationResource) {
+        this.id = resource.id;
```
bwateratmsft, line 8: Using Azure resource ID as tree item ID could bite us.

src/api/registerApplicationResourceResolver.ts
alexweininger: It's because we have `getParent` defined in `CompatibleBranchDataProvider`, but in `BranchDataProvider` we have `getParent?: never;`. We need to remove one of those.

```
Argument of type 'CompatibleBranchDataProvider<ApplicationResource, AzExtTreeItem & ResourceModelBase>' is not assignable to parameter of type 'BranchDataProvider<ApplicationResource, AzExtTreeItem>'.
  Types of property 'getParent' are incompatible.
    Type '(_treeItem: AzExtTreeItem & ResourceModelBase) => Promise<AzExtTreeItem & ResourceModelBase>' is not assignable to type 'undefined'.
```

bwateratmsft: Ah, I see. I would have thought that the `//@ts-expect-error` above the `getParent()` definition in `CompatibleBranchDataProvider` would work. Oh well!

src/api/v2/compatibility/CompatibleBranchDataProvider.ts
alexweininger: It's being set in the constructor of `AzExtTreeDataProvider`. I changed it back to the getter, and added an empty setter.

src/api/registerApplicationResourceResolver.ts
alexweininger: We don't need that cast anymore

bwateratmsft: I put the `never` in on `getParent?` to denote that users of the API do not need to (and should not) implement it--since RGs can do it all. I think I'd rather have the cast here than lose that.

src/api/v2/compatibility/CompatibleBranchDataProvider.ts
bwateratmsft: [I'm not seeing it](https://github.com/microsoft/vscode-azuretools/blob/main/utils/src/tree/AzExtTreeDataProvider.ts#L30), is there some indirect thing? It doesn't look like it ever does a set of `onDidChangeTreeData`.

alexweininger: Sorry, I thought I saw it being set. Well it's being set somewhere, let me find out where. Also TS is unhappy that it's a property in `AzExtTreeDataProvider`, but is overridden here as an accessor.

alexweininger: But if I remove `readonly` it'll work, but I'm not sure we should allow it to be set from elsewhere. Which is why I went with the empty setter.

bwateratmsft: Ugh. TypeScript is so fussy about this. I think an empty setter is fine here.

src/tree/v2/providers/BuiltInApplicationResourceItem.ts
alexweininger: I don't think I meant to include this change in this PR. This id is needed for the view properties command

src/tree/v2/providers/BranchDataProviderManager.ts
```diff
@@ -43,7 +43,7 @@ export class BranchDataProviderManager extends vscode.Disposable {
     }
 
     getApplicationResourceBranchDataProvider(resource: ApplicationResource): BranchDataProvider<ApplicationResource, ResourceModelBase> {
-        const provider = this.applicationResourceBranchDataProviders[resource.type.type];
+        const provider = this.applicationResourceBranchDataProviders[resource.azExtResourceType?.toLowerCase() ?? resource.type.type];
```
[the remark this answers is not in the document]
alexweininger: Yeah, removed all that lowercasing. Also changed it so that branch data providers are registered via an `AzExtResourceType` rather than a string.

src/tree/v2/CompatibleBranchDataItem.ts
alexweininger: Created https://github.com/microsoft/vscode-azuretools/pull/1236

src/tree/v2/CompatibleBranchDataItem.ts
```diff
@@ -1,18 +1,20 @@
...
 import { ResourceGroupsItemCache } from './ResourceGroupsItemCache';
 
-export class CompatibleBranchDataItem implements ResourceGroupsItem, Box, CompatibleContextValueFilterableTreeNode {
+export class CompatibleBranchDataItem extends BranchDataItem {
     constructor(
-        private readonly branchItem: ResourceModelBase,
-        private readonly branchDataProvider: BranchDataProvider<ApplicationResource, ResourceModelBase>,
-        private readonly itemCache: ResourceGroupsItemCache,
-        private readonly options: BranchDataItemOptions | undefined
+        protected readonly branchItem: ResourceModelBase,
```
bwateratmsft, line 12: These four are implicitly overriding the inherited `protected readonly` fields of the same name. So you can remove `protected readonly` from each of the four arguments in this class. The `super()` call will send them up to `BranchDataItem`.

src/api/registerApplicationResourceResolver.ts
```diff
@@ -17,6 +20,10 @@ export function registerApplicationResourceResolver(id: string, resolver: AppRes
     applicationResourceResolvers[id] = resolver;
     ext.emitters.onDidRegisterResolver.fire(resolver);
 
+    const compat = new CompatibleBranchDataProvider(resolver, 'azureResourceGroups.loadMore' /** TODO: what is the correct value for this? */);
+
+    ext.v2.api.registerApplicationResourceBranchDataProvider(id, compat as unknown as BranchDataProvider<ApplicationResource, CompatibleBranchDataItem>);
```
bwateratmsft, line 25: This is now incompatible with the change to register BDPs by type. I'm a little reluctant to go ahead with that, as we'd need to do a built-in BDP for each of the unsupported resource types, whereas before we had the `matchesResource` method that could just check, and so there was a single BDP for all of them. @philliphoff what do you think?

alexweininger: > This is now incompatible with the change to register BDPs by type

I'm not sure what you mean by this. I was planning on changing the IDs of the resolvers in the client extensions to match the `AzExtResourceType`.

Ex: https://github.com/microsoft/vscode-azurefunctions/blob/1d2514971b53806b9352844eadfc350cf31d4a56/src/extension.ts#L98

bwateratmsft: Your change makes `registerApplicationResourceBranchDataProvider` accept an `AzExtResourceType` as the first parameter, not an ID, so line 25 here would need to change. But it carries the disadvantage that no BDP can represent multiple types.

alexweininger: > we'd need to do a built-in BDP for each of the unsupported resource types

I think that's covered by the `defaultBranchDataProvider` argument in `BranchDataProviderManager`. It takes a BDP and will use that if no matching BDPs are found.

https://github.com/microsoft/vscode-azureresourcegroups/blob/alex/compat-shim-changes-1/src/tree/v2/providers/BranchDataProviderManager.ts#L14

alexweininger: > But it carries the disadvantage that no BDP can represent multiple types.

I thought we'd just register the same instance of the BDP multiple times to cover that scenario.

src/tree/v2/CompatibleBranchDataItem.ts
```diff
@@ -27,20 +27,18 @@ export class CompatibleBranchDataItem extends BranchDataItem {
...
+                label: ti.childTypeLabel ?
+                    localize('createNewItem', '$(plus) Create new {0}', ti.childTypeLabel) :
                     localize('createNew', '$(plus) Create new...'),
             } :
             undefined;
 
         return {
             contextValues: ti.contextValue.split(';'),
-            isLeaf: !maybeParent?.loadMoreChildrenImpl,
+            isLeaf: isAzExtParentTreeItem(ti),
```
bwateratmsft, line 41: Logic is backwards here.
```suggestion
            isLeaf: !isAzExtParentTreeItem(ti),
```

src/api/registerApplicationResourceResolver.ts
bwateratmsft: I see, I didn't realize that's how the default BDP worked. I think for more narrowly-scoped cases, where one BDP is handling e.g. 2 types, it's (probably) fine to reregister the same one several times (maybe there are some issues when it comes to change notification events etc.). What I wanted to avoid was ~30 registrations for unsupported types, but the way the default is implemented, that will not be necessary.

bwateratmsft: Still, don't we need to change `id` in this line to the `AzExtResourceType`?

alexweininger: I'm expecting the `AzExtResourceType` to be passed in by the client extensions, so `id` would be an `AzExtResourceType` already. I updated the param name and type from `id: string` to `type: AzExtResourceType`.

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -200,7 +200,7 @@ export interface V2AzureResourcesApi extends AzureResourcesApiBase {
      * @param id The resolver ID. Must be unique.
      * @param resolver The resolver
      */
-    registerApplicationResourceBranchDataProvider<T extends ResourceModelBase>(id: string, provider: BranchDataProvider<ApplicationResource, T>): vscode.Disposable;
+    registerApplicationResourceBranchDataProvider<T extends ResourceModelBase>(id: AzExtResourceType, provider: BranchDataProvider<ApplicationResource, T>): vscode.Disposable;
```
bwateratmsft, line 203: Arg is still named `id` here and in [`v2AzureResourcesApiWrapper.ts`](https://github.com/microsoft/vscode-azureresourcegroups/pull/381/commits/2173759510366be4f3ae771e3e9582585798efbe#diff-960368edff109f85fc750a3ca6cd7df9d3e64c40d40221aae758490a3156526b)

alexweininger:
What's the status on getting this merged?

src/tree/v2/CompatibleBranchDataItem.ts
```diff
@@ -0,0 +1,62 @@
...
+        itemCache.addBranchItem(this.branchItem, this);
+    }
+
+    async getChildren(): Promise<ResourceGroupsItem[] | undefined> {
+        const children = await this.branchDataProvider.getChildren(this.branchItem);
+        const factory = createBranchDataItemFactory(this.itemCache);
+        return children?.map(child => factory(child, this.branchDataProvider));
+    }
+
+    public get quickPickOptions(): CompatibleQuickPickOptions {
```
bwateratmsft, line 27: This can be removed now.

bwateratmsft (approved):

src/tree/v2/CompatibleBranchDataItem.ts
alexweininger: I *think* that's removed in the next PR https://github.com/microsoft/vscode-azureresourcegroups/pull/382

PR #382, opened 2022-09-15
Make `pickAppResource` compatible with Tree API v2

alexweininger:

Changes the `pickAppResource` implementation to use `compatibilityPickAppResourceExperience`.

This means that with minimal changes ([view changes required for Functions](https://github.com/microsoft/vscode-azurefunctions/compare/main...alex/api-v2-compat)), calls to `pickAppResource` from the client work as expected.

Note: Back button functionality is still WIP on the new picker.


src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -180,7 +180,7 @@ export interface V2AzureResourcesApi extends AzureResourcesApiBase {
     /**
      * Show a quick picker of app resources. Set `options.type` to filter the picks.
      */
-    pickResource<TModel>(options?: ResourcePickOptions): vscode.ProviderResult<TModel>
+    pickResource<TModel extends ResourceModelBase>(options?: ResourcePickOptions): Promise<TModel>
```
bwateratmsft, line 183: This doesn't need to be on the API at all anymore. Just the TDP.

alexweininger: Removed. I will wait until I need to access the TDP in the clients to add that to the API (soon).

bwateratmsft: It can be in a different PR but I think we should just add it ASAP. The Storage extension will be converted out of the gate, so it will need it.

src/api/v2/v2AzureResourcesApiWrapper.ts
```diff
@@ -12,10 +12,6 @@ export class V2AzureResourcesApiWrapper implements V2AzureResourcesApi {
         return this.api.apiVersion;
     }
 
-    pickResource<TModel>(options?: ResourcePickOptions | undefined): vscode.ProviderResult<TModel> {
-        return this.callWithTelemetryAndErrorHandling('v2.pickResource', async () => await this.api.pickResource(options));
-    }
-
     revealResource(resourceId: string): Promise<void> {
```
bwateratmsft, line 15: Not related to this PR, but I'm inclined toward having this take the item instead of the ID--same as the VSCode API does. Callers can use the `FindTreeItemByIDExperience` to get said item if they only have the ID. There will need to be some translation, as their model item is what they should send, but RG needs to find its own wrapper item and give that to the VSCode `reveal` API.

alexweininger: I like that because it keeps our API small (and mimics VS Code API).

src/api/pickAppResource.ts
```diff
@@ -3,18 +3,43 @@
...
-    } else {
-        return appResource as unknown as T;
+function convertAppResourceFilterToAzExtResourceType(filter?: PickAppResourceOptions['filter']): AzExtResourceType[] | undefined {
+    if (!filter) {
+        return undefined;
     }
+
+    filter = Array.isArray(filter) ? filter : [filter];
+
+    return filterMap(filter, (f) => getAzExtResourceType({ type: f.type, kind: f.kind }))
```
bwateratmsft, line 28: Since `type` and `kind` are already properties on `f`, you might be able to do this:
```suggestion
    return filterMap(filter, (f) => getAzExtResourceType(f))
```

alexweininger: Nice!

src/tree/v2/BranchDataItem.ts
```diff
@@ -18,6 +18,11 @@ export class BranchDataItem implements ResourceGroupsItem, Wrapper {
         itemCache.addBranchItem(this.branchItem, this);
     }
 
+    /** Needed for tree item picker PickAppResourceStep */
+    public get resource(): ApplicationResource | undefined {
```
bwateratmsft, line 22: If we don't already have it, I think there ought to be an interface for the RG items specifically at the AppResource level, where they will have this `resource` property. i.e.:
```typescript
interface AppResourceItem extends ResourceGroupsItem {
    readonly resource: ApplicationResource | undefined; // TODO: can `undefined` even happen?
}
```

alexweininger: I've added `ApplicationResourceBranchDataItem` which handles special app resource behavior.

PR #383, opened 2022-09-15
Implement view properties command for v2

alexweininger:
**PR 3 of 3** - should merge/review this PR after the others


bwateratmsft (approved):

bwateratmsft:
This is great. We can do something like this for View in Portal as well.

alexweininger:
I was considering doing something like this too:

```ts
export interface ViewPropertiesModel {
    fullId: string;
    label: string;
    data: {};
}

export interface CanViewProperties {
    viewPropertiesModel: ViewPropertiesModel;
}

export async function viewProperties(context: IActionContext, node?: BuiltInApplicationResourceItem): Promise<void> {
    if (!node) {
        node = await appResourceExperience<BuiltInApplicationResourceItem>(context, ext.v2.resourceGroupsTreeDataProvider);
    }

    const { fullId, label, data } = node.viewPropertiesModel;
    await openReadOnlyJson({ fullId, label }, data);
}
```

then in `BuiltInApplicationResourceItem`:
```ts
export class BuiltInApplicationResourceItem implements BuiltInResourceModelBase, CanViewProperties {
    public get viewPropertiesModel(): ViewPropertiesModel {
        return {
            fullId: this.resource.id,
            data: this.resource._raw,
            label: this.resource.name,
        }
    }
```

Then any model that wants to support the view properties command can implement `CanViewProperties`. The thing I like the most about this is that it's clear what data on the item is used for the view properties command. When making changes it's often hard to tell how they might impact different commands.

Might be overkill though, what do you think?

bwateratmsft:
> edit: I guess this is also similar to just requiring a `viewProperties(): Promise<void>` method on the class too. Hmmm

I think I still like this better; I don't want every resource to have to deal with opening the JSON.

alexweininger:
Closing in favor of https://github.com/microsoft/vscode-azureresourcegroups/pull/427. Will add v1.5 client extension view properties command support in future PR.

PR #388, opened 2022-09-15
Add V2 workspace tree view

philliphoff:
Adds a basic implementation of the V2 workspace tree view.  This is mostly a refactoring of the application (i.e. resources) tree view into common base types that work across both.  Despite the number of files changed (mostly due to naming), most of the logic is the same as used by the application tree view, just made more generic.

Notable changes (and open for discussion):

We need an expanded "contribution registration" mechanism to accommodate both application and workspace resource providers.  Here's what I propose (by way of an example `package.json`):

- Contributions are split into `application` and `workspace` sections, with each having resource provider (i.e. `resources`) and brand data provider (i.e. `branches`) sections.  Providers are specified using an object that contains (today) just a `type` property (but could be expanded in the future, such as to help differentiate kinds, etc.).

```json
{
    "contributes": {
        "x-azResourcesV2": {
            "application": {
                "branches": [
                    {
                        "type": "microsoft.storage/storageaccounts"
                    }
                ]
            },
            "workspace": {
                "resources": [
                    {
                        "type": "ms-azuretools.vscode-azurestorage"
                    }
                ],
                "branches": [
                    {
                        "type": "ms-azuretools.vscode-azurestorage"
                    }
                ]
            }
        }
}
```

Screenshot 2022-09-15 at 16 18 18


src/api/v2/ResourceGroupsExtensionManager.ts
```diff
@@ -1,9 +1,22 @@
...
         readonly onFetch?: string[];
     }
+    readonly application: {
+        readonly branches?: { type: string }[];
+        readonly resources?: { type: string }[];
+    }
+    readonly workspace: {
+        readonly branches?: { type: string }[];
+        readonly resources?: { type: string }[];
+    }
```
alexweininger, line 19: For application branches and resources, I think type should be an `AzExtResourceType`.

For workspace branches and resources, it can be an arbitrary string.

src/tree/v2/application/ResourceGroupsTreeDataProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+import { localize } from '../../../utils/localize';
+import { ApplicationResourceGroupingManager } from './ApplicationResourceGroupingManager';
+import { AzureAccountExtensionApi } from '../azure-account.api';
+import { GenericItem } from '../GenericItem';
+import { ResourceGroupsItem } from '../ResourceGroupsItem';
+import { ResourceGroupsItemCache } from '../ResourceGroupsItemCache';
+import { ResourceTreeDataProviderBase } from '../ResourceTreeDataProviderBase';
+import { SubscriptionItem } from './SubscriptionItem';
+
+export class ResourceGroupsTreeDataProvider extends ResourceTreeDataProviderBase {
```
alexweininger, line 20: Rename this to `ApplicationTreeDataProvider` for consistency? Since the other TreeDataProvider is named `WorkspaceTreeDataProvider`.

src/tree/v2/ResourceTreeDataProviderBase.ts
```diff
@@ -0,0 +1,97 @@
+/*---------------------------------------------------------------------------------------------
```
alexweininger, line 1: How do we feel about moving the files in `src/tree/v2/*` to `src/tree/v2/common`? I think it's a bit cleaner and more organized.

image

image

src/api/v2/ResourceGroupsExtensionManager.ts
```diff
@@ -27,4 +40,37 @@ export class ResourceGroupsExtensionManager {
             }
         }
     }
+
+    async activateApplicationResourceProviders(): Promise<void> {
+        const inActiveResourceContributors =
+            vscode.extensions.all
+                .map(extension => ({ extension, resources: (extension.packageJSON as ExtensionPackage)?.contributes?.['x-azResourcesV2']?.application?.resources ?? [] }))
```
alexweininger, line 47: Should make the `x-azResourcesV2` string a shared constant.

alexweininger, line 47: A helper method to get `(extension.packageJSON as ExtensionPackage)?.contributes?.['x-azResourcesV2']` would be nice. It's not fun to type and/or look at 😄

src/api/v2/ResourceGroupsExtensionManager.ts
```diff
@@ -27,4 +40,37 @@ export class ResourceGroupsExtensionManager {
             }
         }
     }
+
+    async activateApplicationResourceProviders(): Promise<void> {
+        const inActiveResourceContributors =
```
alexweininger, line 45: nit: Casing `inActiveResourceContributors` -> `inactiveResourceContributors`

src/api/v2/ResourceGroupsExtensionManager.ts
```diff
@@ -27,4 +40,37 @@ export class ResourceGroupsExtensionManager {
...
+            vscode.extensions.all
+                .map(extension => ({ extension, resources: (extension.packageJSON as ExtensionPackage)?.contributes?.['x-azResourcesV2']?.application?.resources ?? [] }))
+                .filter(extensionAndContributions => extensionAndContributions.resources.length > 0 && !extensionAndContributions.extension.isActive);
+
+        await Promise.all(inActiveResourceContributors.map(contributor => contributor.extension.activate()));
+    }
+
+    async activateWorkspaceResourceProviders(): Promise<void> {
+        const inActiveResourceContributors =
```
alexweininger, line 54: nit: Casing `inActiveResourceContributors` -> `inactiveResourceContributors`

src/api/v2/WorkspaceResourceProviderManager.ts
```diff
@@ -0,0 +1,23 @@
...
+    constructor(extensionActivator: () => Promise<void>) {
+        super(extensionActivator);
+    }
+
+    async provideResources(folder: vscode.WorkspaceFolder, options?: ProvideResourceOptions): Promise<WorkspaceResource[]> {
+        const resourceProviders = await this.getResourceProviders();
+
+        const resources = await Promise.all(resourceProviders.map(resourceProvider => resourceProvider.provideResources(folder, options)));
+
+        return resources.filter(isArray).reduce((acc, result) => acc?.concat(result ?? []), []);
```
alexweininger, line 21: We could make this a util since we use it in a two different places.

* https://github.com/microsoft/vscode-azureresourcegroups/blob/philliphoff-v2-workspaces/src/api/v2/ApplicationResourceProviderManager.ts#L20
* (this file) https://github.com/microsoft/vscode-azureresourcegroups/blob/philliphoff-v2-workspaces/src/api/v2/WorkspaceResourceProviderManager.ts#L21

src/api/v2/WorkspaceResourceProviderManager.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class WorkspaceResourceProviderManager extends ResourceProviderManagerBase<WorkspaceResource, WorkspaceResourceProvider> {
+    constructor(extensionActivator: () => Promise<void>) {
+        super(extensionActivator);
+    }
+
+    async provideResources(folder: vscode.WorkspaceFolder, options?: ProvideResourceOptions): Promise<WorkspaceResource[]> {
+        const resourceProviders = await this.getResourceProviders();
+
+        const resources = await Promise.all(resourceProviders.map(resourceProvider => resourceProvider.provideResources(folder, options)));
```
alexweininger, line 19: Noticed we could be more consistent with the naming. We have `provideResources` on `WorkspaceResourceProvider` and `getResources` on `ApplicationResourceProvider`.

src/api/v2/ApplicationResourceProviderManager.ts
```diff
@@ -0,0 +1,22 @@
...
+        super(extensionActivator);
+    }
+
+    async getResources(subscription: ApplicationSubscription, options?: ProvideResourceOptions | undefined): Promise<ApplicationResource[]> {
+        const resourceProviders = await this.getResourceProviders();
+
+        const resources = await Promise.all(resourceProviders.map(resourceProvider => resourceProvider.getResources(subscription, options)));
+
+        return resources.filter(isArray).reduce((acc, result) => acc?.concat(result ?? []), []);
+    }
```
alexweininger, line 21: This method body is almost identical to the one in `WorkspaceResourceProviderManager`. Might not be worth adding more complexity to share a bit of code though.

https://github.com/microsoft/vscode-azureresourcegroups/blob/philliphoff-v2-workspaces/src/api/v2/WorkspaceResourceProviderManager.ts#L16-L22

src/tree/v2/workspace/registerWorkspaceTreeV2.ts
```diff
@@ -0,0 +1,34 @@
...
+        branchDataProviderManager,
+        refreshEvent,
+        workspaceResourceProviderManager);
+
+    context.subscriptions.push(treeDataProvider);
+
+    const treeView = vscode.window.createTreeView(
+        'azureWorkspaceV2',
+        {
+            canSelectMany: true,
```
alexweininger, line 27:
```suggestion
            canSelectMany: true,
            showCollapseAll: true,
```

src/api/v2/ResourceGroupsExtensionManager.ts
```diff
@@ -27,4 +40,37 @@ export class ResourceGroupsExtensionManager {
...
+    async activateApplicationResourceProviders(): Promise<void> {
+        const inActiveResourceContributors =
+            vscode.extensions.all
+                .map(extension => ({ extension, resources: (extension.packageJSON as ExtensionPackage)?.contributes?.['x-azResourcesV2']?.application?.resources ?? [] }))
+                .filter(extensionAndContributions => extensionAndContributions.resources.length > 0 && !extensionAndContributions.extension.isActive);
+
+        await Promise.all(inActiveResourceContributors.map(contributor => contributor.extension.activate()));
+    }
+
+    async activateWorkspaceResourceProviders(): Promise<void> {
```
bwateratmsft, line 53: Workspace resource providers might not really need an activation mechanism beyond the standard VSCode `onView:...`. Since they will always be providing nodes at the root of the tree, they always need to be activated immediately (in contrast to the remote resource view, which would start with their node collapsed a few layers down)

src/api/v2/ResourceGroupsExtensionManager.ts
```diff
@@ -27,4 +40,37 @@ export class ResourceGroupsExtensionManager {
             }
         }
     }
+
+    async activateApplicationResourceProviders(): Promise<void> {
```
bwateratmsft, line 44: Shouldn't this be to activate branches, not providers? (Since only RG has the single app resource provider)

src/api/v2/v2AzureResourcesApiImplementation.ts
```diff
@@ -1,14 +1,23 @@
...
+import { WorkspaceResourceProviderManager } from './WorkspaceResourceProviderManager';
 
 export class V2AzureResourcesApiImplementation implements V2AzureResourcesApi {
     public static apiVersion: string = '2.0.0';
 
     constructor(
-        private readonly branchDataProviderManager: BranchDataProviderManager,
-        private readonly resourceProviderManager: ApplicationResourceProviderManager) {
+        private readonly branchDataProviderManager: ApplicationResourceBranchDataProviderManager,
+        private readonly resourceProviderManager: ApplicationResourceProviderManager,
```
bwateratmsft, line 18: Nit: Since the resources are provided before the BDP kicks in, I'd flip the arguments around. i.e.
```
resourceProviderManager,
branchDataProviderManager,
workspaceResourceProviderManager,
workspaceResourceBranchDataProviderManager
```

src/utils/v2/isArray.ts
```diff
@@ -0,0 +1,8 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export function isArray<T>(maybeArray: T[] | null | undefined): maybeArray is T[] {
```
bwateratmsft, line 6: Is this really necessary? Doesn't it just replace a one-liner with a different one-liner?

src/tree/v2/ResourceBranchDataProviderManagerBase.ts
```diff
@@ -0,0 +1,78 @@
...
+                for (const providerContext of this.branchDataProviderMap.values()) {
+                    providerContext.listener?.dispose();
+                }
+            });
+    }
+
+    public readonly onDidChangeTreeData = this.onDidChangeTreeDataEmitter.event;
+
+    addProvider(type: string, provider: TBranchDataProvider): void {
+        type = this.normalizeType(type);
```
bwateratmsft, line 30: `AzExtResourceType`!

src/tree/v2/application/ApplicationResourceBranchDataProviderManager.ts
```diff
@@ -0,0 +1,21 @@
...
+import { ResourceBranchDataProviderManagerBase } from '../ResourceBranchDataProviderManagerBase';
+
+export class ApplicationResourceBranchDataProviderManager extends ResourceBranchDataProviderManagerBase<BranchDataProvider<ApplicationResource, ResourceModelBase>>{
+    constructor(
+        defaultProvider: BranchDataProvider<ApplicationResource, ResourceModelBase>,
+        extensionActivator: (type: string) => void
+    ) {
+        super(
+            defaultProvider,
+            extensionActivator);
```
bwateratmsft, line 16: Nit
```suggestion
            extensionActivator
        );
```

src/tree/v2/application/DefaultApplicationResourceBranchDataProvider.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import * as vscode from 'vscode';
+import { ApplicationResource, BranchDataProvider } from '../../../api/v2/v2AzureResourcesApi';
+import { localize } from "../../../utils/localize";
+import { ResourceGroupsItem } from '../ResourceGroupsItem';
+import { DefaultApplicationResourceItem } from './DefaultApplicationResourceItem';
+
+export class DefaultApplicationResourceBranchDataProvider implements BranchDataProvider<ApplicationResource, ResourceGroupsItem> {
+    getChildren(element?: ResourceGroupsItem | undefined): vscode.ProviderResult<ResourceGroupsItem[]> {
+        if (!element) {
```
bwateratmsft, line 14: I think this check is overkill. IMO in BDP's interface we can override `getChildren` to indicate that `element` will never be falsy, and avoid needing checks like this in every BDP.

src/tree/v2/application/ResourceGroupsTreeDataProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+                            })
+                    ];
+                }
+            }
+        }
+
+        return undefined;
+    }
+
+    private async getApi(): Promise<AzureAccountExtensionApi | undefined> {
```
bwateratmsft, line 133: Maybe this?
```suggestion
    private async getAzureAccountExtensionApi(): Promise<AzureAccountExtensionApi | undefined> {
```

src/tree/v2/workspace/WorkspaceResourceBranchDataProviderManager.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ResourceBranchDataProviderManagerBase } from '../ResourceBranchDataProviderManagerBase';
+
+export class WorkspaceResourceBranchDataProviderManager extends ResourceBranchDataProviderManagerBase<BranchDataProvider<WorkspaceResource, ResourceModelBase>> {
+    constructor(
+        defaultProvider: BranchDataProvider<WorkspaceResource, ResourceModelBase>,
+        extensionActivator: (type: string) => void
+    ) {
+        super(
+            defaultProvider,
+            extensionActivator);
```
bwateratmsft, line 16: Nit
```suggestion
            extensionActivator
        );
```

philliphoff: I suppose.  :-) Updated.

src/tree/v2/application/ResourceGroupsTreeDataProvider.ts
philliphoff: Better, yes.  Updated.

src/tree/v2/application/DefaultApplicationResourceBranchDataProvider.ts
philliphoff: Yes, your `getChildren()` override somehow never got into the API definition; I've done that and removed the `undefined` checks.

src/tree/v2/application/ApplicationResourceBranchDataProviderManager.ts
philliphoff: We should add some ESLint rules for this.  Updated.

src/utils/v2/isArray.ts
philliphoff: Strictly necessary, no.  However, the utility is useful as a typeguard in mapping functions (e.g. `collection.filter(isArray)`); `Array.isArray()` doesn't indicate type, so you're then forced to cast after the check.

src/api/v2/v2AzureResourcesApiImplementation.ts
philliphoff: I suppose; I was just going in alphabetical order.  Updated.

src/api/v2/ResourceGroupsExtensionManager.ts
philliphoff: It's to activate all extensions that contribute resources immediately prior to querying for resources; it's true that RG is the only extension today that registers a provider, but it may not be the only one forever.  (The API allows for others.)

src/api/v2/ResourceGroupsExtensionManager.ts
philliphoff: That's true, though that would mean coupling those extensions with our arbitrary ID for our resources/workspaces view, *in addition* to registering themselves as resource providers and/or BDPs.  It seems better to have extensions worry only about their specific registrations.

Unless you're suggesting that workspaces providers not have *any* static registration at all, relying on VS Code view activation and the resulting API provider registration call?  While that *could* work, it still requires registration (again, exposing our specific view ID) that would be very different than BDP registration.  I feel like we'd want registration to be consistent between resources/workspaces and provider/BDP.

bwateratmsft: That's fair. Consistency is probably a worthwhile benefit to streamlining activation through this mechanism.

src/tree/v2/workspace/registerWorkspaceTreeV2.ts
philliphoff: I'm ok with that.  (I was just copying the existing V1 configuration, which omits collapse-all for the workspaces view, assuming it was a specific design choice.)

philliphoff: Updated.

src/api/v2/WorkspaceResourceProviderManager.ts
philliphoff: Good call.  The `get` verb seems to be more prevalent in VS Code, so I'll switch to that.  Updated.

src/tree/v2/application/ResourceGroupsTreeDataProvider.ts
philliphoff: Sounds good; I've renamed them to `ApplicationResourceTreeDataProvider` and `WorkspaceResourceTreeDataProvider`.

src/api/v2/ResourceGroupsExtensionManager.ts
philliphoff: Updated.  The compiler would have ensured proper use of the string, being part of the interface, but this *does* mean that changing it would need be done in a single place.

philliphoff: Consolidated into a common function.

src/api/v2/ResourceGroupsExtensionManager.ts
philliphoff: Updated.

src/tree/v2/ResourceTreeDataProviderBase.ts
philliphoff: I've always preferred to keep more common types in the parent folder, with more specific types in child folders.  I suppose that's more C# style (due to its automatic namespacing), which would otherwise be awkward if the common types ended up in a sibling folder.  I feel like the suggestion might result in a number of `common` folders, which could be confusing, or require a bunch of thought about whether any given type belongs in the parent or a common folder.

src/api/v2/WorkspaceResourceProviderManager.ts
philliphoff: I've found a way to consolidate the implementations into a single base class.

src/api/v2/ApplicationResourceProviderManager.ts
philliphoff: I've found a way to consolidate the implementations into a single base class.

src/api/v2/ResourceGroupsExtensionManager.ts
philliphoff: I don't disagree necessarily, but I think I'd like to address this as a follow up PR so as to not get mixed up with all the rest, and give me time to think about this.

src/tree/v2/ResourceBranchDataProviderManagerBase.ts
philliphoff: As mentioned above, I'd like to address this as a separate PR if necessary.  I'd like to make sure that I fully understand how the "normalized" `AzExtResourceType` works (e.g. for new resource types).

src/api/v2/ResourceGroupsExtensionManager.ts
```diff
@@ -42,29 +47,29 @@ export class ResourceGroupsExtensionManager {
     }
 
     async activateApplicationResourceProviders(): Promise<void> {
-        const inActiveResourceContributors =
+        const inactiveResourceContributors =
             vscode.extensions.all
```
bwateratmsft, line 51: It might make little difference on performance, but VSCode has a fairly large number of built-in extensions (with publisher `vscode`) that would never be contributing things to this view. Should we add one extra `.filter()` statement to exclude those, to save a little time spent looking into package manifests? Basically equivalent to:
https://github.com/microsoft/vscode-azureresourcegroups/blob/main/src/utils/ExtensionActivationManager.ts#L53

Also, can we flip the `.filter()` for inactive to _before_ the mapping, also for the slight perf gain?

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -89,6 +93,14 @@ export interface WrappedResourceModel {
  * The interface that resource resolvers must implement
  */
 export interface BranchDataProvider<TResource extends ResourceBase, TModel extends ResourceModelBase> extends vscode.TreeDataProvider<TModel> {
+    /**
+     * Get the children of `element` or root if no element is passed.
```
bwateratmsft, line 97:
```suggestion
     * Get the children of `element`.
```

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -89,6 +93,14 @@ export interface WrappedResourceModel {
...
+     *
+     * @param element The element from which the provider gets children. Unlike a traditional TreeDataProvider, this will never be `undefined`.
+     * @return Children of `element` or root if no element is passed.
```
bwateratmsft, line 100:
```suggestion
     * @return Children of `element`.
```

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -89,6 +93,14 @@ export interface WrappedResourceModel {
...
+     */
+    getChildren(element: TModel): vscode.ProviderResult<TModel[]>;
+
     /**
```
bwateratmsft, line 104: It's not necessarily required, but do we want to do this as well, to indicate that a BDP should _not_ implement `getParent()`, because it's not necessary for them to do?

```typescript
    getParent?: never;
```

src/api/v2/v2AzureResourcesApi.ts
philliphoff: Updated.

src/api/v2/v2AzureResourcesApi.ts
philliphoff: Updated.

src/api/v2/v2AzureResourcesApi.ts
philliphoff: Added it.

src/api/v2/ResourceGroupsExtensionManager.ts
philliphoff: Agree it likely won't make a significant performance difference, but makes sense just the same. Added both.

bwateratmsft (approved):

alexweininger (approved):

PR #389, opened 2022-09-16
(experiment) Implement view properties command

alexweininger:
Just putting this up as an alternative to https://github.com/microsoft/vscode-azureresourcegroups/pull/383. Mostly for discussion purposes.

The way this works is that any element that defines the `viewProperties` property, will have the `canViewProperties` context value automatically added to it. And then the "View properties" context menu item will appear on that tree item.


src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -44,6 +44,8 @@ export interface ApplicationResource extends ResourceBase {
     readonly tags?: {
         [propertyName: string]: string;
     };
+    /** A copy of the raw resource intended for the viewProperties command */
+    _raw: Record<string, unknown>;
```
bwateratmsft, line 48: A couple of thoughts.

1. Can we rename to `raw`? This doesn't need to be regarded as a private property.
2. Will extensions want this View Properties command to show up for subtree items they own? If so, we should move this field to `ResourceBase`.

src/commands/registerCommands.ts
```diff
@@ -36,7 +36,7 @@ export function registerCommands(refreshEventEmitter: vscode.EventEmitter<void>)
     registerCommand('azureResourceGroups.refresh', async (context: IActionContext, node?: AzExtTreeItem) => { await ext.appResourceTree.refresh(context, node); refreshEventEmitter.fire(); });
     registerCommand('azureResourceGroups.revealResource', revealResource);
     registerCommand('azureResourceGroups.selectSubscriptions', () => commands.executeCommand('azure-account.selectSubscriptions'));
-    registerCommand('azureResourceGroups.viewProperties', viewProperties);
+    registerCommandWithTreeNodeUnwrapping('azureResourceGroups.viewProperties', viewProperties);
```
bwateratmsft, line 39: Does this necessarily need unwrapping, or can `BranchDataItem` just have `get` methods that reflect the necessary data?

src/tree/v2/ResourceGroupsTreeDataProvider.ts
```diff
@@ -78,8 +79,19 @@ export class ResourceGroupsTreeDataProvider extends vscode.Disposable implements
...
+    async getTreeItem(element: ResourceGroupsItem): Promise<vscode.TreeItem> {
+
+        const contextValuesToAdd: string[] = [];
+        if (canViewProperties(element)) {
+            contextValuesToAdd.push('canViewProperties');
+        }
+
+        const ti = await element.getTreeItem();
+
+        ti.contextValue = [...(ti.contextValue?.split(';') ?? []), ...contextValuesToAdd].sort().join(';');
```
bwateratmsft, line 91: Doesn't have to be in this PR, but we definitely need a class for consistently handling context values. I don't like having to do all kinds of sort/split/join/etc. all over the place.

@philliphoff this makes me wonder though; we maybe don't need the `quickPickOptions` field at all on `ResourceModelBase`. We are _already_ calling `getTreeItem` on each element (for presenting in the quick pick), which is going to give us whether or not it's a leaf node (via the `collapsibleState` field, same as what VSCode uses to decide if it's a leaf or not), and the context value as a string (in the `contextValue` field). `getTreeItem`'s implementation should multiplex over to the BDPs when needed, so we'll be getting all the context value(s) that the client extension wants.

From there, I think it's _reasonable_ to:
1. Split
2. Dedupe
3. Add additional stuff like `canViewProperties`
4. Sort
5. Join and give VSCode a single string

This carries the disadvantage that there's an implicit contract that `contextValue` is a delimiter-separated string. We could override VSCode's default `TreeItem` to have it be a string array, thus making it explicit, or just live with the implicit contract. What do you think?

bwateratmsft: Also @alexweininger doing the above approach probably eliminates a lot of the compat code since, from the flow of the quick picker, everything is just a VSCode `TreeItem`.

src/api/v2/v2AzureResourcesApi.ts
alexweininger: > Can we rename to `raw`? This doesn't need to be regarded as a private property.

Sure

> Will extensions want this View Properties command to show up for subtree items they own? If so, we should move this field to ResourceBase.

I thought about this but wasn't sure. Sounds like a good idea and we could remove some code in the client extensions. However, some client extensions have async view properties commands that fetch additional data before showing the properties. So not sure if we'd want to try to handle that vs. just letting the clients control all items they own.

bwateratmsft: For deeply custom behavior I think they can just DIY it. I bet it's sufficient most of the time to kick off loading that data and getting it attached to `.raw`, ignoring the small chance of the user right click -> View Properties before the load is finished.

alexweininger:
Closing in favor of https://github.com/microsoft/vscode-azureresourcegroups/pull/427

PR #395, opened 2022-10-05
Implement tree item pinning

bwateratmsft:
TODO: add some screenshots once I have IDs and the build is successful

TODO: is it possible to make the inline icon always visible for the "unpin" action? It seems like most of the time (across the software world in general, not necessarily just VSCode), when something is pinned, that icon is always visible.


alexweininger (commented):
Looks good to me. Should we revisit once we have tree item IDs figured out?

bwateratmsft:
> Looks good to me. Should we revisit once we have tree item IDs figured out?

Yep, I think that makes sense. @philliphoff do you have any plans on the IDs yet?

nturinski:
> Looks good to me. Should we revisit once we have tree item IDs figured out?

With [this](https://github.com/microsoft/vscode-azureresourcegroups/pull/422) being merged, should we revisit this PR now?

alexweininger:
> > Looks good to me. Should we revisit once we have tree item IDs figured out?
> 
> With [this](https://github.com/microsoft/vscode-azureresourcegroups/pull/422) being merged, should we revisit this PR now?

Yes

src/commands/explorer/pinning.ts
```diff
@@ -0,0 +1,43 @@
...
+import { ResourceGroupsItem } from '../../tree/v2/ResourceGroupsItem';
+import { settingUtils } from '../../utils/settingUtils';
+
+export async function pinTreeItem(_context: IActionContext, treeItem: ResourceGroupsItem): Promise<void> {
+    const memento = getPersistenceMemento();
+    const pinnedItems = memento.get<string[]>('pinnedTreeItems', []);
+
+    pinnedItems.push(treeItem.id); // TODO: need an ID field
+
+    await memento.update('pinnedTreeItems', pinnedItems);
```
alexweininger, line 18: We need to refresh the tree after pinning or unpinning. However, inside this function we can't tell which tree view the tree item belongs to. 

An easy solution might be to register the pin/unpin commands for each view.

bwateratmsft: Do we expect the IDs to overlap between one view and the other?

alexweininger: > Do we expect the IDs to overlap between one view and the other?

I don't think so. But I'm not quite sure what you're getting at here... 😕

bwateratmsft: If I'm not mistaken, as long as the IDs are universally unique, this approach would work for any number of nodes at any level of any number of trees.

alexweininger: > I _do_ want to make sure we scope the refreshes to the specific subset we need

Yeah, I think hope we can do that by utilizing `TreeDataProvider.getParent`. However, we'll run into issues if branch tree items want to do pinning, but the branch data provider doesn't implement `getParent`. Though I think that's fine for now since we don't plan on doing that.

bwateratmsft: Sounds good. Didn't we also attach the parent to all of RG's models? i.e. the ones it owns + the wrappers? Or was it just some of them?

alexweininger: Only nodes above the wrappers have the parent property. We could restrict the pinning stuff to only those nodes.

bwateratmsft: It sounds like we're gonna do that anyway, at least for now, so this should work 👍

src/api/compatibility/application/CompatibleApplicationResourceBranchDataProvider.ts
```diff
@@ -26,6 +26,7 @@ export class CompatibleApplicationResourceBranchDataProvider<TResource extends A
         const subscriptionContext: ISubscriptionContext = createSubscriptionContext(element.subscription);
 
         const resolved = await this.resolver.resolveResource(subscriptionContext, oldAppResource);
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
bwateratmsft, line 29: Fix for an unrelated linter error.

bwateratmsft:
Closing in favor of #679

PR #396, opened 2022-10-06
Use normalized resource types for registration

philliphoff:
Updates the V2 API and implementation to use "normalized" resource types (i.e. `AzExtResourceType`) for branch data provider registration (which follows how V1 currently operates). Previously, the Azure resource type was used which is not ideal as both type and kind(s) are needed to totally distinguish between Azure resources.

Also updates the grouping logic to better replicate the existing V1 behavior in terms of hidden types, static resource type groups, and sorting.

Also removes the notion of IDs when registering resource providers, as they weren't used.  Extensions can now indicate that they contribute application/workspace resources using a `"resources": true` property within the `application` or `workspace` property.  (The thinking was that, in the future, if we need more complex registration, the `resources` property could be expanded to allow both a `boolean` as well as array of objects.

```json
{
    "contributes": {
        "x-azResourcesV2": {
            "application": {
                "branches": [
                    {
                        "type": "StorageAccounts"
                    }
                ]
            },
            "workspace": {
                "resources": true,
                "branches": [
                    {
                        "type": "ms-azuretools.vscode-azurestorage"
                    }
                ]
            }
        }
    }
}
```

### DIFF
--- a/src/api/v2/ResourceGroupsExtensionManager.ts
+++ b/src/api/v2/ResourceGroupsExtensionManager.ts
@@ -44,12 +44,10 @@ function getInactiveExtensions(): vscode.Extension<unknown>[] {
 
 export class ResourceGroupsExtensionManager {
     async activateApplicationResourceBranchDataProvider(type: AzExtResourceType): Promise<void> {
-        const normalizedType = type.toLowerCase();
-
         const extensionAndContributions =
             getInactiveExtensions()
-                .map(extension => ({ extension, contributions: getV2ResourceContributions(extension)?.application?.branches?.map(resource => resource.type.toLowerCase()) ?? [] }))
-                .find(extensionAndContributions => extensionAndContributions.contributions.find(contribution => contribution === normalizedType) !== undefined);
+                .map(extension => ({ extension, contributions: getV2ResourceContributions(extension)?.application?.branches?.map(resource => resource.type) ?? [] }))
+                .find(extensionAndContributions => extensionAndContributions.contributions.find(contribution => contribution === type) !== undefined);
 
         if (extensionAndContributions) {
             await extensionAndContributions.extension.activate();
@@ -73,11 +71,9 @@ export class ResourceGroupsExtensionManager {
     }
 
     async activateWorkspaceResourceBranchDataProvider(type: string): Promise<void> {
-        type = type.toLowerCase();
-
         const extensionAndContributions =
             getInactiveExtensions()
-                .map(extension => ({ extension, contributions: getV2ResourceContributions(extension)?.workspace?.branches?.map(resources => resources.type.toLowerCase()) ?? [] }))
+                .map(extension => ({ extension, contributions: getV2ResourceContributions(extension)?.workspace?.branches?.map(resources => resources.type) ?? [] }))
                 .find(extensionAndContributions => extensionAndContributions.contributions.find(contribution => contribution === type) !== undefined);
 
         if (extensionAndContributions) {
--- a/src/api/v2/ResourceGroupsExtensionManager.ts
+++ b/src/api/v2/ResourceGroupsExtensionManager.ts
@@ -3,16 +3,17 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { AzExtResourceType } from '@microsoft/vscode-azext-utils';
 import * as vscode from 'vscode';
 
 interface ResourceGroupsContribution {
     readonly application: {
         readonly branches?: { type: string }[];
-        readonly resources?: { type: string }[];
+        readonly resources?: boolean;
     }
     readonly workspace: {
         readonly branches?: { type: string }[];
-        readonly resources?: { type: string }[];
+        readonly resources?: boolean;
     }
 }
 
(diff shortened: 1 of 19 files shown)


src/tree/v2/application/ApplicationResourceTreeDataProvider.ts
```diff
@@ -41,6 +43,19 @@ export class ApplicationResourceTreeDataProvider extends ResourceTreeDataProvide
                 this.statusSubscription?.dispose();
             });
 
+        registerEvent(
+            'treeView.onDidChangeConfiguration',
+            vscode.workspace.onDidChangeConfiguration,
+            async (context: IActionContext, e: vscode.ConfigurationChangeEvent) => {
+                context.errorHandling.suppressDisplay = true;
```
bwateratmsft, line 50: Any reason not to just `suppressAll` here? Do errors thrown by listeners to `this.onDidChangeTreeData` end up here?

src/tree/v2/application/ApplicationResourceGroupingManager.ts
```diff
@@ -141,11 +142,24 @@ export class ApplicationResourceGroupingManager extends vscode.Disposable {
...
+
+        // Pre-populate the initial grouping with the supported resource types...
+        azureExtensions.forEach(extension => {
+            extension.resourceTypes.forEach(resourceType => {
+                initialGrouping[resourceType] = [];
+            });
+        });
+
+        // Exclude resource groups...
+        resources = resources.filter(resource => resource.azureResourceType.type !== 'microsoft.resources/resourcegroups');
```
bwateratmsft, line 155: Don't we have an `AzExtResourceType` for this? If not we should add one.

src/api/v2/ResourceGroupsExtensionManager.ts
```diff
@@ -42,13 +43,13 @@ function getInactiveExtensions(): vscode.Extension<unknown>[] {
 }
 
 export class ResourceGroupsExtensionManager {
-    async activateApplicationResourceBranchDataProvider(type: string): Promise<void> {
-        type = type.toLowerCase();
+    async activateApplicationResourceBranchDataProvider(type: AzExtResourceType): Promise<void> {
+        const normalizedType = type.toLowerCase();
```
bwateratmsft, line 47: This step isn't desirable. These are strings but they should be treated as constants.

src/tree/v2/application/SubscriptionItem.ts
```diff
@@ -20,7 +29,13 @@ export class SubscriptionItem implements ResourceGroupsItem {
     }
 
     async getChildren(): Promise<ResourceGroupsItem[]> {
-        const resources = await this.resourceProviderManager.getResources(this.subscription);
+        let resources = await this.resourceProviderManager.getResources(this.subscription);
+
+        const showHiddenTypes = settingUtils.getWorkspaceSetting<boolean>(showHiddenTypesSettingKey);
+
+        if (!showHiddenTypes) {
+            resources = resources.filter(resource => resource.azureResourceType.type === 'microsoft.resources/resourcegroups' || (resource.resourceType && supportedResourceTypes.find(type => type === resource.resourceType)));
```
bwateratmsft, line 37: (Same as above) Do we not have an `AzExtResourceType` for RGs?

src/tree/v2/application/ApplicationResourceGroupingManager.ts
philliphoff: I guess it depends on whether we want to treat resource groups as first-class Azure resources.  The existing tree view does not.

src/api/v2/ResourceGroupsExtensionManager.ts
philliphoff: Within the context of *this* extension, sure.  This is just an attempt at being reasonably permissive with respect to how other extensions may register their BDPs in *their* `package.json`.  We *could* be super strict about matching the exact case of the resource type, but do we need to be?

src/tree/v2/application/ApplicationResourceGroupingManager.ts
bwateratmsft: I don't think we have to treat them as first-class while still assigning them a resource type

src/api/v2/ResourceGroupsExtensionManager.ts
bwateratmsft: If we aren't strict, I can see us quickly finding our way back to always having to do `toLowerCase()` all over the place again.

src/tree/v2/application/ApplicationResourceGroupingManager.ts
philliphoff: Ok, added an [issue](https://github.com/microsoft/vscode-azuretools/issues/1244) for them to be added to the enumeration.

src/tree/v2/application/SubscriptionItem.ts
philliphoff: Not yet, but added an [issue](https://github.com/microsoft/vscode-azuretools/issues/1244) for them to be added to the enumeration.

src/api/v2/ResourceGroupsExtensionManager.ts
philliphoff: I don't think normalizing at the "human" boundaries is a bad thing, or will necessarily result in `toLowerCase()` further infiltrating the codebase.  That said, it's something that can be added if/when we find those humans insisting on their own arbitrary casing and I just don't feel strongly about it either way, so I've reverted it.

bwateratmsft (approved):

PR #401, opened 2022-10-11
Model for providing branch data provider item IDs

philliphoff:
Allow BDPs to provide an ID for their items, which can then be accumulated to represent a "path" through the tree.

Every application/workspace item will have an ID of some sort.  BDPs can provide IDs but, if omitted, reasonable default IDs will be chosen (e.g. application resource items will use their Azure resource IDs, generic items will use their label, all others will be assigned a unique ID.)

The `ResourceGroupsItemCache` can be used to generate item "paths" (an array that contains the IDs of each item in the tree, in order, from the root item to the item in question) via `getPathForItem()` or retrieve the item represented by a path via `getItemForPath()`.

### DIFF
--- a/src/api/v2/v2AzureResourcesApi.ts
+++ b/src/api/v2/v2AzureResourcesApi.ts
@@ -73,6 +73,8 @@ export interface ResourceQuickPickOptions {
 }
 
 export interface ResourceModelBase {
+    // TODO: Should IDs be optional?
+    readonly id?: string;
     readonly quickPickOptions?: ResourceQuickPickOptions;
     readonly azureResourceId?: string;
 }
--- a/src/tree/v2/BranchDataProviderItem.ts
+++ b/src/tree/v2/BranchDataProviderItem.ts
@@ -23,7 +23,7 @@ export class BranchDataProviderItem implements ResourceGroupsItem, WrappedResour
         itemCache.addBranchItem(this.branchItem, this);
     }
 
-    readonly id: string = this?.options?.defaultId ?? this.branchItem.id ?? randomUUID();
+    readonly id: string = this.branchItem.id ?? this?.options?.defaultId ?? randomUUID();
 
     async getChildren(): Promise<ResourceGroupsItem[] | undefined> {
         const children = await this.branchDataProvider.getChildren(this.branchItem);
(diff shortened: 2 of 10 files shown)


src/tree/v2/BranchDataProviderItem.ts
```diff
@@ -21,6 +23,8 @@ export class BranchDataProviderItem implements ResourceGroupsItem, WrappedResour
         itemCache.addBranchItem(this.branchItem, this);
     }
 
+    readonly id: string = this?.options?.defaultId ?? this.branchItem.id ?? randomUUID();
```
bwateratmsft, line 26: Should `branchItem.id` override `defaultId`?

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -73,6 +73,8 @@ export interface ResourceQuickPickOptions {
 }
 
 export interface ResourceModelBase {
+    // TODO: Should IDs be optional?
```
bwateratmsft, line 76: I'd love if we could keep this optional. Label is a required result from `getTreeItem`, perhaps we can use that if nothing else is specified? VSCode uses the label as (at least part of) the ID if it isn't specified, though I've not seen the actual code there.

bwateratmsft: A bonus to keeping this and the other values optional is that the model objects don't necessarily have to be `object`s. They could be e.g. `string`. I've really hoped we could keep that true.

(As a related aside, I think when applicable, e.g. when we're checking if the model object has an `azureResourceId` or `azurePortalUrl` or whatever, we should first check if the model object is actually an `object`--make no assumptions about what those models are)

philliphoff: I can't think of a reason why ID *must* be specified; i.e. why we can't just choose an arbitrary unique value if omitted.  It's obviously *better* if BDPs use proper IDs, but there are definately cases where it's just unlikely to matter (e.g. generic nodes, error nodes, etc.).

We can't use the tree item ID--at least, not generally--as we don't necessarily have the tree item, as we're working solely with model objects.  It would probably make sense, if BDPs give their tree items IDs, that they use the same ID as the model itself.  If we let VS Code be the sole interpretor of tree item IDs, then we could leave that as a "best practice" but not necessarily care.

bwateratmsft (approved):

PR #402, opened 2022-10-12
Expose `TreeDataProvider`s in API

alexweininger:
* Exposing the workspace and application tree data providers so they can be used by client extensions.
* Some cleanup and renaming around the registration functions

### DIFF
--- a/src/api/v2/v2AzureResourcesApi.ts
+++ b/src/api/v2/v2AzureResourcesApi.ts
@@ -179,6 +179,10 @@ export interface ResourcePickOptions {
  * The current (v2) Azure Resources extension API.
  */
 export interface V2AzureResourcesApi extends AzureResourcesApiBase {
+
+    applicationResourceTreeDataProvider: vscode.TreeDataProvider<unknown>;
+    workspaceResourceTreeDataProvider: vscode.TreeDataProvider<unknown>;
+
     /**
      * Reveals an item in the application/workspace resource tree
      * @param resourceId The ID of the resource to reveal.
--- a/src/api/v2/v2AzureResourcesApi.ts
+++ b/src/api/v2/v2AzureResourcesApi.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import type { Environment } from '@azure/ms-rest-azure-env';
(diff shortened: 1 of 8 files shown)


src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -179,6 +179,10 @@ export interface ResourcePickOptions {
  * The current (v2) Azure Resources extension API.
  */
 export interface V2AzureResourcesApi extends AzureResourcesApiBase {
+
+    applicationResourceTreeDataProvider: vscode.TreeDataProvider<unknown>;
```
bwateratmsft, line 183: For the purposes of the API, these can probably be `vscode.TreeDataProvider<ResourceGroupsItem>` or something along those lines.

src/api/v2/v2AzureResourcesApiWrapper.ts
```diff
@@ -17,6 +17,14 @@ export class V2AzureResourcesApiWrapper implements V2AzureResourcesApi {
         return this.api.apiVersion;
     }
 
+    get applicationResourceTreeDataProvider(): vscode.TreeDataProvider<unknown> {
+        return this.callWithTelemetryAndErrorHandlingSync('v2.getApplicationResourceTreeDataProvider', () => this.api.applicationResourceTreeDataProvider);
```
bwateratmsft, line 21: Does a simple `get` method really deserve telemetry wrapping?

src/api/v2/v2AzureResourcesApi.ts
alexweininger: Ok. I wasn't sure how much we wanted to hide from the clients.

src/api/v2/v2AzureResourcesApiWrapper.ts
alexweininger: Ya... it's probably overkill

src/api/v2/v2AzureResourcesApi.ts
bwateratmsft: I think they'll almost _need_ at least something, e.g. to indicate that the objects are wrappers. What did we say as the input to the quick pick experience? Isn't that `ResourceGroupsItem`?

alexweininger: Currently the quick pick experience needs a `TreeDataProvider<unknown>`

PR #412, opened 2022-10-26
WIP: (V2) Workspace Welcome Views

philliphoff:
So, it bothers me when nothing is shown in the (V2) workspace view as I can never be sure it is intentional or not.  VS Code has the ability to show "welcome views" when tree views are otherwise empty.  This change adds welcome views for the workspace view that differ when:

- No workspace is open

Screenshot 2022-10-26 at 12 26 18

- No workspace resources were enumerated

Screenshot 2022-10-26 at 12 27 30

- No workspace resource providers are registered (i.e. no extension that registers a provider is installed)

Screenshot 2022-10-26 at 12 26 46

- Resources were enumerated (i.e. the traditional view)

Screenshot 2022-10-26 at 12 29 05

The text content is mostly a placeholder at this point; I'd appreciate some "PM wordsmithing".  In particular, would it make sense to try to point users somewhere, for example, when no resource providers have been installed?

### DIFF
--- a/package.json
+++ b/package.json
@@ -252,6 +252,16 @@
             {
                 "view": "azureActivityLog",
                 "contents": "View all of your recent activities and quickly access resources you've recently created. \n [Create an Azure resource](command:azureResourceGroups.createResource)"
+            },
+            {
+                "view": "azureWorkspace",
+                "contents": "No workspace has been opened.",
+                "when": "azureWorkspace.state == 'noWorkspace'"
+            },
+            {
+                "view": "azureWorkspace",
+                "contents": "No local workspace resources exist.",
+                "when": "azureWorkspace.state == 'noWorkspaceResources'"
             }
         ],
         "menus": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -252,6 +252,21 @@
             {
                 "view": "azureActivityLog",
                 "contents": "View all of your recent activities and quickly access resources you've recently created. \n [Create an Azure resource](command:azureResourceGroups.createResource)"
+            },
+            {
+                "view": "azureWorkspace",
+                "contents": "No workspace has been opened.",
+                "when": "azureWorkspace.state == 'noWorkspace'"
```
bwateratmsft, line 259: Can you use the builtin `workspaceFolderCount` to replace this? e.g. `"when": "workspaceFolderCount < 1"`? Or is the goal just consistency with the other two?

bwateratmsft (approved):

package.json
philliphoff: I didn't realize `workspaceFolderCount` existed, actually. :-) Still, I think I would prefer the consistency of using a specific state value.

PR #413, opened 2022-10-28
Follow Azure resource id pattern

alexweininger:
`findTreeItem` (at least the version we pass v1.5 client extensions) needs to use caching so that it can return references to tree items on the actual tree. If we don't cache tree items, and always create new items then it breaks quite a few things in the client extensions.

Since it needs to use the cache, can we just implement this as a method on the TDP instead of `findByIdExperience`? It has to access the internal cache and doesn't quite fit with the `_Experience` pattern anymore.

Tested the following uses of `findTreeItem` work:
* loadPersistedPostgresDatabase which passes a tree item ID
* getDeployNode which passes an Azure resource ID to deploy to


src/tree/v2/ResourceTreeDataProviderBase.ts
```diff
@@ -89,9 +93,59 @@ export abstract class ResourceTreeDataProviderBase extends vscode.Disposable imp
...
+        }
+        while (parent) {
+            let id = parent.id;
+            if (!id.startsWith('/')) {
+                id = `/${id}`;
+            }
+            // don't include grouping item IDs to the full ID
+            if (!id.startsWith('/groupings')) {
+                fullId = id + fullId;
+            }
```
alexweininger, line 145: I special cased this for now. We could add a standard way for items to declare that their id should not contribute to the fullId of child items, but for now this works.

alexweininger:
Closing in favor of #418

PR #418, opened 2022-11-01
Follow Azure id format, implement findTreeItem

alexweininger:
In this PR:
* Made tree item ids mimic Azure ids (just like v1  and pre-unified)
* Implemented a find tree item method on the TDP.

I have implemented find item without getting children from the cache. I did this so we could update and use `findByIdExperience` in utils since it won't have access to the cache. However, without using the cache to get children, find item is _extremely_ slow. For reveal, this hurts the UX quite a bit. I'm open to discussions here, but I think we have to use the cache since reveal (relies on find tree item most times) is just too slow.

PR unblocks #395


alexweininger:
Closing as Azure ids have been implemented in #422, and find is implemented in #431

Issue #416, opened 2022-10-31
`findTreeItem` fails for an app resource child contributed by a yet-to-be-activated extension


alexweininger:
Databases will persist the connected database for each workspace. On activation, if the workspace has a connected database saved it does `findTreeItem` to get the tree item and then connects to it.

This specific use case of `findTreeItem` fails fairly consistently. Since the tree item we're looking for is a child of an application resource, when `findTreeItem` is executed, it's success depends on if the tree has updated to use the newly registered Databases branch data provider to get the tree items (rather than the `DefaultApplicationResourceBranchDataProvider`). Since if the tree still has the tree items given by `DefaultApplicationResourceBranchDataProvider`, they won't have children.

Not sure what the best solution is to this. Should we have mechanisms in RGs to solve this? This could be a common scenario, basically anytime you try to do `findTreeItem` for a tree item that is contributed by a yet-to-be-activated extension.

nturinski:
Doesn't the tree item have to be resolved before we can invoke `findTreeItem` on it? Maybe we can check the `azExtResourceType` property on the tree item and make sure the appropriate extension is activated before we look at which resolver to use?

alexweininger:
> Doesn't the tree item have to be resolved before we can invoke `findTreeItem` on it?

`findTreeItem` takes a tree item id, so you can call it whenever/however you want - even if the items haven't been resolved yet.

> EDIT: I'm not entirely what the new implementation of findTreeItem is, but I assume it still has to travel down the tree (eventually hitting AppResourceTreeItem). If that's not the case, this wouldn't work whatsoever.

Correct, however we need this functionality for at least a few features I know of, and probably a lot more that I haven't thought of. 

Calling `findTreeItem` to reveal a tree item that hasn't yet been displayed is pretty common. And with how the unified view works by lazy activating extensions, this scenario is probably pretty common. 

I think one solution is if an extension contributes a BDP, to not show shallow tree items for those resources. Instead, just wait for the extension to activate and show the resolved tree items. I don't really like how it behaves anyway.

Video of what I'm talking about:
https://user-images.githubusercontent.com/12476526/199114378-245102a4-2483-4579-ae25-37f6926df20c.mov

alexweininger:
Note: this only reproduces for Databases if the extension hasn't activated when user expands the group tree item.

bwateratmsft:
[...]

PR #420, opened 2022-11-02
Implement reveal

### LINKED ISSUE #393: An error occurs when clicking "Click to view resource" without Azure view opening

alexweininger:
We have to do some additional acrobatics 🥷  to make reveal work consistently.

For one, if `onDidChangeTreeData` is fired while reveal is executing, reveal will fail. When `onDidChangeTreeData` is fired, VS Code's internal cache is cleared right before it attempts to reveal, and reveal fails since VS Code no longer has a reference to the element to reveal. This might be a VS Code issue, but it's fixable for now by blocking firing that event until reveal has finished executing.

I've implemented so that `onDidChangeTreeData` events fired while reveal is executing are deferred until reveal finishes.

We expose application and workspace resource tree views to the client extensions. Client extensions use `TreeView.reveal` directly. To apply this custom logic around reveal, I had to directly modify the `TreeView.reveal` method.

Fixes #393 🥳 

Also, having a little dilemma...

Find tree item is super slow if it calls `getChildren` instead of getting the children from the cache. This results in reveal being super slow too since they work in tandem in many cases. I measured reveal taking 2s when using the cache, and 15s without using it. 15s is long enough users will assume the operation failed IMO.

Well Alex why don't you just use get the children from the cache then?... 😄

A few issues arise:

A. Fetching the children from the cache causes #416. When the cache isn't used, #416 does not reproduce.

B. Ideally the only thing find tree item needs to work is a `vscode.TreeDataProvider<ResourceGroupsItem>`. If we need to use the cache in find tree item, then we'll need a reference to it and expose that in the API. We'd like to avoid this as we want to keep the API as simple as possible. And as close to VS Code's API as possible.

For now I'm using the cache because I don't think we can allow reveal to be that slow, but I'm open to opinions and suggestions here.


src/tree/v2/application/registerResourceGroupsTreeV2.ts
```diff
@@ -35,21 +36,23 @@ export function registerApplicationTree(context: vscode.ExtensionContext, option
 
     context.subscriptions.push(resourceGroupingManager);
 
-    const applicationResourceTreeDataProvider = new ApplicationResourceTreeDataProvider(branchDataProviderManager.onDidChangeTreeData, itemCache, refreshEvent, resourceGroupingManager, resourceProviderManager);
+    const applicationResourceTreeDataProvider = new ApplicationResourceTreeDataProvider(
+        branchDataProviderManager.onDidChangeTreeData,
+        itemCache, refreshEvent,
```
bwateratmsft, line 41:
```suggestion
        itemCache,
        refreshEvent,
```

bwateratmsft:
> Find tree item is super slow if it calls getChildren instead of getting the children from the cache. This results in reveal being super slow too since they work in tandem in many cases. I measured reveal taking 2s when using the cache, and 15s without using it. 15s is long enough users will assume the operation failed IMO.

🤮 

@alexweininger we need to find out what the heck is taking so long. I'd expect two seconds to be enough time to enumerate the whole tree top to bottom. Fifteen is just unacceptable.

alexweininger:
> 🤮
> 
> @alexweininger we need to find out what the heck is taking so long. I'd expect two seconds to be enough time to enumerate the whole tree top to bottom. Fifteen is just unacceptable.

I have 15 App Service apps in my subscription, each App Service takes ~700ms to resolve (GET request). That's at least 10s. Also I think 15 resources is a realistic number.

bwateratmsft:
Agreed, 15 resources is reasonable. We need to make some changes then. The calls should be able to be "parallel"ized (inasmuch as that's a thing when there's only one thread). That should help a lot. Additionally, the less we can do during resolve, the better.

alexweininger:
> Agreed, 15 resources is reasonable. We need to make some changes then. The calls should be able to be "parallel"ized (inasmuch as that's a thing when there's only one thread). That should help a lot since these things are all network-bound, not CPU-bound. Additionally, the less we can do during resolve, the better.

For a group, we're already using `Promise.all` for concurrency when calling `getResourceItem`. So I don't think there's much more optimization to do there.

However, inside of `getResourceItem` instead of making individual GET requests we can make GET request to list resources, which should help significantly for when groups have multiple resources of the same type.

We could change `getResourceItem` from 

```ts
getResourceItem(element: TResource): TModel | Thenable<TModel>;
```

to 

```ts
getResourceItem(elements: TResource[]): TModel[] | Thenable<TModel[]>;
```

We could also make list the resources on the first `getResourceItem` call, and then use the cached result on subsequent calls. However other than setting an arbitrary cache invalidation time, I'm not sure how to properly invalidate this cache.

bwateratmsft:
I don't want to change `getResourceItem`. I'd rather fix the client extensions to be less awful. Though I'm surprised that even with concurrency it's taking this long.

alexweininger:
Converting to draft until find is finalized.

alexweininger:
Closing as stale and replaced by newer PRs

PR #422, opened 2022-11-02
Make tree item ids mimic Azure ids

alexweininger:
Cherry-picked non-compatibility related changes from #418.

Changed how we cache child -> parent relations. Instead of a map, I'm storing a reference to the parent on the child item itself. This way, no cache modification can accidentally remove a reference.

### DIFF
--- a/src/tree/v2/ResourceGroupsItem.ts
+++ b/src/tree/v2/ResourceGroupsItem.ts
@@ -8,13 +8,6 @@ import * as vscode from 'vscode';
 export interface ResourceGroupsItem {
     readonly id: string;
 
-    /**
-    * Override default isAncestorOf behavior based on paths.
-    *
-    * If defined, the item id will be excluded from paths of child items
-    */
-    isAncestorOf?(id: string): boolean;
-
     getChildren(): vscode.ProviderResult<ResourceGroupsItem[]>;
     getTreeItem(): vscode.TreeItem | Thenable<vscode.TreeItem>;
 }
--- a/src/tree/v2/ResourceGroupsItem.ts
+++ b/src/tree/v2/ResourceGroupsItem.ts
@@ -8,6 +8,13 @@ import * as vscode from 'vscode';
 export interface ResourceGroupsItem {
     readonly id: string;
 
+    /**
+    * Override default isAncestorOf behavior based on paths.
+    *
+    * If defined, the item id will be excluded from paths of child items
+    */
+    isAncestorOf?(id: string): boolean;
+
     getChildren(): vscode.ProviderResult<ResourceGroupsItem[]>;
     getTreeItem(): vscode.TreeItem | Thenable<vscode.TreeItem>;
 }
--- a/src/tree/v2/ResourceGroupsItemCache.ts
+++ b/src/tree/v2/ResourceGroupsItemCache.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { localize } from '../../utils/localize';
+import { GroupingItem } from './application/GroupingItem';
 import { ResourceGroupsItem } from './ResourceGroupsItem';
 
 interface InternalResourceGroupsItem extends ResourceGroupsItem {
@@ -91,8 +92,7 @@ export class ResourceGroupsItemCache {
         while (currentItem) {
             const nextItem = this.getParentForItem(currentItem);
 
-            // if item has custom ancestor logic, exclude its id from the path of other items
-            if (!currentItem.isAncestorOf || item === currentItem) {
+            if (!currentItem.id.includes('groupings') || item === currentItem) {
                 path.push(currentItem.id);
             }
 
@@ -129,7 +129,10 @@ export class ResourceGroupsItemCache {
     }
 
     isAncestorOf(element: ResourceGroupsItem, id: string): boolean {
-        return element.isAncestorOf?.(id) ?? id.startsWith(this.getId(element) + '/');
+        if (element instanceof GroupingItem) {
+            return element.resources.some(resource => id.startsWith(resource.id));
+        }
+        return id.startsWith(this.getId(element) + '/');
     }
 
     private createInternalResourceGroupsItem(child: ResourceGroupsItem, parent: ResourceGroupsItem): InternalResourceGroupsItem {
--- a/src/tree/v2/ResourceGroupsItemCache.ts
+++ b/src/tree/v2/ResourceGroupsItemCache.ts
@@ -79,13 +79,6 @@ export class ResourceGroupsItemCache {
         return this.branchItemToItemCache.get(branchItem);
     }
 
-    getChildrenForItem(item?: ResourceGroupsItem): ResourceGroupsItem[] | undefined {
-        if (!item) {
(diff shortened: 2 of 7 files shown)


src/tree/v2/ResourceGroupsItemCache.ts
```diff
@@ -109,8 +118,17 @@ export class ResourceGroupsItemCache {
         return currentItem;
     }
 
-    updateItemChildren(item: ResourceGroupsItem, children: ResourceGroupsItem[]): void {
+    updateItemChildren(item: ResourceGroupsItem, children: ResourceGroupsItem[]): InternalResourceGroupsItem[] {
         this.itemToChildrenCache.set(item, children);
-        children.forEach(child => this.itemToParentCache.set(child, item));
+        // cache the parent on the item
+        return children.map(child => Object.assign(child, { parent: item }));
```
bwateratmsft, line 124: We should just make it writable or add a method to do so, I think that's cleaner than this `Object.assign()`

src/utils/v2/getApplicationResourceId.ts
```diff
@@ -0,0 +1,13 @@
...
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+/**
+ * @param resourceId Azure resource Id
+ * @returns Azure resource id minus the subscription part
+ */
+export function getApplicationResourceId(resourceId: string): string {
+    const splitId = resourceId.split(/(resourceGroups)/);
```
bwateratmsft, line 11: Does `resourceGroups` need to be in its own regex group?

src/tree/v2/application/GroupingItem.ts
```diff
@@ -34,7 +34,11 @@ export class GroupingItem implements ResourceGroupsItem {
         public readonly resources: ApplicationResource[]) {
     }
 
-    readonly id: string = this.label;
+    readonly id: string = `groupings/${this.label}`;
+
+    isAncestorOf(id: string): boolean {
+        return this.resources.some(resource => id === resource.id || id.startsWith(resource.id));
```
bwateratmsft, line 40: `startsWith` will still return true for `id === resourceId`, so technically that first equality comparison isn't needed. Though, having it adds some clarity (i.e., I had to go look and see if `startsWith` actually did that or not). I'm fine with either.

src/tree/v2/ResourceTreeDataProviderBase.ts
```diff
@@ -63,8 +63,13 @@ export abstract class ResourceTreeDataProviderBase extends vscode.Disposable imp
 
     onDidChangeTreeData: vscode.Event<void | ResourceGroupsItem | ResourceGroupsItem[] | null | undefined> = this.onDidChangeTreeDataEmitter.event;
 
-    getTreeItem(element: ResourceGroupsItem): vscode.TreeItem | Thenable<vscode.TreeItem> {
-        return element.getTreeItem();
+    async getTreeItem(element: ResourceGroupsItem): Promise<vscode.TreeItem> {
+        const treeItem = await element.getTreeItem();
+        treeItem.id = this.itemCache.getId(element);
+
+        // TODO: remove this when we're done working with ids
```
bwateratmsft, line 70: Should this be done before or after this PR is merged?

src/tree/v2/ResourceGroupsItemCache.ts
```diff
@@ -109,8 +118,17 @@ export class ResourceGroupsItemCache {
...
+    }
+
+    getId(element: ResourceGroupsItem): string {
+        return '/' + this.getPathForItem(element).join('/');
+    }
+
+    isAncestorOf(element: ResourceGroupsItem, id: string): boolean {
+        return element?.isAncestorOf?.(id) || id.startsWith(this.getId(element) + '/');
```
bwateratmsft, line 132: We should check `isAncestorOf` a little more specifically here. If it is defined but returns `false`, the `||` will still move forward with checking `id.startsWith(...)`. I'm guessing the intent is that if it is defined, then _only_ its result is considered, not the `id.startsWith()` check.

Something like this I think is more appropriate:

```typescript
if (element.isAncestorOf) { // Or: `if (typeof element.isAncestorOf === 'function')` if we need to be more resilient, but I don't think we necessarily do here
    return element.isAncestorOf(id);
} else {
    return id.startsWith(this.getId(element) + '/');
}
```

src/tree/v2/ResourceGroupsItem.ts
```diff
@@ -8,6 +8,13 @@ import * as vscode from 'vscode';
 export interface ResourceGroupsItem {
     readonly id: string;
 
+    /**
+    * Override default isAncestorOf behavior based on paths.
+    *
+    * If defined, the item id will be excluded from paths of child items
```
bwateratmsft, line 14: Is it really necessary to alter the path based on whether `isAncestorOf` is defined?

alexweininger: This is used to exclude the `GroupingItem` ids from the paths of their children. I initially had it special cased, but changed it to be based off of if `isAncestorOf` is defined.

I can't foresee us needing to use it for anything else going forward, so it might make sense to special case it.

alexweininger: Correct

src/utils/v2/getApplicationResourceId.ts
alexweininger: Yes, that way it gets included in the resulting array.

src/tree/v2/ResourceGroupsItemCache.ts
alexweininger: Good catch. I think

```ts
return element.isAncestorOf?.(id) ?? id.startsWith(this.getId(element) + '/');
```

will work the same as your suggestion.

src/tree/v2/ResourceTreeDataProviderBase.ts
alexweininger: I would like to keep it around until reveal and find are merged. I've experienced issues where the cache is cleared in the middle of a `getId` call and the id returned is wrong. Although I've fixed this issue by caching the parent on the items, putting the id here makes debugging a lot easier in case I stumble upon a similar issue later on.

src/tree/v2/ResourceGroupsItemCache.ts
bwateratmsft: Yup, I'd temporarily forgotten about `??`.

src/tree/v2/ResourceGroupsItem.ts
alexweininger: Did you want to special case this?

bwateratmsft: Yeah, I think that's the best approach.

alexweininger: Ok, since this is the only use of `isAncestorOf`, I will just special case it all and remove `isAncestorOf`.

bwateratmsft (approved):

PR #423, opened 2022-11-07
Expose registerActivity() from V2 API.

philliphoff:
Expose the `registerActivity()` function from the V2 API. This eliminates the need for extensions to obtain references to both V1 and V2 APIs.

### DIFF
--- a/src/api/v2/v2AzureResourcesApi.ts
+++ b/src/api/v2/v2AzureResourcesApi.ts
@@ -5,6 +5,7 @@
 
 import type { Environment } from '@azure/ms-rest-azure-env';
 import { AzExtResourceType } from '@microsoft/vscode-azext-utils';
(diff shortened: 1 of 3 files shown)


src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -5,6 +5,7 @@
 
 import type { Environment } from '@azure/ms-rest-azure-env';
 import { AzExtResourceType } from '@microsoft/vscode-azext-utils';
+import { Activity } from '@microsoft/vscode-azext-utils/hostapi';
```
alexweininger, line 8:
```suggestion
import type { Activity } from '@microsoft/vscode-azext-utils/hostapi';
```

alexweininger (approved):

PR #426, opened 2022-11-09
Add `getParent` method to `BranchDataProvider` interface

alexweininger:
Previously, the Resource Groups Tree Data Provider was caching child -> parent relationships for all items so it could do two things:
1. generate a path based id for an item and attach it to the `vscode.TreeItem` in `getTreeItem`.
2. implement `getParent`

This change breaks up this responsibility so that whatever `___DataProvider` owns the item is also responsible for implementing `getParent` and generating TreeItem ids. Meaning the RGs TDP is responsible for caching the base items subscription items, grouping items, and application resource items). And any items beneath application resources in the tree will need to be cached (or not) by their respective `BranchDataProvider`.

To accomplish this, I've written a `CachedBranchDataProvider` class, which wraps `BranchDataProvider` and implements `getParent` and sets `TreeItem.id` by caching `getChildren` results.

In the future, client extensions can share an implementation of a cache that allows for similar features that `AzExtTreeDataProvider` had.


src/tree/v2/application/SubscriptionItem.ts
```diff
@@ -50,4 +50,8 @@ export class SubscriptionItem implements ResourceGroupsItem {
 
         return treeItem;
     }
+
+    getParent(): vscode.ProviderResult<ResourceGroupsItem> {
```
philliphoff, line 54: `getParent()` is optional in `ResourceGroupsItem`, right?  If so, then I think this definition isn't needed.

src/tree/v2/ResourceTreeDataProviderBase.ts
```diff
@@ -65,7 +66,10 @@ export abstract class ResourceTreeDataProviderBase extends vscode.Disposable imp
 
     async getTreeItem(element: ResourceGroupsItem): Promise<vscode.TreeItem> {
         const treeItem = await element.getTreeItem();
-        treeItem.id = this.itemCache.getId(element);
+
+        if (!(element instanceof BranchDataProviderItem)) {
+            treeItem.id = this.itemCache.getId(element);
```
philliphoff, line 71: If the returned tree item already has an `id`, should we abide by it (rather than override it)?

src/tree/v2/ResourceTreeDataProviderBase.ts
```diff
@@ -65,7 +66,10 @@ export abstract class ResourceTreeDataProviderBase extends vscode.Disposable imp
 
     async getTreeItem(element: ResourceGroupsItem): Promise<vscode.TreeItem> {
         const treeItem = await element.getTreeItem();
-        treeItem.id = this.itemCache.getId(element);
+
+        if (!(element instanceof BranchDataProviderItem)) {
```
philliphoff, line 70: Is there a way to avoid special casing not-`BranchDataProviderItem`?  Perhaps by ensuring that all such cases return a tree item with `id` set, so that the case reduces to just setting `id` if not already set?

src/tree/v2/CachedBranchDataProvider.ts
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            currentItem = nextItem;
+        }
+
+        return path.reverse();
+    }
+
+    private updateItemChildren(parent: ResourceModelBase, children: ResourceModelBase[]): CachedResourceModelBase[] {
+        (children as CachedResourceModelBase[]).forEach(child => child.parent = parent);
```
philliphoff, line 77: This caching alters the model's returned by the original BDP (i.e. adding a `parent`) property. Do we expect users of this caching layer to use that `parent` property separately from RG's purpose of generating paths, IDs, and parents for find/reveal?  If not (which I feel like will be the case), then there's a chance of collision with that BDP as it's not going to expect models to be modified.  In addition, I don't think we restrict models from appearing twice in a tree.  In that case, it might be better for this caching layer to return its own wrapped model (i.e. with the original model and `parent` properties as siblings).  The quirk about that approach is that such types would need to be unwrapped *twice*.  (We could have `BrandDataProviderItem.unwrap()` repeatedly call `unwrap()` until no more `unwrap()` functions exist on the inner model.

Another approach would be to store the child -> parent mapping in a separate type (but then requires more set/clear management).

src/tree/v2/CachedBranchDataProvider.ts
```diff
@@ -0,0 +1,80 @@
...
+        const treeItem = await this.branchDataProvider.getTreeItem(element);
+        treeItem.id = this.getId(element);
+        return treeItem;
+    }
+
+    resolveTreeItem(item: TreeItem, element: ResourceModelBase, token: CancellationToken): ProviderResult<TreeItem> {
+        return this.branchDataProvider.resolveTreeItem?.(item, element, token);
+    }
+
+    getId(element: CachedResourceModelBase): string {
```
philliphoff, line 54: Nit: `private`?

src/tree/v2/BranchDataProviderItem.ts
```diff
@@ -72,6 +72,19 @@ export class BranchDataProviderItem implements ResourceGroupsItem, WrappedResour
         };
     }
 
+    async getParent(): Promise<ResourceGroupsItem | undefined> {
+        if (this.branchDataProvider.getParent) {
+            const branchItem = await this.branchDataProvider.getParent(this);
```
philliphoff, line 77: I'm not sure I follow here; the BDP is expecting one of its own model types, but we're passing it a `BranchDataProviderItem` which is a RG type?

philliphoff (commented):
I like the overall approach; added a few comments.

src/tree/v2/BranchDataProviderItem.ts
alexweininger: I think this was a mistake. I changed it so that we're passing the branch item.

src/tree/v2/CachedBranchDataProvider.ts
alexweininger: If we change the property name to something more obscure like `_cachedParent`, and make it clear in the docs that using `CachedBranchDataProvider` modifies the models by adding this property, then I think it's ok. Especially considering the main reason to use `CachedBranchDataProvider` is to implement `getParent`, and it's optional.

philliphoff: I think that's probably a reasonable compromise (and can always be changed if/when it ever becomes an issue).  I might pick a name that's more "namespaced", though, something like `_rgApiCachedParent`.

alexweininger: Ok, I changed the name to `_rgApiCachedParent`.

src/tree/v2/CachedBranchDataProvider.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    resolveTreeItem(item: TreeItem, element: ResourceModelBase, token: CancellationToken): ProviderResult<TreeItem> {
+        return this.branchDataProvider.resolveTreeItem?.(item, element, token);
+    }
+
+    private getId(element: CachedResourceModelBase): string {
+        return this.getPathForItem(element).join('/');
+    }
+
+    private getPathForItem(item: CachedResourceModelBase): string[] {
```
nturinski, line 58: The name for this is a bit weird to me especially since the return type is a `string[]`.

Maybe `getAncestorsForItem` or something?

src/tree/v2/CachedBranchDataProvider.ts
```diff
@@ -0,0 +1,80 @@
...
+                path.push(currentItem.id);
+            }
+
+            currentItem = nextItem;
+        }
+
+        return path.reverse();
+    }
+
+    private updateItemChildren(parent: ResourceModelBase, children: ResourceModelBase[]): CachedResourceModelBase[] {
```
nturinski, line 76: Should children be a parameter in this case? I'd assume you'd just pass the parent and it'd update its own children. It seems weird to me that you have to pass in the children since you could pass in mismatching parent/children combinations.

Is the use case for specifying which children of the parent need updating instead of blindly updating all of them?

src/tree/v2/ResourceGroupsItemCache.ts
```diff
@@ -119,6 +120,11 @@ export class ResourceGroupsItemCache {
 
     updateItemChildren(item: ResourceGroupsItem, children: ResourceGroupsItem[]): InternalResourceGroupsItem[] {
         this.itemToChildrenCache.set(item, children);
+
+        if (item instanceof BranchDataProviderItem) {
```
nturinski, line 124: Could I get some context/explanation on what this is doing?

src/tree/v2/ResourceTreeDataProviderBase.ts
```diff
@@ -95,7 +93,7 @@ export abstract class ResourceTreeDataProviderBase extends vscode.Disposable imp
     }
 
     getParent(element: ResourceGroupsItem): vscode.ProviderResult<ResourceGroupsItem> {
-        return this.itemCache.getParentForItem(element);
+        return element.getParent?.();
```
nturinski, line 96: I have never seen this and am not sure how to google what it is 😅

alexweininger: `maybeDefinedMethod?.()` calls `maybeDefinedMethod` if it's defined. Similar to conditionally accessing a maybe defined property with `maybeDefinedProperty?.anotherProperty`.

src/tree/v2/BranchDataProviderItem.ts
```diff
@@ -79,6 +79,19 @@ export class BranchDataProviderItem implements ResourceGroupsItem, WrappedResour
...
     }
 
+    async getParent(): Promise<ResourceGroupsItem | undefined> {
+        if (this.branchDataProvider.getParent) {
+            const branchItem = await this.branchDataProvider.getParent(this.branchItem);
+            if (branchItem) {
+                return this.itemCache.getItemForBranchItem(branchItem);
+            }
+            // if this is an application resource, only RGs knows the parent
+            return this.itemCache.getParentForItem(this);
```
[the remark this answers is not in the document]
alexweininger: Should be a `GroupingItem`.

alexweininger:
I wasn't super happy with how these changes had to be implemented due to how we were handling caching in RGs. So in #434 I simplified the cache, and I've re-implemented the changes (the changes are a lot smaller now) in #435.

PR #427, opened 2022-11-10
Expose and implement "View Properties" for application resources

alexweininger:
Followed how Open in Portal was implemented in #419

The Resource Groups extension now exposes a "View Properties" command for all application resources. BranchDataProvider's have the option to override the properties. The command is also exposed if any child of an application resource sets the optional `ApplicationResourceModel.viewProperties` property.

Command for application resources without a registered BDP:
image

Command for application resources with a registered BDP:
image

Command for arbitrary children that expose the viewProperties property:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -350,7 +350,12 @@
                 },
                 {
                     "command": "azureResourceGroups.viewProperties",
-                    "when": "view == azureResourceGroups && viewItem =~ /azureResource(?!Type)/",
+                    "when": "view == azureResourceGroups && viewItem =~ /hasProperties/",
+                    "group": "9@1"
+                },
+                {
+                    "command": "azureResourceGroups.viewProperties",
+                    "when": "view == azureResourceGroups && viewItem =~ /azureResourceGroup/",
                     "group": "9@1"
                 },
                 {
--- a/src/api/v2/v2AzureResourcesApi.ts
+++ b/src/api/v2/v2AzureResourcesApi.ts
@@ -188,6 +188,23 @@ export interface ApplicationResource extends ResourceBase {
     readonly tags?: {
         [propertyName: string]: string;
     };
+
+    /**
+     * A copy of the raw resource.
+     */
+    readonly raw: {};
+}
+
+export interface ViewPropertiesModel {
+    /**
+     * File name displayed in VS Code.
+     */
+    label: string;
+
+    /**
+     * Raw data associated with the resource to populate the properties file.
+     */
+    data: {};
 }
 
 /**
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -350,7 +350,12 @@
                 },
                 {
                     "command": "azureResourceGroups.viewProperties",
-                    "when": "view == azureResourceGroups && viewItem =~ /azureResource(?!Type)/",
+                    "when": "view == azureResourceGroups && viewItem =~ /hasProperties/",
+                    "group": "9@1"
+                },
+                {
+                    "command": "azureResourceGroups.viewProperties",
+                    "when": "view == azureResourceGroups && viewItem =~ /azureResourceGroup/",
```
alexweininger, line 358: View Properties is still broken for resource groups, will fix in a future PR

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -188,6 +188,23 @@ export interface ApplicationResource extends ResourceBase {
...
+export interface ViewPropertiesModel {
+    /**
+     * File name displayed in VS Code.
+     */
+    label: string;
+
+    /**
+     * Raw data associated with the resource to populate the properties file.
+     */
+    data: {};
```
bwateratmsft, line 207: Maybe `Record<string, unknown>` or maybe even `unknown`?

src/api/v2/v2AzureResourcesApi.ts
```diff
@@ -188,6 +188,23 @@ export interface ApplicationResource extends ResourceBase {
     readonly tags?: {
         [propertyName: string]: string;
     };
+
+    /**
+     * A copy of the raw resource.
+     */
+    readonly raw: {};
```
bwateratmsft, line 195: What's the typing on the resources returned from Azure? `GenericItem` or something like that?

src/api/v2/v2AzureResourcesApi.ts
alexweininger: I do prefer `Record<string, unknown>`, but in this case it `GenericResource` doesn't satisfy `Record<string, unknown>`. `{}` is probably the next best thing, representing a non-undefined object.

src/api/v2/v2AzureResourcesApi.ts
alexweininger: It's `GenericResource`, but we also use this for a resource group, which we doesn't satisfy `GenericResource`.

src/api/v2/v2AzureResourcesApi.ts
alexweininger: `unknown` works fine, but I have to cast it to `{}` to pass it to `openReadonlyJson`. 

image

bwateratmsft: Well that's annoying. Let's just stick with `{}` then.

bwateratmsft (approved):

PR #434, opened 2022-11-28
Don't cache so many things

alexweininger:
Purpose of these changes it to reduce the responsibilities of the cache in resource groups down to one. In doing so, it drastically minimizes the amount of changes needed for #426.

Currently:
* Named `ResourceGroupsItemCache`
* Caches the following relationships:
	* `parent item -> child item`
	* `item -> child items`
	* `item -> branch item`
	* `branch item -> item`
* Generated tree item ids

With these changes:
* Renamed to `BranchDataItemCache`
* Stores `item -> branch item`, and `branch item -> item` relationships. That's it.

I accomplished this by altering the internal Resource Groups items, (`SubscriptionItem`, `GroupingItem`, and `ApplicationResourceItem`) to generate their ids on their own. 

And now `GroupingItem` and `ApplicationResourceItem` track their (optional) parent on their own, instead of secretly attaching a parent property.


src/tree/v2/BranchDataItemCache.ts
```diff
@@ -0,0 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ResourceGroupsItem } from './ResourceGroupsItem';
+
+export class BranchDataItemCache {
+    private readonly branchItemToItemCache: Map<unknown, ResourceGroupsItem> = new Map();
+    private readonly itemToBranchItemCache: Map<ResourceGroupsItem, unknown> = new Map();
```
bwateratmsft, line 10: Where does the `itemToBranchItemCache` get used?

src/tree/v2/BranchDataItemCache.ts
```diff
@@ -0,0 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ResourceGroupsItem } from './ResourceGroupsItem';
+
+export class BranchDataItemCache {
+    private readonly branchItemToItemCache: Map<unknown, ResourceGroupsItem> = new Map();
```
bwateratmsft, line 9: Not critical, but should we change their names to `branchItemToResourceItemCache` / `resourceItemToBranchItemCache`? Or something along those lines?

bwateratmsft (approved):
Love it. The best code is deleted code :grin:

src/tree/v2/BranchDataItemCache.ts
alexweininger: Figured we might as well go all the way and name it `branchItemToResourceGroupsItemCache` 😄

bwateratmsft (approved):

src/tree/v2/BranchDataItemCache.ts
bwateratmsft: Sounds good. We can always add it back if we find the need to.

PR #435, opened 2022-11-28
Add `getParent` method to `BranchDataProvider` interface

alexweininger:
#434 is meant to be reviewed and merged before this PR. I'm replacing #426 with #434 + this PR.

I wasn't super happy with how #426 had to be implemented due to how we were handling caching in RGs. So in #434 I simplified the cache, and this PR is re-implementing the changes from #426 on top of the newer simpler changes done in #434.

---

The description from #426 still applies, except I've now removed the `CachedBranchDataProvider` class since it's not needed anymore.

> Previously, the Resource Groups Tree Data Provider was caching child -> parent relationships for all items so it could do two things:
> 1. generate a path based id for an item and attach it to the `vscode.TreeItem` in `getTreeItem`.
> 2. implement `getParent`
> 
> This change breaks up this responsibility so that whatever `___DataProvider` owns the item is also responsible for implementing `getParent` and generating TreeItem ids. Meaning the RGs TDP is responsible for caching the base items subscription items, grouping items, and application resource items). And any items beneath application resources in the tree will need to be cached (or not) by their respective `BranchDataProvider`.
> 
> ~To accomplish this, I've written a `CachedBranchDataProvider` class, which wraps `BranchDataProvider` and implements `getParent` and sets `TreeItem.id` by caching `getChildren` results.~
> 
> In the future, client extensions can share an implementation of a cache that allows for similar features that `AzExtTreeDataProvider` had.


src/tree/v2/application/ApplicationResourceItem.ts
```diff
@@ -7,28 +7,34 @@ import { TreeItem } from 'vscode';
...
 import { BranchDataItemOptions, BranchDataProviderItem } from '../BranchDataProviderItem';
+import { ResourceGroupsItem } from '../ResourceGroupsItem';
 
 export class ApplicationResourceItem<T extends ResourceBase> extends BranchDataProviderItem {
     constructor(
         public readonly resource: T,
         branchItem: ResourceModelBase,
         branchDataProvider: BranchDataProvider<ResourceBase, ResourceModelBase>,
         itemCache: BranchDataItemCache,
+        private readonly parent?: ResourceGroupsItem,
```
bwateratmsft, line 18: Should it even be optional?

bwateratmsft (approved):

src/tree/v2/application/ApplicationResourceItem.ts
nturinski: Is the root considered the parent for top level tree nodes, like subscriptions? 

If it isn't, I could see parent being optional for whenever we do something like pinning/favorites.

PR #441, opened 2022-11-30
Rename method to `clear`

alexweininger:
Nit


nturinski (approved):
Will there be an implementation to clear only one resource from cache? if so, the only thing I might suggest is renaming it to `clearAll`.

alexweininger:
> Will there be an implementation to clear only one resource from cache? if so, the only thing I might suggest is renaming it to `clearAll`.

Lol I had it as `clearAll` then changed it to clear. I felt it was redundant. If we do add the ability to clear single items, then we can change the name.

PR #442, opened 2022-12-01
Add `CompatibleAzExtTreeDataProvider`

alexweininger:
I found that I separate this class from the rest of the compatibility code easily. Figured it's worth it to reduce the size of #358. I will probably keep trying to apply the changes in #358 in separate smaller PRs like this.

Merging this PR should reduce #358 to a more reasonable size. 

Brandon and I worked on this code together, so it'd be good to get a final review from another source.

### DIFF
--- a/src/api/v2/compatibility/CompatibleAzExtTreeDataProvider.ts
+++ b/src/api/v2/compatibility/CompatibleAzExtTreeDataProvider.ts
@@ -1,0 +1,100 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzExtParentTreeItem, AzExtTreeDataProvider, AzExtTreeItem, IActionContext, isWrapper, ITreeItemPickerContext } from "@microsoft/vscode-azext-utils";
+import { Disposable, Event, TreeItem, TreeView } from "vscode";
+import { ResourceGroupsItem } from "../../../tree/v2/ResourceGroupsItem";
+import { ResourceTreeDataProviderBase } from "../../../tree/v2/ResourceTreeDataProviderBase";
+
+/**
+ * An intermediate class that exists just to redeclare several events as abstract, so they
+ * can be re-redeclared as a accessors in {@link CompatibleAzExtTreeDataProvider} below
+ */
+abstract class IntermediateCompatibleAzExtTreeDataProvider extends AzExtTreeDataProvider {
+    public abstract onDidChangeTreeData: Event<AzExtTreeItem | undefined>;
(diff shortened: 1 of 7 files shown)


src/tree/v2/azure/registerAzureTree.ts
```diff
@@ -20,7 +20,11 @@ interface RegisterAzureTreeOptions {
     refreshEvent: vscode.Event<void>,
 }
 
-export function registerAzureTree(context: vscode.ExtensionContext, options: RegisterAzureTreeOptions): void {
+interface RegisterAzureTreeResult {
+    azureResourceTreeDataProvider: AzureResourceTreeDataProvider;
+}
+
+export function registerAzureTree(context: vscode.ExtensionContext, options: RegisterAzureTreeOptions): RegisterAzureTreeResult {
```
philliphoff, line 27: Nit: as the type is not used by callers (e.g. it's not exported) and unlikely to require returning other properties, I might dispense with the interface and just return the TDP directly.  I think it's a reasonable thing for callers to expect to be returned from such a function.  (And same comment below.)

src/api/v2/compatibility/CompatibleAzExtTreeDataProvider.ts
```diff
@@ -0,0 +1,100 @@
...
+        // Trigger a refresh at the given treeItem
+        this.tdp.notifyTreeDataChanged(treeItem as unknown as ResourceGroupsItem);
+
+        return Promise.resolve();
+    }
+
+    public override refreshUIOnly(treeItem: AzExtTreeItem | undefined): void {
+
+        this.tdp.notifyTreeDataChanged(treeItem as unknown as ResourceGroupsItem);
+
```
philliphoff, line 69: Nit: extra line.

philliphoff (approved):
Just a couple of minor comments.

src/tree/v2/azure/registerAzureTree.ts
bwateratmsft: Agreed, it felt a little strange to me.

src/extension.ts
```diff
@@ -149,12 +110,15 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
                 workspaceResourceProviderManager,
-                workspaceResourceBranchDataProviderManager);
+                workspaceResourceBranchDataProviderManager
+            );
         }
 
         return v2Api;
     }
 
+    v2ApiFactory();
```
bwateratmsft, line 120: It feels like this method should have a return.

bwateratmsft (approved):
My opinion kinda doesn't count but :thumbsup:

src/extension.ts
alexweininger: Technically this isn't used yet. And later it will be `ext.v2.api = vsApiFactory();`. I'll remove it for now and add that stuff later when it's actually used.

PR #451, opened 2022-12-13
Don't provide duplicate resources

alexweininger:
In #449 we made it so `getResources` is called for each open workspace folder, and once with undefined as an argument for system-level resources.

For compatibility, we'll only return resources when undefined is passed.


bwateratmsft (approved):

src/api/v2/compatibility/workspace/CompatibilityWorkspaceResourceProvider.ts
```diff
@@ -15,7 +15,10 @@ export class CompatibilityWorkspaceResourceProvider implements V2WorkspaceResour
     // No comparable mechanism in v1, leave as undefined
     onDidChangeResource?: Event<WorkspaceResource | undefined> = undefined;
 
-    public async getResources(source: WorkspaceFolder): Promise<WorkspaceResource[]> {
+    public async getResources(source: WorkspaceFolder | undefined): Promise<WorkspaceResource[]> {
+        if (source) {
```
bwateratmsft, line 19: Maybe a comment to explain what's happening?

PR #453, opened 2022-12-14
Use factory pattern for exposed API

alexweininger:


src/extension.ts
```diff
@@ -100,53 +98,51 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
-                    workspaceResourceProviderManager,
-                    workspaceResourceBranchDataProviderManager,
-                    workspaceResourceTreeDataProvider,
-                ),
-                activity: {
-                    registerActivity
+    const v2ApiFactory: AzureExtensionApiFactory<AzureResourcesApiInternal> = {
+        apiVersion: '2.0.0',
+        createApi: (options: GetApiOptions) => {
+            return createWrappedAzureResourcesExtensionApi(
```
bwateratmsft, line 104: Is this pattern going to create a new API object for each extension--thus, separate lists of registrations, etc.?

alexweininger: No, it just wraps the same API instance with new telemetry handling each time.

bwateratmsft (approved):

PR #456, opened 2022-12-15
Implement `pickAppResource` with v1 compatibility

alexweininger:


src/api/pickAppResource.ts
```diff
@@ -3,18 +3,39 @@
...
+    return filterMap(filter, getAzExtResourceType);
+}
+
+function filterMap<T, TMapped>(source: T[], predicateMapper: (item: T, index: number) => TMapped | null | undefined): TMapped[] {
+    let index = 0;
+    return source.reduce<TMapped[]>((accumulator, current) => {
+        const mapped = predicateMapper(current, index++);
+        // handles null or undefined
+        // eslint-disable-next-line eqeqeq
+        if (mapped != null) {
```
bwateratmsft, line 36: Does this work instead and remove the need for the ESLint suppression? Not sure what other falsy values might be problems or not.
```suggestion
        if (!!mapped) {
```

Otherwise I'd almost prefer
```suggestion
        if (mapped !== null && mapped !== undefined) {
```

alexweininger: I like doing `mapped !== null && mapped !== undefined`.

src/api/pickAppResource.ts
```diff
@@ -3,18 +3,39 @@
...
+
+    filter = Array.isArray(filter) ? filter : [filter];
+    return filterMap(filter, getAzExtResourceType);
+}
+
+function filterMap<T, TMapped>(source: T[], predicateMapper: (item: T, index: number) => TMapped | null | undefined): TMapped[] {
+    let index = 0;
+    return source.reduce<TMapped[]>((accumulator, current) => {
+        const mapped = predicateMapper(current, index++);
+        // handles null or undefined
```
bwateratmsft, line 34: These two comment lines can be removed

bwateratmsft (approved):

PR #458, opened 2022-12-15
Update contribution keys

alexweininger:
Use `x-azResources` key instead of `x-azResourcesV2` to look for contributions. We don't need to distinguish between v1 and v2 here, and I'll update the client extensions to reflect this change.

Also changing the `application` part to `azure`.


alexweininger:
> Will we be able to (or will we even need to) distinguish between V1 and V2 client extensions?

I don't think we need to. With that said, we won't be able to based on the package.json alone after this change.

bwateratmsft (approved):
Looks fine. The old entries under `x-azResources` were `commands` and `activation`, this won't contradict.

PR #460, opened 2022-12-20
Refresh individual tree items, and fix refresh title button

alexweininger:
Two changes:

1. Allow refreshing a single tree item instead of the entire tree everywhere. This means changing the `Event<void>` to `Event<void | ResourceGroupsItem | ResourceGroupsItem[] | null | undefined>` everywhere.
2. Split the Azure view refresh command in two. One that refreshes the entire tree, ignoring any selected nodes. This makes the refresh button on the Azure view title behave like it has in the past. The new command is `refreshAzureItem`, intended for the context menus of items in the Azure view.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -31,6 +31,7 @@ import { HelpTreeItem } from './tree/HelpTreeItem';
 import { AzureResourceBranchDataProviderManager } from './tree/v2/azure/AzureResourceBranchDataProviderManager';
 import { DefaultAzureResourceBranchDataProvider } from './tree/v2/azure/DefaultAzureResourceBranchDataProvider';
 import { registerAzureTree } from './tree/v2/azure/registerAzureTree';
+import { ResourceGroupsItem } from './tree/v2/ResourceGroupsItem';
 import { registerWorkspaceTree } from './tree/v2/workspace/registerWorkspaceTree';
 import { WorkspaceDefaultBranchDataProvider } from './tree/v2/workspace/WorkspaceDefaultBranchDataProvider';
 import { WorkspaceResourceBranchDataProviderManager } from './tree/v2/workspace/WorkspaceResourceBranchDataProviderManager';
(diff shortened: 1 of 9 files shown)


src/tree/v2/ResourceTreeDataProviderBase.ts
```diff
@@ -82,7 +82,7 @@ export abstract class ResourceTreeDataProviderBase extends vscode.Disposable imp
         return await this.onGetChildren(element);
     }
 
-    getParent(element: ResourceGroupsItem): vscode.ProviderResult<ResourceGroupsItem> {
+    getParent(element: ResourceGroupsItem): vscode.ProviderResult<ResourceGroupsItem | null | undefined> {
```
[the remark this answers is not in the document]
alexweininger: Nope. Fixed.

src/extension.ts
```diff
@@ -44,13 +45,13 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
     registerUIExtensionVariables(ext);
     registerAzureUtilsExtensionVariables(ext);
 
-    const refreshAzureTreeEmitter = new vscode.EventEmitter<void>();
+    const refreshAzureTreeEmitter = new vscode.EventEmitter<void | ResourceGroupsItem | ResourceGroupsItem[] | null | undefined>();
```
nturinski, line 48: I didn't see any changes that would accommodate all these new data types. Do we have a function to handle all of these different types? I haven't been able to find it anywhere.

alexweininger: https://github.com/microsoft/vscode-azureresourcegroups/pull/460/files#diff-5ca32a8fc22d6ae9e10d28b887089aca193084aaf7b5b9dc3d5089b650741fe9L58

nturinski: Yeah, but even if it fires the data, doesn't the event handler need to actually handle all the types? Does the tools package already handle that case?

src/extensionVariables.ts
```diff
@@ -23,8 +24,8 @@ namespace extEvents {
 }
 
 export namespace extActions {
-    export let refreshWorkspaceTree: () => void;
-    export let refreshAzureTree: () => void;
+    export let refreshWorkspaceTree: (data?: ResourceGroupsItem | ResourceGroupsItem[] | null | undefined | void) => void;
```
nturinski, line 27: Do we need `null | undefined | void`? It seems a little excessive. The parameter is also already optional as it is.

src/extension.ts
alexweininger: VS Code listens to the `onDidChangeTreeData` event on the tree data provider.

src/extensionVariables.ts
alexweininger: I'm keeping consistent with the `onDidChangeTreeData` event data type.

```typescript
onDidChangeTreeData?: Event<T | T[] | undefined | null | void>;
```

nturinski (approved):

PR #461, opened 2022-12-20
v2 edit tags command for resources

alexweininger:
Depends on https://github.com/microsoft/vscode-azuretools/pull/1257

I will add support for editing tags on resource groups in a follow up PR.

Also, I need to add a `dontUnwrap` flag to `appResourceExperience` in utils so to support getting an item from the picker here.

### DIFF
--- a/src/commands/tags/TagFileSystem.ts
+++ b/src/commands/tags/TagFileSystem.ts
@@ -76,7 +76,7 @@ export class TagFileSystem extends AzExtTreeFileSystem<ITagsModel> {
             context.telemetry.measurements.tagDiagnosticsLength = diagnostics.length;
 
             const showErrors: MessageItem = { title: localize('showErrors', 'Show Errors') };
-            const message: string = localize('errorsExist', 'Failed to upload tags for {0}.', this.getDisplayName(model));
+            const message: string = localize('errorsExist', 'Failed to upload tags for {0}.', this.getDetailedName(model));
             void window.showErrorMessage(message, showErrors).then(async (result) => {
                 if (result === showErrors) {
                     const openedUri: Uri | undefined = window.activeTextEditor?.document.uri;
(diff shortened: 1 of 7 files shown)


src/commands/tags/TagFileSystem.ts
```diff
@@ -105,10 +132,11 @@ export class TagFileSystem extends AzExtTreeFileSystem<ResourceGroupTreeItem | A
...
-        if (node instanceof ResourceGroupTreeItem) {
-            return (await node.getData())?.tags;
-        }
-        return node.data.tags;
+    private async getTagsFromNode(node: ITagsModel): Promise<Tags | undefined> {
+        return await node.getTags();
+    }
+
+    private getDisplayName(node: ITagsModel): string {
+        return `${node.displayType} "${node.displayName}"`;
```
nturinski, line 140: A little bit confusing that there is a displayName property that is different from getDisplayName. Maybe something like getDetailedName or getFullName? getDisplayTypeName?

src/tree/AppResourceTreeItem.ts
```diff
@@ -46,7 +45,7 @@ export class AppResourceTreeItem extends ResolvableTreeItemBase implements Group
         this.groupConfig = createGroupConfigFromResource(resource, root.id);
 
         this.contextValues.add(AppResourceTreeItem.contextValue);
-        ext.tagFS.fireSoon({ type: FileChangeType.Changed, item: this });
+        // ext.tagFS.fireSoon({ type: FileChangeType.Changed, item: this });
```
nturinski, line 48: Could you explain why you commented this out? Should it go back eventually?

alexweininger: This is a v1 class that isn't used anymore. It's a lot of work to delete everything, will do in a follow up PR.

alexweininger: When I start deleting files it just shows how much there is left to do. For example:
1. Toggle show all resources command
2. Group by tags needs updated so that it doesn't rely on the root tree item to get subscriptions
3. `revealTreeItem`

I'm sure there's more

nturinski (approved):

PR #474, opened 2022-12-27
Don't refresh entire tree on add/remove BDP

alexweininger:
Before this change:
When a BDP was registered we refreshed the entire tree. A case where this becomes super annoying is if you have expanded the Azure Functions group but then expand another group, say App Service. The App Service extension will activate and register a BDP. The whole tree is then refreshed, including the Function Apps.

After this change:
Group items listen for changes to the registered BDPs. They will refresh when the BDP registered for a type (`AzExtResourceType`) they care about changes. The types they care about are the set of types of the group children.

For the Workspace tree, the refresh behavior remains the same. On any changes to registered BDPs, the whole tree will refresh. This is still needed for now since the client-controlled tree items are root tree items.

### DIFF
--- a/src/tree/v2/azure/GroupingItem.ts
+++ b/src/tree/v2/azure/GroupingItem.ts
@@ -30,7 +30,7 @@ export class GroupingItem implements ResourceGroupsItem {
         public readonly context: ResourceGroupsTreeContext,
         private readonly resourceItemFactory: ResourceItemFactory<AzureResource>,
         private readonly branchDataProviderFactory: (azureResource: AzureResource) => AzureResourceBranchDataProvider<AzureResourceModel>,
-        private readonly onChangeBranchDataProviders: vscode.Event<AzExtResourceType>,
+        private readonly onDidChangeBranchDataProviders: vscode.Event<AzExtResourceType>,
         private readonly contextValues: string[] | undefined,
         private readonly iconPath: TreeItemIconPath | undefined,
         public readonly label: string,
@@ -49,7 +49,7 @@ export class GroupingItem implements ResourceGroupsItem {
     async getChildren(): Promise<ResourceGroupsItem[] | undefined> {
         const sortedResources = this.resources.sort((a, b) => a.name.localeCompare(b.name));
 
-        this.onChangeBranchDataProviders((type: AzExtResourceType) => {
+        this.onDidChangeBranchDataProviders((type: AzExtResourceType) => {
             const azExtResourceTypes = sortedResources.map(r => r.resourceType);
             if (azExtResourceTypes.includes(type)) {
                 ext.actions.refreshAzureTree(this);
--- a/src/tree/v2/azure/GroupingItem.ts
+++ b/src/tree/v2/azure/GroupingItem.ts
@@ -50,7 +50,7 @@ export class GroupingItem implements ResourceGroupsItem {
         const sortedResources = this.resources.sort((a, b) => a.name.localeCompare(b.name));
 
         this.onChangeBranchDataProviders((type: AzExtResourceType) => {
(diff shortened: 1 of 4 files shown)


src/tree/v2/azure/GroupingItem.ts
```diff
@@ -46,6 +48,14 @@ export class GroupingItem implements ResourceGroupsItem {
 
     async getChildren(): Promise<ResourceGroupsItem[] | undefined> {
         const sortedResources = this.resources.sort((a, b) => a.name.localeCompare(b.name));
+
+        this.onChangeBranchDataProviders((type: AzExtResourceType) => {
```
nturinski, line 52: Does it make more sense to do this on the root of the tree data provider? It could look for which branch data providers need to be refreshed once.

Like this, every branch data providers will have to do this map and refresh individually. I doubt the performance is really that bad, but it does feel a little wasteful.

alexweininger: I initially tried to do it the way you describe. Basically expose a method like 
```typescript
getOnChangeBranchDataProviderForType(type: AzExtResourceType): Event<void>
```

which is called and listened to like

```typescript
getOnChangeBranchDataProviderForType(type)(() => {
	// do something
})
```

First of all, that second snippet is ugly IMO. And I couldn't find a nice way to avoid it. And more importantly, we'd have to keep a `Map<AzExtResourceType, Listener>`. Otherwise we would be just moving where the check happens, and not reducing the number of checks. I didn't think it was worth it considering this should have no real performance impact.

src/tree/v2/azure/GroupingItem.ts
```diff
@@ -46,6 +48,14 @@ export class GroupingItem implements ResourceGroupsItem {
...
+            const azExtResourceTypes: AzExtResourceType[] = sortedResources.map(r => r.resourceType) as AzExtResourceType[];
```
bwateratmsft, line 53: Why was this `as` cast necessary?

src/tree/v2/azure/GroupingItem.ts
```diff
@@ -29,6 +30,7 @@ export class GroupingItem implements ResourceGroupsItem {
         public readonly context: ResourceGroupsTreeContext,
         private readonly resourceItemFactory: ResourceItemFactory<AzureResource>,
         private readonly branchDataProviderFactory: (azureResource: AzureResource) => AzureResourceBranchDataProvider<AzureResourceModel>,
+        private readonly onChangeBranchDataProviders: vscode.Event<AzExtResourceType>,
```
bwateratmsft, line 33: Can we change the name to match VSCode's naming conventions for events?
```suggestion
        private readonly onDidChangeBranchDataProviders: vscode.Event<AzExtResourceType>,
```

src/tree/v2/azure/GroupingItem.ts
alexweininger: I removed the cast, but now the type is `(AzExtResourceType | undefined)[]`, but I think that's fine.

src/tree/v2/azure/GroupingItem.ts
nturinski: Then would it make sense for the branch data providers to keep track what what types they own? This actually might be necessary. 

For example, if the Functions branch data provider doesn't have any resources yet and a refresh is fired, it wouldn't know to refresh itself because it wouldn't find any of that "type" in its `sortedResources`, right?

alexweininger: Branch data providers aren't responsible for refreshing the list of resources. The list of resources is handled by the subscription items.

nturinski (approved):

src/tree/v2/azure/GroupingItem.ts
nturinski: Discussed offline.

Issue #479, opened 2023-01-04
Implement reveal for v2


alexweininger:

bwateratmsft:
Several of the commands planned for the Dev CLI extension will require reveal capability.

alexweininger:
Finished in #481, #482, and #483!

PR #481, opened 2023-01-04
Implement `revealResource` for v2

alexweininger:
Note that this currently doesn't support revealing subscriptions or resource groups, only resources. I hope to be able to modify this function to support them in a future PR.

Working towards https://github.com/microsoft/vscode-azureresourcegroups/issues/479

PR #482, opened 2023-01-04
Add compatibility shim for TreeView.reveal

alexweininger:
These changes implement a compatibility shim for `TreeView.reveal`. This will only be used by the v1.5 client extensions that use the tree views exposed by the v1 host API to reveal tree items.

For now, we are not exposing the tree views or a way to reveal tree items for v2 client extensions.

Since v1.5 client extensions are using the exposed `vscode.TreeView.reveal` methods directly. To implement a compatibility shim, I'm directly modifying the `TreeView.reveal` method. I am replacing it with a method that wraps the original `TreeView.reveal`, but adds some necessary compatibility logic.

The necessary compatibility logic is converting the branch item into a resource groups item. To do this we look in the cache, if it's not there, then we use `findItemById` to locate the item. Calling `findItemById` also makes sure that the parents of this item are loaded so that reveal works.

I bundled up all the weird stuff in an easy to use `createTreeView` function that we now use for creating both the Azure and Workspace tree views.

Working towards https://github.com/microsoft/vscode-azureresourcegroups/issues/479

### DIFF
--- a/src/commands/revealResource.ts
+++ b/src/commands/revealResource.ts
@@ -4,11 +4,11 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { parseAzureResourceId } from '@microsoft/vscode-azext-azureutils';
-import { IActionContext, parseError } from '@microsoft/vscode-azext-utils';
+import { AzExtTreeItem, IActionContext, parseError } from '@microsoft/vscode-azext-utils';
 import { AppResource } from '@microsoft/vscode-azext-utils/hostapi';
 import { ext } from '../extensionVariables';
-import { AppResourceTreeItem } from '../tree/AppResourceTreeItem';
-import { SubscriptionTreeItem } from '../tree/SubscriptionTreeItem';
+import { ResourceGroupsItem } from '../tree/v2/ResourceGroupsItem';
+import { ResourceTreeDataProviderBase } from '../tree/v2/ResourceTreeDataProviderBase';
 
 export async function revealResource(context: IActionContext, resourceId: string): Promise<void>;
 export async function revealResource(context: IActionContext, resource: AppResource): Promise<void>;
@@ -18,12 +18,9 @@ export async function revealResource(context: IActionContext, arg: AppResource |
     context.telemetry.properties.resourceType = parseAzureResourceId(resourceId).provider.replace(/\//g, '|');
 
     try {
-        const subscriptionNode: SubscriptionTreeItem | undefined = await ext.appResourceTree.findTreeItem(`/subscriptions/${parseAzureResourceId(resourceId).subscriptionId}`, { ...context, loadAll: true });
-        const appResourceNode: AppResourceTreeItem | undefined = await subscriptionNode?.findAppResourceByResourceId(context, resourceId);
-        if (appResourceNode) {
-            // ensure the parent node loaded this AppResourceTreeItem
-            await appResourceNode.parent?.getCachedChildren(context);
-            await ext.appResourceTreeView.reveal(appResourceNode, { select: true, focus: true, expand: false });
+        const item: ResourceGroupsItem | undefined = await (ext.v2.api.resources.azureResourceTreeDataProvider as ResourceTreeDataProviderBase).findItemById(resourceId);
+        if (item) {
+            await ext.appResourceTreeView.reveal(item as unknown as AzExtTreeItem, { expand: false, focus: true, select: true });
         }
     } catch (error) {
         context.telemetry.properties.revealError = parseError(error).message;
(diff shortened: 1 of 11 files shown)


src/commands/revealResource.ts
```diff
@@ -18,12 +18,9 @@ export async function revealResource(context: IActionContext, arg: AppResource |
...
     try {
-        const subscriptionNode: SubscriptionTreeItem | undefined = await ext.appResourceTree.findTreeItem(`/subscriptions/${parseAzureResourceId(resourceId).subscriptionId}`, { ...context, loadAll: true });
-        const appResourceNode: AppResourceTreeItem | undefined = await subscriptionNode?.findAppResourceByResourceId(context, resourceId);
-        if (appResourceNode) {
-            // ensure the parent node loaded this AppResourceTreeItem
-            await appResourceNode.parent?.getCachedChildren(context);
-            await ext.appResourceTreeView.reveal(appResourceNode, { select: true, focus: true, expand: false });
+        const item: ResourceGroupsItem | undefined = await (ext.v2.api.resources.azureResourceTreeDataProvider as ResourceTreeDataProviderBase).findItemById(resourceId);
+        if (item) {
+            await ext.appResourceTreeView.reveal(item as unknown as AzExtTreeItem, { expand: false, focus: true, select: true });
```
bwateratmsft, line 23: Same as in #481, why did this need to be cast as an `AzExtTreeItem`?

src/tree/v2/createTreeView.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    modifyReveal(treeView, options.treeDataProvider, options.itemCache);
+
+    return treeView;
+}
+
+function modifyReveal(treeView: TreeView<ResourceGroupsItem>, treeDataProvider: ResourceTreeDataProviderBase, itemCache: BranchDataItemCache): void {
+    (treeView as InternalTreeView)._reveal = treeView.reveal.bind(treeView) as typeof treeView.reveal;
+
+    treeView.reveal = async (element, options) => {
```
bwateratmsft, line 42: I thought none of the v1.5 extensions called reveal? Or am I thinking of something else? If they aren't using it then we shouldn't need to monkeypatch a solution to deal with wrapping/unwrapping/compat.

alexweininger: The v1.5 extensions do call reveal. There was discussion about removing all the calls, but we never went through with it.

src/commands/revealResource.ts
alexweininger: `ext.appResourceTreeView` is still typed as `TreeView<AzExtTreeItem>`. We need to change the type and rename it. I'll do that in a follow up PR, as I'll also have to remove some old v1 stuff that relies on it.

bwateratmsft: I thought we changed the interface to `TreeView<unknown>`

bwateratmsft: Ah yeah. Maybe add a comment for that too.

src/tree/v2/workspace/WorkspaceResourceTreeDataProvider.ts
```diff
@@ -16,14 +16,13 @@ export class WorkspaceResourceTreeDataProvider extends ResourceTreeDataProviderB
     constructor(
         private readonly branchDataProviderManager: WorkspaceResourceBranchDataProviderManager,
         onRefresh: vscode.Event<void | ResourceGroupsItem | ResourceGroupsItem[] | null | undefined>,
-        private readonly resourceProviderManager: WorkspaceResourceProviderManager) {
+        private readonly resourceProviderManager: WorkspaceResourceProviderManager,
+        itemCache: BranchDataItemCache) {
```
nturinski, line 20: Could you explain why the workspace tree view needs the `itemCache` now? I don't really have an opposing opinions on it, just wondering if it's only for reveal and if it's something that we'll remove in the future.

alexweininger: The workspace tree already had an item cache. I've just moved where it's instantiated so that I can use it for reveal as well.

alexweininger: > EDIT: Never mind, I re-read your PR description and it makes sense. But maybe make a note to remove it from the `TreeDataProvider` as well?

In the tree data provider the item cache is used for more than just reveal. It's also used for refreshing, since we need to map branch data items into resource groups items to pass to `onDidChangeTreeData`.

src/commands/revealResource.ts
alexweininger: Will do. We need to cleanup the extension variables. There are a bunch of residual variables from v1 that probably can be removed, or at least need to be renamed.

alexweininger:
I wish the "Re-request review" was more like a nudge. It's not urgent, just a nudge 😄

bwateratmsft (approved):

PR #483, opened 2023-01-05
Add revealAzureResource to v2 API

alexweininger:
Adding `revealAzureResource` to the v2 API interface in https://github.com/microsoft/vscode-azuretools/pull/1333.

`revealAzureResource` supports revealing subscriptions, resource groups, and resources. Until now, it has only supported revealing resources, and passing an id that wasn't a full resource id threw an error. Now you can pass partial Azure resource ids (that's what I'm calling them).

To allow this I had to make some changes to how we parse the id. I also added a telemetry property `resourceKind`, which can either be subscription, resource group, or resource.


src/api/v2/createAzureResourcesHostApi.ts
```diff
@@ -42,5 +43,15 @@ export function createAzureResourcesHostApi(
...
+
+        revealAzureResource: (id: string, options?: VSCodeRevealOptions) => {
+            // TODO: don't create new telemetry context here, instead use the context created from the initial wrapped API call
+            return callWithTelemetryAndErrorHandling('internalRevealResource', context => {
+                context.errorHandling.rethrow = true;
+                context.errorHandling.suppressDisplay = true;
+                context.errorHandling.suppressReportIssue = true;
+                return revealResource(context, id, options);
+            });
+        },
```
alexweininger, line 55: More info on this todo...

We wrap methods on the API with telemetry handling and I want to be able to access the telemetry context created there here. The best solution I found was async hooks. I'm open for suggestions on what to do.

src/commands/revealResource.ts
```diff
@@ -3,26 +3,56 @@
...
+    const parsedAzureResourceId = parsePartialAzureResourceId(resourceId);
+    const resourceKind = getResourceKindFromId(parsedAzureResourceId);
+    context.telemetry.properties.resourceKind = resourceKind;
+
+    if (resourceKind === 'resource') {
+        context.telemetry.properties.resourceType = parsedAzureResourceId.provider?.replace(/\//g, '|');
+    }
+}
+
+function parsePartialAzureResourceId(id: string): Partial<ParsedAzureResourceId> & Pick<ParsedAzureResourceId, 'rawId'> {
```
bwateratmsft, line 36: I thought we had utility methods for this in `vscode-azuretools`.

src/commands/revealResource.ts
```diff
@@ -3,26 +3,56 @@
...
+    return {
+        rawId: id,
+        subscriptionId: matches?.[1],
+        resourceGroup: matches?.[3],
+        provider: matches?.[5],
+        resourceName: matches?.[6]
+    }
+}
+
+function getResourceKindFromId(parsedId: Partial<ParsedAzureResourceId>): 'subscription' | 'resourceGroup' | 'resource' {
```
bwateratmsft, line 47: We should have this return an `AzExtResourceType`.

src/commands/revealResource.ts
alexweininger: The utility we have expects a full ARM ID, and throws an error if you pass a partial id.

src/commands/revealResource.ts
alexweininger: Unfortunately, we have no way of getting the `AzExtResourceType` from the ARM ID, since the `kind` property isn't present.

Should we add subscription, and a generic "resource" type to `AzExtResourceType`?

bwateratmsft: Subscription makes sense to have a type for, IMO. I don't think a generic resource though makes sense. Though, since all of this is internal/private, I don't think it matters too much.

src/api/v2/createAzureResourcesHostApi.ts
bwateratmsft: This would effectively be reusing an action context for multiple events...can't say I'm a fan.

bwateratmsft: IMO a shared correlation ID would be reasonable. Certainly easier to match up than timestamps.

alexweininger: Ok, I don't think it's necessary now. But that's an option if we need to add correlation later. I don't see us needing to know the answer to "Which client extensions are revealing which resources?".

bwateratmsft: That's fair. I think we'll already know the answer anyway.

bwateratmsft (approved):

Issue #467, opened 2022-12-26
There is no "Open in Portal" action in the context menu


v-xinda:
**OS:** Win10
**Build Version:** Azure Resources v0.5.7-alpha.0

**Repro Steps:**
1. Right click a resource group.
2. Check whether there is a "Open in Portal" action in the context menu.

**Expect:**
There is a "Open in Portal" action in the context menu.

**Actual:**
There is no "Open in Portal" action in the context menu.

alexweininger:
Fixed by #505

PR #505, opened 2023-01-10
Fix Open in Portal for subscriptions and RGs

### LINKED ISSUE #502: An error occurs after executing "Open in Portal" action
**OS:** Win10
**Build Version:** 20230106.42

**Repro Steps:**

1. Right click a subscription -> Click "Open in Portal".
2. Check whether no error occurs.

**Expect:**
No error occurs.

**Actual:**
An error occurs.
image

**More Info:**
An error occurs after executing command "Azure: Open in Portal" from command palette.
image

alexweininger:
Fixes #502, and #467

### DIFF
--- a/src/commands/openInPortal.ts
+++ b/src/commands/openInPortal.ts
@@ -3,16 +3,14 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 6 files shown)


alexweininger:
> Do we need to change the `openInPortal` command to look at when viewItem ~= /hasPortalUrl/ in the context value?

Already done.

PR #509, opened 2023-01-11
Stop setting tree item tooltip to id

alexweininger:


nturinski (approved):

nturinski:
Consider leaving it as a debugging tool similar to the context value stuff? Or maybe expose a new command, "View Debugging Info" that could show ids, context values, etc.?

Issue #487, opened 2022-12-23
The new added setting doesn't auto display in the application setting list when executing command "Azure Functions: Add New Setting..." from command palette


v-xinda:
**OS:** Win10
**Build Version:** Azure Function v1.9.1-alpha.0 + Azure Resources v0.5.7-alpha.0

**Repro Steps:**
1. F1 -> Click "Azure Functions: Add New Setting..." -> Select a resource -> Enter a name -> Enter a value.
2. Check whether the new added setting auto displays in the application setting list.

**Expect:**
The new added setting auto displays in the application setting list.

**Actual:**
The new added setting doesn't auto display in the application setting list.

**More Info:**
1. This issue doesn't reproduce when executing "Add New Setting..." from context menu of the Application Settings node.
2. This issue also reproduces when executing "Azure Functions: Delete Setting..." from command palette.
3. There is no notification "Create XXX..." in the application setting list when executing command "Azure Functions: Add New Setting..." from command palette. As below:
<img width="416" alt="Screenshot 2022-12-23 at 16 38 52" src="https://user-images.githubusercontent.com/57662131/209308568-90b8ac87-a150-4985-82e8-a505bbcda0d6.png">
4. The application setting doesn't auto refresh after executing "Rename/Edit/Toggle as Slot Setting..." command from command palette.

alexweininger:
Fixed by #496

v-xinda:
@alexweininger  This issue also reproduces on the latest build 20230106.42. 
**Repro Steps:** 
Expand the Application setting node -> Execute "Add New Setting..." from command palette.

This issue doesn't reproduce if don't expand the Application setting node to create a new setting by executing "Add New Setting..." from command palette.
**Don't Repro Steps:**
Create a new function app (Don't expand the Application setting node) -> Execute "Add New Setting..." from command palette.

v-xinda:
The new created **slot** doesn't auto display under the Slot node if expand the Slot node previously -> then create a new slot by executing "Create Slot..." from command palette.

alexweininger:
This is happening because until now, the tree item picker could return cached tree nodes if nodes had already been shown on the tree UI. Now it always returns new tree nodes that aren't the same as any tree nodes visible on the tree.
 
For v1.5 extensions a band aid fix is to try to translate the picked node into a node that we already know about before giving it back to the client. This isn’t a sustainable solution as it requires access to the branch item cache, which v2 client extensions won’t have.

[...]

PR #515, opened 2023-01-11
Don't store state on tree nodes

### LINKED ISSUE #513: There is no response after executing "Azure: Delete Resource Group..." from command palette
**OS:** Win10
**Build Version:** 20230111.7

**Repro Steps:**
1. F1 -> Click "Azure: Delete Resource Group...".
2. Check whether the resource groups are listed.

**Expect:**
The resource groups are listed.

**Actual:**
There is no response after executing "Azure: Delete Resource Group..." from command palette.

alexweininger:
This is a direction I've liked to go in for a while now. But the gist is that we shouldn't be storing state on tree nodes since they are ephemeral. This is even more important now that we aren't caching tree nodes. Instead of storing state on tree items, we can store state centrally mapped to resource ids. And use event emitters to handle updates.

One common case for why this is needed is for commands invoked from the command palette, which then select a resource using a quick pick. For some more context, see https://github.com/microsoft/vscode-azureresourcegroups/issues/487#issuecomment-1376616009. I've added a band aid fix for v1.5 extensions, however moving forward we need a better solution.

The state we store is very simple. It mostly has to do with conveying an action is currently happening to a tree item. Think `runWithTemporaryDescription` kind of stuff. Currently all of this is handled by `AzExtTreeItem`, but now it will be handled by this.

A simplified version of how it works: you can fire an update event for a resource id, tree nodes can listen to this event and refresh themselves if they represent that resource id.

Some of the other benefits to this approach are:
* Only the resource id is needed to apply state updates. This is important for when commands are called from the palette and pick a resource
* State is preserved through refreshes
* State can be applied to all tree items that represent a resource. This means that if we in the future have multiple tree items on the tree that represent the same resource, they both will reflect state changes (if we want).

In the near future when we're rewriting client extensions for v2, I plan on using the same pattern, and making shared utilities for this. For now, I'll keep it in resource groups.

Another note: the reason I'm doing this now is it's required for #513. I will follow this PR up with for #513 and it will show a bit more as to why this is needed.

### DIFF
--- a/src/tree/azure/GroupingItem.ts
+++ b/src/tree/azure/GroupingItem.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzExtResourceType, createContextValue, ISubscriptionContext, TreeItemIconPath } from '@microsoft/vscode-azext-utils';
-import { AzureResource, AzureResourceBranchDataProvider, AzureResourceModel, ViewPropertiesModel } from '@microsoft/vscode-azext-utils/hostapi.v2';
+import { AzureResource, AzureResourceBranchDataProvider, AzureResourceModel, AzureSubscription, ViewPropertiesModel } from '@microsoft/vscode-azext-utils/hostapi.v2';
 import * as vscode from 'vscode';
 import { ITagsModel, ResourceTags } from '../../commands/tags/TagFileSystem';
 import { ext } from '../../extensionVariables';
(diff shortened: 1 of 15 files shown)


src/tree/azure/GroupingItem.ts
```diff
@@ -101,7 +102,7 @@ export class GroupingItem implements ResourceGroupsItem {
         treeItem.iconPath = this.iconPath;
         treeItem.id = this.id;
 
-        return treeItem;
+        return treeItemState.applyToTreeItem({ ...treeItem, id: this.id });
```
alexweininger, line 105: An idea of mine is to find the best way to apply this to all tree items, rather than manually applying it in each `getTreeItem` method.

Possible an abstract base class for tree items, but I'm still working on it.

nturinski (approved):

PR #516, opened 2023-01-11
Fix delete resource group from command palette

### LINKED ISSUE #466: An error occurs when creating a resource group from Command Palette
**OS:** Win10
**Build Version:** Azure Resources v0.5.7-alpha.0

**Repro Steps:**
F1 -> Azure: Create Resource Group...

**Expect:**
It navigates to next step without any error.

**Actual:**
An error occurs.
image

**More Info:**
1. This issue also reproduces when creating resource group with "Azure: Create Resource..." command.
2. This issue does not reproduce when creating a resource group from context menu.

### LINKED ISSUE #513: There is no response after executing "Azure: Delete Resource Group..." from command palette

alexweininger:
Fixes #513 
Fixes #466

Since the user could have their resources grouped by something other than resource groups, we need to pick resource groups by listing them from the SDK instead of using the tree data provider.


src/commands/deleteResourceGroup/v2/deleteResourceGroupV2.ts
```diff
@@ -3,72 +3,110 @@
...
+    // unset nodes that are not resource groups
+    selectedNodes = selectedNodes?.filter(n => !!n.resourceGroup);
+    if (!primaryNode?.resourceGroup) {
+        primaryNode = undefined;
+    }
+
+    let subscription: AzureSubscription;
+    let resourceGroupsToDelete: AzureResource[] = [];
+
     if (!selectedNodes) {
```
nturinski, line 32: This is a nightmare. I think we may be able to have a wrapper to handle this logic for all of our functions at some point.

src/commands/deleteResourceGroup/v2/deleteResourceGroupV2.ts
```diff
@@ -3,72 +3,110 @@
...
 import { settingUtils } from '../../../utils/settingUtils';
+import { createSubscriptionContext } from '../../../utils/v2/credentialsUtils';
 import { DeleteResourceGroupContext } from '../DeleteResourceGroupContext';
 import { DeleteResourceGroupStep } from '../DeleteResourceGroupStep';
 
 export async function deleteResourceGroupV2(context: IActionContext, primaryNode?: GroupingItem, selectedNodes?: GroupingItem[]): Promise<void> {
+
+    // unset nodes that are not resource groups
+    selectedNodes = selectedNodes?.filter(n => !!n.resourceGroup);
+    if (!primaryNode?.resourceGroup) {
```
nturinski, line 25: Is this property unique to resource group resources? I'm pretty sure it exists on other resources. Shouldn't we check the `type` property?

src/commands/deleteResourceGroup/v2/deleteResourceGroupV2.ts
```diff
@@ -3,72 +3,110 @@
...
-            // TODO: Deal with picker.
-            // const subscription: SubscriptionTreeItem = await ext.appResourceTree.showTreeItemPicker(SubscriptionTreeItem.contextValue, context);
-            // selectedNodes = await subscription.pickResourceGroup(context, {
-            //    canPickMany: true,
-            //    placeholder: localize('selectResourceGroupToDelete', 'Select resource group(s) to delete'),
-            //});
+            ({ subscription, resourceGroupsToDelete } = await pickResourceGroups(context));
         }
     } else {
         selectedNodes = selectedNodes.filter(n => !!n.resourceGroup);
```
nturinski, line 41: Do `selectedNodes` also include the `primaryNode`? I could see it technically being "selected".

src/commands/deleteResourceGroup/v2/deleteResourceGroupV2.ts
alexweininger: Good catch. I don't think it currently exists on items other than the resource groups `GroupingItems`, however I could see this causing a bug in the future if it was added to another item.

I'll do an `instanceof` check as well. I really like avoiding using `instanceof` because it can cause other issues, but it's necessary here for now.

src/commands/deleteResourceGroup/v2/deleteResourceGroupV2.ts
alexweininger: I'll test that, regardless it's very confusing 😄

src/commands/deleteResourceGroup/v2/deleteResourceGroupV2.ts
alexweininger: Agreed, the whole selected vs primary is pretty confusing.

src/commands/deleteResourceGroup/v2/deleteResourceGroupV2.ts
nturinski: Why not just check for this `"type": "Microsoft.Resources/resourceGroups"?

alexweininger: The object is a `GroupingItem`, not an `AzureResource`.

src/commands/deleteResourceGroup/v2/deleteResourceGroupV2.ts
alexweininger: So yeah. The primary node is whatever node has been right clicked to bring up the context menu. And it is included in the selected nodes array.

src/commands/deleteResourceGroup/v2/deleteResourceGroupV2.ts
alexweininger: I've added a utility to make it not as ugly.

bwateratmsft: Docker has a similar challenge: https://github.com/microsoft/vscode-docker/blob/main/src/utils/multiSelectNodes.ts

When multiple nodes are selected and an action performed on one of them, we simply ignore those for which the action is not valid.

alexweininger: Thanks Brandon, I'll file an issue on azuretools to create a shared utility for this stuff. Hopefully we can come up with a few utils that cover everyone's use case.

nturinski (approved):

PR #517, opened 2023-01-12
Implement Group By Arm Tag command

### LINKED ISSUE #465: Fix Group by Arm Tag command
Group By doesn't currently work in v2 if run from the command palette since in v1 it relies on `getCachedChildren`.

alexweininger:
Fixes #465

### DIFF
--- a/src/commands/explorer/groupBy.ts
+++ b/src/commands/explorer/groupBy.ts
@@ -3,10 +3,15 @@
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 
-import { IActionContext } from "@microsoft/vscode-azext-utils";
+import { uiUtils } from "@microsoft/vscode-azext-azureutils";
+import { IActionContext, nonNullProp, subscriptionExperience } from "@microsoft/vscode-azext-utils";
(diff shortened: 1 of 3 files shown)


src/commands/explorer/groupBy.ts
```diff
@@ -3,29 +3,39 @@
...
 
 export function buildGroupByCommand(setting: string) {
     return (context: IActionContext): Promise<void> => groupBy(context, setting);
 }
 
 async function groupBy(context: IActionContext, setting: string): Promise<void> {
     if (setting === 'armTag') {
-        const tag = await context.ui.showQuickPick(getQuickPicks(context), {
-            placeHolder: localize('groupByArmTagKey', 'Select the tag key to group by...')
+        const subscription = await subscriptionExperience(context, ext.v2.api.resources.azureResourceTreeDataProvider);
```
nturinski, line 22: Should we move this to `getQuickPicks`? Put all the async code in one spot and all.

nturinski (approved):

src/commands/explorer/groupBy.ts
alexweininger: Since the subscription experience involves showing a quick pick, I think it needs to (and should) stay out of `getQuickPicks` because while `getQuickPicks` is being awaited, we show loading state in the UI.

PR #524, opened 2023-01-19
Don't do case-sensitive Azure resource ID comparison

bwateratmsft:
Fixes a bug where nothing was happening when I was calling for a resource to be revealed

### DIFF
--- a/src/commands/revealResource.ts
+++ b/src/commands/revealResource.ts
@@ -45,14 +45,13 @@ function parsePartialAzureResourceId(id: string): Partial<ParsedAzureResourceId>
 }
 
 function getResourceKindFromId(parsedId: Partial<ParsedAzureResourceId>): 'subscription' | 'resourceGroup' | 'resource' {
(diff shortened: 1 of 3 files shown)


src/commands/revealResource.ts
```diff
@@ -45,14 +45,13 @@ function parsePartialAzureResourceId(id: string): Partial<ParsedAzureResourceId>
 }
 
 function getResourceKindFromId(parsedId: Partial<ParsedAzureResourceId>): 'subscription' | 'resourceGroup' | 'resource' {
-    if (parsedId.subscriptionId) {
```
bwateratmsft, line 48: The logic here was incorrect; the Azure resource ID of, e.g., an app service site includes a subscription, and was mistakenly being identified as a subscription by this logic as a result. Flipping it upside down fixes it. I also added `else`s because I like the way they look a little better for this.

alexweininger (approved):

PR #527, opened 2023-01-23
Dedupe Azure resources

### LINKED ISSUE #526: v2 API - Element with id "..." is already registered
Experienced this while testing static web apps. When you expand any grouping item, it shows the following error (with a different tree item id)
image

Refreshing with things collapsed sometimes fixes it, but the error comes back on subsequent refreshes. Happens regardless of what grouping you've selected.

alexweininger:
Fixes #526 

Also added some telemetry to this method. If this keeps happening we'll be able to give the telemetry to the service team.

### DIFF
--- a/src/api/DefaultAzureResourceProvider.ts
+++ b/src/api/DefaultAzureResourceProvider.ts
@@ -17,18 +17,27 @@ export class DefaultAzureResourceProvider implements AzureResourceProvider {
 
     getResources(subscription: AzureSubscription): Promise<AzureResource[] | undefined> {
         return callWithTelemetryAndErrorHandling(
-            'provideResources',
+            'defaultAzureResourceProvider.getResources',
             async (context: IActionContext) => {
                 const subContext = createSubscriptionContext(subscription);
                 const client = await createResourceClient([context, subContext]);
+
                 // Load more currently broken https://github.com/Azure/azure-sdk-for-js/issues/20380
+                const allResources = await uiUtils.listAllIterator(client.resources.list());
 
-                const allResources: GenericResource[] = await uiUtils.listAllIterator(client.resources.list());
-                const appResources = allResources.map(resource => this.createAppResource(subscription, resource));
+                // dedupe resources to fix https://github.com/microsoft/vscode-azureresourcegroups/issues/526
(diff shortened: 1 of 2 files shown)


src/api/DefaultAzureResourceProvider.ts
```diff
@@ -17,18 +17,27 @@ export class DefaultAzureResourceProvider implements AzureResourceProvider {
...
+                const allResources = await uiUtils.listAllIterator(client.resources.list());
+
+                // dedupe resources to fix https://github.com/microsoft/vscode-azureresourcegroups/issues/526
+                const allResourcesDeduped: GenericResource[] = dedupeAzureResources(allResources);
+                context.telemetry.measurements.resourceCount = allResourcesDeduped.length;
 
-                const allResources: GenericResource[] = await uiUtils.listAllIterator(client.resources.list());
-                const appResources = allResources.map(resource => this.createAppResource(subscription, resource));
+                if (allResourcesDeduped.length !== allResources.length) {
+                    context.telemetry.properties.duplicateResources = 'true';
```
bwateratmsft, line 33: Should we capture the # of duplicates too?

bwateratmsft (approved):

src/api/DefaultAzureResourceProvider.ts
nturinski: I agree. I think it'll be a more compelling data point.

src/api/DefaultAzureResourceProvider.ts
```diff
@@ -70,3 +79,7 @@ export class DefaultAzureResourceProvider implements AzureResourceProvider {
         };
     }
 }
+
+function dedupeAzureResources(array: GenericResource[]): GenericResource[] {
+    return [...new Map(array.map((item) => [item.id, item])).values()];
```
nturinski, line 84: Might make sense to capture the telemetry in here. I know we have to pass more args then to do the calculation, but I think it would be nice to have it all encapsulated in this function especially if we ever end up using it in other places.

alexweininger: I think I'll make a function that lists the resources, dedupes them, and also records telemetry.

bwateratmsft (approved):

PR #528, opened 2023-01-24
Megan/auth provider

nturinski:
Pull in changes from [Philip's VS Code auth provider branch](https://github.com/microsoft/vscode-azureresourcegroups/tree/philliphoff-drop-aa)

### DIFF
--- a/package.json
+++ b/package.json
@@ -301,7 +301,7 @@
             ],
             "view/item/context": [
                 {
-                    "command": "azureResourceGroups.selectSubscriptions",
+                    "command": "azureResourceGroups..accounts.selectSubscriptions",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -282,7 +301,7 @@
             ],
             "view/item/context": [
                 {
-                    "command": "azureResourceGroups.selectSubscriptions",
+                    "command": "azureResourceGroups..accounts.selectSubscriptions",
```
alexweininger, line 304: extra `.` here?

nturinski: It's how Phillip originally wrote it, but we can always change it up.

nturinski: Oh wait, you're right, I could barely see that with the highlighting 👍

PR #529, opened 2023-01-24
Fix `AzExtTreeItem.deleteTreeItem` compat

alexweininger:
We need to call refresh when a tree item calls removeChildFromCache on the parent. That way the removed items actually go away.

Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/783

I tested with VMs, but fix should apply to all extensions equally.

### DIFF
--- a/src/api/compatibility/application/CompatibleApplicationResourceTreeItem.ts
+++ b/src/api/compatibility/application/CompatibleApplicationResourceTreeItem.ts
@@ -69,7 +69,7 @@ export class CompatibleResolvedApplicationResourceTreeItem extends AzExtParentTr
                 subscription: __subscription,
                 parent: undefined,
                 removeChildFromCache: () => {
-                    //
+                    this.treeDataProvider.refreshUIOnly(undefined);
                 },
             }) as unknown as AzExtParentTreeItem
         );


src/api/compatibility/application/CompatibleApplicationResourceTreeItem.ts
```diff
@@ -69,7 +69,7 @@ export class CompatibleResolvedApplicationResourceTreeItem extends AzExtParentTr
                 subscription: __subscription,
                 parent: undefined,
                 removeChildFromCache: () => {
-                    //
+                    this.treeDataProvider.refreshUIOnly(undefined);
```
bwateratmsft, line 72: It's for compat so it matters less, but do we have a handle to the child's parent, to scope the refresh?

bwateratmsft (approved):

src/api/compatibility/application/CompatibleApplicationResourceTreeItem.ts
alexweininger: Follow up: I've had that scoped refresh thing in a git stash, will make a PR for it.

PR #530, opened 2023-01-24
Create package for consuming extension API

alexweininger:
Summary: 
The API v2 type declarations that used to live in utils as `hostapi.v2.d.ts` are now derived from `/api` in this repo.

The main benefit of this is that the API types and implementation are in the same place. It also allows us to publish a package which makes consuming the API more convenient.

`/api` contains:
* v2 API types and interfaces
* `AzExtResourceType` enum
* A `getAzureResourcesExtensionApi` utility for consuming the API
* `Wrapper` and `isWrapper`

You can imagine the code in `/api` as code that lives inside `/src` that has been moved into a new folder. It's imported directly by the code in `/src`. As far as resource groups is concerned, it's not published to npm. The code is bundled into the extension via webpack like normal.

I'm also using the [api-extractor](https://api-extractor.com/) tool here to create the types for the api package. It currently rolls up all the types produces by `tsc` and puts them into a single file. I couldn't get this to work if I used `.d.ts` files, so I put the typings into `.ts` files. I think that's fine.

Running `npm run api-extractor` generates the types file `/api/vscode-azureresources-api.d.ts`, when making changes to the api we'll need to remember to do this. This file is tracked by git so we can track changes to it.  I will investigate adding a commit hook for this later.

[See related changes in utils](https://github.com/microsoft/vscode-azuretools/pull/1356)

Follow up PRs:
* Add some meta files like readme and changelog
* Add a GitHub Action for publishing the package to npm
* Add api documentation here instead of in tools
* Add tests for things that were moved from utils to here, specifically `isWrapper` tests. They can go into the resource groups tests.


api/package.json
```diff
@@ -0,0 +1,22 @@
...
+        "compile": "tsc -p ./ -watch",
+        "prepack": "npm run build",
+        "api-extractor": "npm run build && api-extractor run"
+    },
+    "author": "Microsoft",
+    "license": "MIT",
+    "devDependencies": {
+        "@azure/ms-rest-azure-env": "^2.0.0",
+        "@microsoft/api-extractor": "^7.33.8",
+        "@microsoft/vscode-azext-utils": "^0.4.0",
```
alexweininger, line 18: Ideally this wouldn't depend on `utils` at all. However the activity types are unable to be moved here because they rely on `AzExtTreeItem`, and I really don't want that here.

alexweininger: But it's not a huge deal since it's just a dev dependency.

tsconfig.json
```diff
@@ -15,16 +15,21 @@
         "noUnusedParameters": true,
         "baseUrl": "./",
         "paths": {
+            "@hostapiv2": [
```
alexweininger, line 18: I'm trying out using the [paths](https://www.typescriptlang.org/tsconfig#paths) feature for importing things from the api folder.

It's sorta quirky when doing the auto-imports quick fix, but we can always change everything to be path based later.

api/package.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "@microsoft/vscode-azureresources-api",
```
alexweininger, line 2: `@microsoft/vscode-azureresources-api` is my favorite name because to me it makes the most sense that the API for the Azure Resources extension would be listed as `vscode-azureresources-api`. Only unfortunate part is that the extension id is actually `vscode-azureresourcegroups`, but I think it's fine to base the API package on the extension name and not the id.

Also `@microsoft/vscode-azureresourcegroups-api` is really long.

My second favorite is `@microsoft/vscode-azext-host-api` because it's shorter, and also doesn't have to deal with the naming/id awkwardness.

tsconfig.json
alexweininger: Never mind, I can't get this to work properly

api/src/utils/apiUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Gets the exported API from the given extension id and version range.
+     *
+     * @param extensionId - The extension id to get the API from
+     * @param apiVersionRange - The version range of the API you need. Any semver syntax is allowed. For example "1" will return any "1.x.x" version or "1.2" will return any "1.2.x" version
+     * @param options - The options to pass when creating the API. If `options.extensionId` is left undefined, it's set to the caller extension id.
+     * @throws Error if extension with id is not installed.
+     */
+    export async function getAzureExtensionApi<T extends AzureExtensionApi>(context: vscode.ExtensionContext, extensionId: string, apiVersionRange: string, options?: GetApiOptions): Promise<T> {
```
bwateratmsft, line 54: I think it makes more sense for this to stay in the `@microsoft/vscode-azext-utils` package.

api/src/utils/getApi.ts
```diff
@@ -0,0 +1,12 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import type { AzureResourcesExtensionApi } from '../extensionApi';
+import { GetApiOptions, _apiUtils } from "./apiUtils";
+
+export async function getAzureResourcesExtensionApi(extensionContext: vscode.ExtensionContext, apiVersionRange: '2.0.0', options?: GetApiOptions): Promise<AzureResourcesExtensionApi> {
```
bwateratmsft, line 10: And then this can be in this package.

src/AzExtWrapper.ts
```diff
@@ -3,11 +3,11 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { AzExtResourceType, IAzureQuickPickItem } from "@microsoft/vscode-azext-utils";
+import { IAzureQuickPickItem } from "@microsoft/vscode-azext-utils";
 import { AzureExtensionApiProvider } from "@microsoft/vscode-azext-utils/api";
 import { AppResource } from "@microsoft/vscode-azext-utils/hostapi";
-import { AzureResource } from '@microsoft/vscode-azext-utils/hostapi.v2';
 import { commands, Extension, extensions } from "vscode";
+import { AzExtResourceType, AzureResource } from '../api/src/index';
```
bwateratmsft, line 10: Relative imports instead of package imports?

api/vscode-azureresources-api.d.ts
```diff
@@ -0,0 +1,417 @@
+import type { Activity } from '@microsoft/vscode-azext-utils/hostapi';
```
bwateratmsft, line 1: Is it possible to cause comments to be written at the top of this file? I'm thinking:
1. Copyright header
2. A note that it is a generated file and manual changes would be overwritten
3. Instructions on how to generate it

Though, would it be more appropriate to generate it only for the build, and not check it in?

alexweininger: > Is it possible to cause comments to be written at the top of this file? I'm thinking:
> 
> 1. Copyright header
> 2. A note that it is a generated file and manual changes would be overwritten
> 3. Instructions on how to generate it

Yes, adding a comment or at the very least a readme with those is needed. Sadly it doesn't seem possible to add a comment using api-extractor alone. But we could add a script to do it.

> Though, would it be more appropriate to generate it only for the build, and not check it in?

I thought about this and I think it's best to check it in:
1. I want to be able to track changes to it
2. In case api-extractor does anything weird or anything we don't like we can catch it

I'm open to generating it at build if you think it's better.

bwateratmsft: I don't feel strongly about it and I think those are good reasons. One counterpoint is that, if we _don't_ pack only the built version, then the checked-in file could get out of sync with what actually is built (whether we rerun api-extractor every build or not, either way what's checked in could get out-of-sync with what's build). We can certainly try to enforce it at PR time but the potential for mistakes is definitely there; I feel like we make that sort of mistake pretty regularly in the `@microsoft/vscode-azext-utils` package. 

RE: copyright header, it stinks we'd have to go as far as writing a script to update this file with those things, it seems like an obvious feature for api-extractor to have a config for.

api/src/utils/apiUtils.ts
alexweininger: And then depend on utils as a runtime dependency? It'd be a circular dependency if this package depends on utils at runtime.

bwateratmsft: Ah, yeah. That's fair. Maybe put it just in here, and reexport it from the utils package?

alexweininger: Oh yeah, haven't thought of that, genius!

src/AzExtWrapper.ts
alexweininger: I know it's a bit weird, but that was a choice I made on purpose. I want the code to be a native part of the resource groups extension, that also happens to be published as a package. This is a pretty common pattern for monorepos.

I wasn't so happy with having to import from a different "project" like this, so I did create an alternative:

Instead of putting the API code into a different folder from the extension code, I just put it in a folder within the extension src. And then you can easily configure a TS Config file to target only that folder and it builds only that folder. It works exactly the same, except the code lives in the same src directory.

bwateratmsft: That's reasonable. I'll take a look at #532 and see what looks best.

api/vscode-azureresources-api.d.ts
alexweininger: > RE: copyright header, it stinks we'd have to go as far as writing a script to update this file with those things, it seems like an obvious feature for api-extractor to have a config for.

https://github.com/microsoft/rushstack/issues/2584

alexweininger: > I don't feel strongly about it and I think those are good reasons. One counterpoint is that, if we _don't_ pack only the built version, then the checked-in file could get out of sync with what actually is built (whether we rerun api-extractor every build or not, either way what's checked in could get out-of-sync with what's built). We can certainly try to enforce it at PR time but the potential for mistakes is definitely there; I feel like we make that sort of mistake pretty regularly in the `@microsoft/vscode-azext-utils` package.

Ok, I think I'm convinced doing it at build time (npm pack time) is best.

alexweininger: However, I might still want to check it in purely for reference and documentation. (I like to be able to link to it on GitHub).

api/src/utils/apiUtils.ts
alexweininger: Though, eventually it'd make sense to put the api utils and types into their own package as well.

src/AzExtWrapper.ts
bwateratmsft: image

alexweininger: I think I'll just stick with this pattern

alexweininger:
This is probably good to go now 😄

bwateratmsft (approved):

PR #531, opened 2023-01-24
Scoped global refresh for BDPs

alexweininger:
When a BDP does a root refresh, don't refresh the entire tree.

Instead, this refreshes any item which has a child that is provided by said BDP.

### DIFF
--- a/src/tree/ResourceBranchDataProviderManagerBase.ts
+++ b/src/tree/ResourceBranchDataProviderManagerBase.ts
@@ -34,7 +34,14 @@ export abstract class ResourceBranchDataProviderManagerBase<TResourceType, TBran
             type,
             {
                 provider,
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

alexweininger:
This change is causing bugs with creating/deleting resources. We should revert it until we have a better solution.

The issue is since BDPs were able to signal a root refresh which resulted in refetching the list of resources, BDPs used it to signal resources had been created/deleted. With this change, the list is never updated and Azure resources thinks the deleted resource should still exist even if it doesn't, which results in errors.

However, I don't think it was ever intended for BDPs to be able to do this. It should be the responsibility of the resource provider to signal that the list of resources needs updating. The issue with that is the default Azure resource provider is built-in, and clients have no way of signaling that the list needs to be refetched.

We could make it so that when the list is refetched, only parts of the tree with new or removed nodes are refreshed. Or we could keep it simple and revert to always refetching the list, and refreshing the entire tree again.

PR #540, opened 2023-02-01
Fixup workflow_dispatch to allow manually running workflow

alexweininger:


.github/workflows/api-publish.yaml
```diff
@@ -2,7 +2,7 @@ name: 'API: Publish to npm'
 # Publishes the api package to npm
 
 on:
-  workflow_call:
+  workflow_dispatch:
```
bwateratmsft, line 5: It's always seemed kinda weird that you have to specify a trigger _at all_ to be able to manually run something.

bwateratmsft (approved):

.github/workflows/api-publish.yaml
alexweininger: I guess they want workflows to be automated-only by default. 🤷

PR #542, opened 2023-02-01
Webpack and VS Code Auth for WebExt

nturinski:
Incorporates the changes to make this run as a web extension. 

Note that if you debug this in VS Code Desktop as a web extension, the `UIKind` _does not_ change to reflect that. It effectively makes testing/debugging impossible 🙄

I had to change the name of the commands of the `extension.ts` to activate/deactivate since that's what VS Code is expecting. VS Code Web doesn't allow more than one file, so you can't use `main.js` as an entry point without breaking a bunch of things.

Also, path and open do not work in VS Code Web so those were replaced.


package.json
```diff
@@ -577,6 +578,7 @@
         "@vscode/test-electron": "^2.1.5",
         "eslint": "^7.25.0",
         "eslint-plugin-import": "^2.22.1",
+        "fs-extra": "^11.1.0",
```
bwateratmsft, line 581: Do we actually need `fs-extra` (even as a dev dependency)?

src/services/AzureSubscriptionProvider.ts
```diff
@@ -3,31 +3,17 @@
 *  Licensed under the MIT License. See License.md in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 import * as arm from '@azure/arm-subscriptions';
+import type { Environment } from '@azure/ms-rest-azure-env';
 import { uiUtils } from '@microsoft/vscode-azext-azureutils';
+import { AzureSubscription } from '@microsoft/vscode-azext-utils/hostapi.v2';
 import * as vscode from 'vscode';
+import { AzureLoginStatus } from '../tree/azure-account.api';
```
bwateratmsft, line 10: Let's just dump this now...

src/services/AzureSubscriptionProvider.ts
```diff
@@ -3,31 +3,17 @@
 *  Licensed under the MIT License. See License.md in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 import * as arm from '@azure/arm-subscriptions';
+import type { Environment } from '@azure/ms-rest-azure-env';
 import { uiUtils } from '@microsoft/vscode-azext-azureutils';
+import { AzureSubscription } from '@microsoft/vscode-azext-utils/hostapi.v2';
```
bwateratmsft, line 8: Can we use the latest that has this in the RG package?

src/tree/azure/VSCodeAuthentication.ts
```diff
@@ -0,0 +1,53 @@
...
+export function createSubscriptionContext(subscription: AzureSubscription): ISubscriptionContext {
+    return {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+        environment: {} as Environment,
+        isCustomCloud: false,
+        subscriptionDisplayName: subscription.name,
+        subscriptionId: subscription.subscriptionId,
+        subscriptionPath: '',
+        tenantId: '',
+        userId: '',
```
bwateratmsft, line 50: I'm concerned about potential impact of putting these strings in blank. The eventual consumer of the `ISubscriptionContext` might not expect that.

src/utils/openUrl.ts
```diff
@@ -3,11 +3,9 @@
...
  *--------------------------------------------------------------------------------------------*/
 
-import * as open from 'open';
+import * as vscode from 'vscode';
 
 export async function openUrl(url: string): Promise<void> {
     // Using this functionality is blocked by https://github.com/Microsoft/vscode/issues/85930
-    // await vscode.env.openExternal(vscode.Uri.parse(url));
-
-    await open(url);
```
bwateratmsft, line 12: Woo get rid of `open`!

package.json
nturinski: This is just for the gulpfile, so we technically don't need it. But we can't use the vscode `fs`  to replace it.

src/tree/azure/VSCodeAuthentication.ts
nturinski: Should be resolved Megan's changes

api/.npmignore
```diff
@@ -13,7 +13,10 @@ test-results.xml
 .eslintrc.js
 resources/*.png
 *.tsbuildinfo
+<<<<<<< HEAD
 docs/
+=======
+>>>>>>> 4c48d43 (Create package for consuming extension API (#530))
```
bwateratmsft, line 19: Oops

PR #551, opened 2023-02-07
Fix Open in Portal command

motm32:


src/services/AzureSubscriptionProvider.ts
```diff
@@ -3,7 +3,7 @@
 *  Licensed under the MIT License. See License.md in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
 import * as arm from '@azure/arm-subscriptions';
-import type { Environment } from '@azure/ms-rest-azure-env';
+import { Environment } from '@azure/ms-rest-azure-env';
```
bwateratmsft, line 6: This package will need to be added as an explicit dependency in package.json

src/services/AzureSubscriptionProvider.ts
```diff
@@ -184,7 +184,7 @@ export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implement
                     authentication: {
                         getSession: () => session
                     },
-                    environment: {} as Environment,
+                    environment: Environment.AzureCloud,
```
alexweininger, line 187: Can we hardcode this to `AzureCloud` or do we need to support other environments?

motm32: I think we're only trying to support `AzureCloud` at the moment.

alexweininger: Ok, that's fine. Are we planning to add support for the other clouds later?

bwateratmsft (approved):
Code looks fine to me but please work out the sovereign/custom cloud question before merging :smile:

src/services/AzureSubscriptionProvider.ts
motm32: For the first iteration were not planning to support any other clouds, but as we migrate the vs code auth provider we may add support for sovereign/custom cloud.

PR #552, opened 2023-02-07
Report if extension was installed via command

alexweininger:
This will allow us to get some useful insights about how often users are installing an extension from the "Install extension to enable additional features..." tree item.

image

### DIFF
--- a/src/commands/installExtension.ts
+++ b/src/commands/installExtension.ts
@@ -18,12 +18,11 @@ export async function installExtension(context: IActionContext, extensionId: str
     context.telemetry.properties.alreadyInstalled = String(alreadyInstalled);
 
     if (!alreadyInstalled) {
+        context.telemetry.properties.installedAfter1m = 'false';
         return new Promise((resolve) => {
             const disposable = extensions.onDidChange(() => {
-                const installed = isInstalled();
-                context.telemetry.properties.installedAfter1m = String(isInstalled());
-
-                if (installed) {
+                if (isInstalled()) {
+                    context.telemetry.properties.installedAfter1m = 'true';
                     disposable.dispose();
                     clearTimeout(timeout);
                     resolve();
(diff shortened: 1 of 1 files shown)


src/commands/installExtension.ts
```diff
@@ -4,9 +4,37 @@
...
+
+                if (installed) {
+                    disposable.dispose();
+                    clearTimeout(timeout);
+                    resolve();
+                }
+            });
+
+            const timeout = setTimeout(() => {
+                context.telemetry.properties.installedAfter30s = 'false';
```
alexweininger, line 34: If it's been 30 seconds and they haven't installed the extension, I think it's safe to assume they didn't install it.

bwateratmsft: Ehh, I dunno. Not everyone has fast internet. Especially for really large extensions (looking at you, storage), it could easily take longer than 30 seconds.

There's no right time limit, since internet speeds range from a 56K to multi-gigabit--so what happens when we exceed the timeout becomes important. It looks like for this, it's just telemetry?

alexweininger: It's just for telemetry. And I'm open to extending the timeout. Maybe 1 or 2 minutes?

bwateratmsft: I think the main downside to longer times is that there's a chance the event never gets sent if the user closes VSCode before download completes. I think 1 minute is reasonable; it would cover the overwhelming majority of cases, enough to give some solid data, hopefully without losing many to dropoffs.

bwateratmsft (dismissed):
Maybe add a comment explaining what happens if the timeout is reached (we just send an event saying it wasn't installed after 1m), or if VSCode closes before download finishes (event probably gets dropped)?

alexweininger:
We could probably be fancier here and send two events, one when the command is executed. And another to follow up with the installation result. But I think this implementation is fine.

bwateratmsft (approved):

PR #566, opened 2023-02-10
Publish web bundle and package.json to a blob container for testing on vscode.dev

MRayermannMSFT:
Adds steps to the package.yml for:
1. Publishing bits for testing vscode.dev to a blob container
2. Generating a URL with a SAS token to the directory for the build
3. Publishing that SAS as a build artifact, which you can then give to vscode.dev

Some FYIs:
- Target storage account and container are defined by a pipeline variable.
- The SAS URL is currently both outputted to the build logs and uploaded as a build artifact. If ADO ever starts to redact the SAS in the logs, we can refer to the artifact until we work aroudn the redaction. To use the URL, copy and paste it, and then use it with the vscode.dev command to install from a location.
- The SAS is only good for 30 days. If you need to test a build for longer you can either:
   1. Generate a SAS on your own using Storage Explorer.
   2. Just use a SAS from a new-build on the URL for the build-you-want-to-test, since the SAS's are good for any blob/directory in the container.
- Blob service CORS rules have been configured for the storage account. You can go view those in Storage Explorer if you'd like. I'm allowing any host to make any GET request. Went with any host because I saw a variety of hosts making requests and it didn't seem worth it to try and figure out all of them. If we want to restrict the hosts we should probably just get a list from the VS Code team.
- A data lifecycle rule for deleting blobs which have not been modified in 90 days has been configured for the storage account.
- The container has been leased to help avoid being deleted.

Example build which is compatible with vscode.dev can be found here: https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=58210&view=results

### DIFF
--- a/.azure-pipelines/common/package.yml
+++ b/.azure-pipelines/common/package.yml
@@ -15,12 +15,74 @@ steps:
   displayName: 'Copy vsix to staging directory'
   inputs:
     Contents: '**/*.vsix'
-    TargetFolder: '$(build.artifactstagingdirectory)'
+    TargetFolder: '$(build.artifactstagingdirectory)/vsix'
+
+- task: CopyFiles@2
+  displayName: 'Copy web bundle to staging directory'
+  inputs:
+    Contents: 'dist/web/*.js*'
+    TargetFolder: '$(build.artifactstagingdirectory)/web/'
+
+- task: CopyFiles@2
+  displayName: 'Copy package.json to staging directory'
+  inputs:
+    Contents: 'package.json'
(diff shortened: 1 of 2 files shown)


alexweininger:
> The SAS has to be published as a build artifact because Azure pipelines redacts it if you just log it. To use it, you simply open the text file, copy and paste the URL, and then use it with the vscode.dev command to install from a location.

That's certainly good enough for now. If we want to get fancy later we can have it create a comment on the PR with the SAS Url.

alexweininger (approved):

nturinski (approved):

Issue #543, opened 2023-02-02
Web Extension Runtime Issues


nturinski:
The following commands do not work in vscode.dev

- [x] `Azure: Create Resource`
  - Freezes on the location list step (something is probably wrong with the request)
  - https://github.com/microsoft/vscode-azuretools/pull/1381
  - https://github.com/microsoft/vscode-azureresourcegroups/pull/579
- [x] `Azure: Create Resource Group`
  - Should be the same as `Create Resource`
  - https://github.com/microsoft/vscode-azureresourcegroups/pull/579
- [x] `Azure: Edit Tags`
  - Something is wrong with the buffer
  - https://github.com/microsoft/vscode-azureresourcegroups/pull/567
  - New error `Error "parameters.properties.tags must be of type object." occurred in serializing the payload - undefined.`
  - https://github.com/microsoft/vscode-azureresourcegroups/pull/580
- [x] `Azure: Open in Portal`
  - Unable to read file (probably have to convert to VS Code fs API)
  - https://github.com/microsoft/vscode-azureresourcegroups/pull/551

alexweininger:
> Unable to read file (probably have to convert to VS Code fs API)

I wasn't aware the open in portal ever read a file?

PR #567, opened 2023-02-13
Add `Buffer` Node standard library to `webpack` bundle

MicroFish91:
Should solve the issue we're seeing with `edit tags` in #543

I'm using the `node-stdlib-browser` package as a dev dependency to help us bring in Node's `Buffer` library, let me know if there is another package or method we would prefer using instead to solve this issue.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -13,6 +13,9 @@ const process = require('process');
 const dev = require("@microsoft/vscode-azext-dev");
 const webpack = require('webpack');
 
+const stdLibBrowser = require('node-stdlib-browser');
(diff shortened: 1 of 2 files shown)


bwateratmsft (approved):

webpack.config.js
```diff
@@ -45,7 +48,15 @@ const webConfig = dev.getDefaultWebpackConfig({
         '../build/default/bufferutil': 'commonjs ../build/default/bufferutil',
     },
     target: 'webworker',
-
+    resolve: {
```
nturinski, line 51: I don't think you need this line. `getDefaultWebpackConfig` doesn't actually extract out the options for `resolve.alias`. And if it's working without it, then I guess that's okay.

nturinski (approved):

PR #572, opened 2023-02-14
Publish resources and package.nls.json for vscode.dev testing

MRayermannMSFT:


bwateratmsft (commented):
How did `package-lock.json` end up with sooo many changes?

MRayermannMSFT:
> How did `package-lock.json` end up with sooo many changes?

Because I did a full update of it. Build was failing during the `npm ci` phase. Figured a refresh of the lock should help resolve that, which it did. If you'd rather it be done separately or more surgically, I don't mind, but in that case I'll pull that work out of this change.

bwateratmsft (approved):
That makes sense. I'm fine with it, just was surprised by the volume and wasn't sure if it was intentional or not.

PR #575, opened 2023-02-15
Fix refreshing subscription after creating a RG

### LINKED ISSUE #569: The created resource group doesn't show in the resource list
**OS:** All
**Build Version:** 20230213.13
**Regression:** Previous release

**Repro Steps:**
1. Make sure the resource groups by Resource Group.
2. Create a resource group.
3. Check whether the created resource group shows in the resource list

**Expect:**
 The created resource group shows in the resource list.

**Actual:**
The created resource group doesn't show in the resource list.

**More Info:**
1. The resource list view does not auto-refresh when deleting resource group.
2. Refresh can fix this issue.

alexweininger:
Fix #569

### DIFF
--- a/src/commands/createResourceGroup.ts
+++ b/src/commands/createResourceGroup.ts
@@ -37,5 +37,5 @@ export async function createResourceGroup(context: IActionContext, node?: Subscr
     if (!wizardContext.suppressNotification) {
         void window.showInformationMessage(localize('createdRg', 'Created resource group "{0}".', newResourceGroupName));
     }
-    ext.azureTreeState.notifyChildrenChanged(subscription.subscriptionId);
+    ext.azureTreeState.notifyChildrenChanged(`/subscriptions/${subscription.subscriptionId}`);
 }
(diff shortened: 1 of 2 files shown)


src/tree/ResourceTreeDataProviderBase.ts
```diff
@@ -90,7 +90,7 @@ export abstract class ResourceTreeDataProviderBase extends vscode.Disposable imp
                 if (child instanceof BranchDataItemWrapper) {
                     return child;
                 }
-                return this.state.wrapItemInStateHandling(child, (item) => this.notifyTreeDataChanged(item));
```
alexweininger, line 93: I'm going to refactor and rename this method so it makes more sense when it should be used vs. when the emitter should be used directly.

The gist is that `notifyTreeDataChanged` is for branch items. We need to rename it so that is easily understood from the code.

nturinski (approved):

src/commands/createResourceGroup.ts
```diff
@@ -37,5 +37,5 @@ export async function createResourceGroup(context: IActionContext, node?: Subscr
     if (!wizardContext.suppressNotification) {
         void window.showInformationMessage(localize('createdRg', 'Created resource group "{0}".', newResourceGroupName));
     }
-    ext.azureTreeState.notifyChildrenChanged(subscription.subscriptionId);
+    ext.azureTreeState.notifyChildrenChanged(`/subscriptions/${subscription.subscriptionId}`);
```
nturinski, line 40: Can you not just use the full azure id?

alexweininger: What do you mean by full azure id?

PR #577, opened 2023-02-16
rename walkthrough id

esweet431:
- renames the walkthrough id from azure-get-started to get-started

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
         "onView:azureResourceGroups",
         "onView:azureResourceGroupsV2",
         "onView:ms-azuretools.helpAndFeedback",
-        "onWalkthrough:azure-get-started"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -464,7 +464,7 @@
         ],
         "walkthroughs": [
             {
-                "id": "azure-get-started",
+                "id": "get-started",
```
bwateratmsft, line 467: I'd suggest `azureStart`, `azrgStart` or something like that. `get-started` is too generic and likely to collide with some other extensions' walkthrough.

esweet431: good call - changed to `azrgStart`

bwateratmsft (approved):

esweet431:
Closing because we were able to work around the character limit

PR #580, opened 2023-02-16
Reset byteOffset when reading unitArray from VS Codes writeFile

nturinski:
Also replace the buffer polyfill with something a bit more popular/straight-forward


src/commands/tags/TagFileSystem.ts
```diff
@@ -76,7 +76,10 @@ export class TagFileSystem extends AzExtTreeFileSystem<ITagsModel> {
     }
 
     public async writeFileImpl(context: IActionContext, model: ITagsModel, content: Uint8Array, originalUri: Uri): Promise<void> {
-        const text: string = content.toString();
+        // weird issue when in vscode.dev, the content Uint8Array has a giant byteOffset that causes it impossible to decode
```
alexweininger, line 79: You may want to ask around about this, maybe there is a reason for the offset. At the very least it'd be nice to know why.

alexweininger (approved):

bwateratmsft (approved):

src/commands/tags/TagFileSystem.ts
nturinski: Asked in https://github.com/microsoft/vscode-discussions/discussions/469

PR #583, opened 2023-02-17
Show workspace resources even if workspace isn't open

alexweininger:
Instead of not displaying resources if no workspace is opened, show resources and also add a tree item that prompts users to open a folder/workspace.

image

This more closely mirrors the current behavior, and fixes https://github.com/microsoft/vscode-azurestorage/issues/1214

### DIFF
--- a/package.json
+++ b/package.json
@@ -244,11 +244,6 @@
             },
             {
                 "view": "azureWorkspace",
-                "contents": "No workspace has been opened.",
-                "when": "azureWorkspace.state == 'noWorkspace'"
-            },
-            {
-                "view": "azureWorkspace",
                 "contents": "No local workspace resources exist.",
                 "when": "azureWorkspace.state == 'noWorkspaceResources'"
             },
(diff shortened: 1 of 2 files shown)


nturinski:
Have we explored other solutions? I thought I remember Phillip suggestion to have a "getting started" view if there were no tree items. For example:

image
image

alexweininger:
> Have we explored other solutions? I thought I remember Phillip suggestion to have a "getting started" view if there were no tree items.

Yes Phillip actually implemented 3 different welcome views for the workspace view. The PR removes the "no workspace" welcome view. The "no workspace resources" and "no workspace resource providers" welcome views remain.

nturinski (approved):
I think we should consider improving these views with some buttons to get users started, but I won't block for that.

alexweininger:
> I think we should consider improving these views with some buttons to get users started, but I won't block for that.

Yeah, I wasn't quite sure what to put in terms of buttons. It'd be nice to have a button that took them to the extensions view but only showed extensions that contribute workspace resource providers.

PR #586, opened 2023-02-21
Fix default web app to deploy compatibility

alexweininger:
During v1, the tree item ids were not in the ARM id format. So to support the default resource to deploy feature, which uses find tree item, we must convert the ids.


nturinski (dismissed):
🤮

alexweininger:
> 🤮

Don't worry, I added tests 😄

bwateratmsft (approved):

PR #611, opened 2023-03-16
Migration work for vscode.dev

nturinski:
TODO: Will leave clarifying comments in the PR review


src/services/AzureSubscriptionProvider.ts
```diff
@@ -0,0 +1,270 @@
...
+    onFiltersChangedEvent: vscode.Event<void>;
+    onSessionsChangedEvent: vscode.Event<void>;
+    onSubscriptionsChangedEvent: vscode.Event<void>;
+
+    readonly waitForFilters: () => Promise<boolean>;
+    readonly waitForLogin: () => Promise<boolean>;
+    readonly waitForSubscriptions: () => Promise<boolean>;
+}
+
+export class VSCodeAzureSubscriptionProvider extends vscode.Disposable implements AzureSubscriptionProvider {
```
bwateratmsft, line 42: Doesn't have to be in this PR, but IMO we should put this into a package sooner than later. We already _know_ we are going to have to, in order to support extensions like Docker.

src/utils/openUrl.ts
```diff
@@ -3,11 +3,14 @@
...
-    // await vscode.env.openExternal(vscode.Uri.parse(url));
+export async function openUrl(url: string | Uri): Promise<void> {
+    // Workaround according to https://github.com/Microsoft/vscode/issues/85930
+    // When a string is passed to openExternal, it is not encoded by VS Code which seems
+    // to be what is causing issues with some URLs.
+    // https://github.com/microsoft/vscode/blob/2edb004e8386b2a86b26e8f30ea6969f4f26ffa7/src/vs/editor/browser/services/openerService.ts#L223-L226
+    // So if a URI is causing errors, try passing a string
 
-    await open(url);
+    await env.openExternal(url as unknown as Uri);
```
bwateratmsft, line 15: Can we just fix this in the utils package and reuse it here?

src/utils/v2/credentialsUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+import { AzExtServiceClientCredentials, ISubscriptionContext } from '@microsoft/vscode-azext-dev';
+import * as vscode from 'vscode';
+import { AzureSubscription } from '../../../api/src/index';
+import * as vscodeAccount from '../../tree/azure/VSCodeAuthentication';
+import { localize } from '../../utils/localize';
+
+/**
+ * Converts a VS Code authentication session to an Azure Track 1 & 2 compatible compatible credential.
+ */
+export function createCredential(getSession: (scopes?: string[]) => vscode.ProviderResult<vscode.AuthenticationSession>): AzExtServiceClientCredentials {
```
bwateratmsft, line 15: How does this differ from `src/tree/azure/VSCodeAuthentication.ts`?

src/extension.ts
```diff
@@ -37,13 +38,17 @@ import { registerWorkspaceTree } from './tree/workspace/registerWorkspaceTree';
...
+    // the perf stats don't matter for vscode because there is no main file to load-- we may need to see if we can track the download time
+    perfStats ||= { loadStartTime: Date.now(), loadEndTime: Date.now() };
+
     ext.context = context;
     ext.ignoreBundle = ignoreBundle;
     ext.outputChannel = createAzExtOutputChannel('Azure Resource Groups', ext.prefix);
     context.subscriptions.push(ext.outputChannel);
 
-    registerUIExtensionVariables(ext);
+    await registerUIExtensionVariables(ext);
```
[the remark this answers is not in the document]
alexweininger: Was this a mistake? I don't see where `registerUIExtensionVariables` was changed to async anywhere.

package.json
```diff
@@ -595,14 +595,12 @@
...
         "buffer": "^6.0.3",
         "jsonc-parser": "^2.2.1",
         "uuidv4": "^6.2.13",
         "vscode-nls": "^5.0.1",
         "vscode-uri": "^3.0.7"
     },
-    "extensionDependencies": [
-        "ms-vscode.azure-account"
-    ]
+    "extensionDependencies": []
```
nturinski, line 605: @alexweininger Could you revert this?

alexweininger: Yes

alexweininger: What was the final decision for handling this anyway? Do I need to start working on the GH Action?

src/extension.ts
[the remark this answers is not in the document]
alexweininger: Bump

package.json
nturinski: I don't think we need it for this release, but if you can whip it out really quickly.

We would want it to edit the `package.json`'s extension dependency and then just `vsce publish --target web`, I think

nturinski: Yeah, I think we should be able to edit that upload script fairly easily.

src/extension.ts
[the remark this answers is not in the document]
nturinski: It's a bit convoluted, but at some point, `getPackageJson` was changed to async while I was working on this.  However, Alex implemented it with the vscode api, which is synchronous, so I reverted those changes.

So this will no longer be async.

src/utils/v2/credentialsUtils.ts
nturinski: image

package.json
```diff
@@ -547,6 +578,7 @@
         "@vscode/vsce": "^2.17.0",
         "eslint": "^7.25.0",
         "eslint-plugin-import": "^2.22.1",
+        "fs-extra": "^11.1.0",
```
bwateratmsft, line 581: Can we use VSCode's FS instead (or, the one in Utils that uses VSCode's)?

nturinski: It's actually not even used. Not really why I ever installed this as a devDependency at any point

alexweininger (approved):
I'm just a tiny bit scared, but :shipit:

PR #563, opened 2023-02-09
Add API tests

alexweininger:
Leaving as a draft for now since I want to cleanup some of the testing infrastructure before merging. Specifically the resource mocking code. Right now it's thrown together but ideally you could set the resource list to be mocked before each test is run.

I'll probably split this up into separate PRs when I'm ready to merge.

I also want to look into adding any API needed to allow clients to mock subscriptions/resources for their own tests.

PR #616, opened 2023-03-17
Refactor VS Code auth

alexweininger:
Refactor VS Code auth portion of the vscode.dev work to hide the auth implementation details from the rest of the code. Now both the VS Code auth (used in vscode.dev) and the Azure Account API auth are shoehorned into a common interface.

This makes it easier to adapt for testing purposes and is similar to what I've done in https://github.com/microsoft/vscode-azureresourcegroups/pull/563

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -27,7 +27,8 @@ import { registerCommands } from './commands/registerCommands';
 import { registerTagDiagnostics } from './commands/tags/registerTagDiagnostics';
 import { TagFileSystem } from './commands/tags/TagFileSystem';
 import { ext } from './extensionVariables';
-import { VSCodeAzureSubscriptionProvider } from './services/AzureSubscriptionProvider';
+import { createAzureAccountSubscriptionProviderFactory } from './services/DesktopSubscriptionProvider';
+import { createWebSubscriptionProviderFactory } from './services/WebAzureSubscriptionProvider';
 import { AzureResourceBranchDataProviderManager } from './tree/azure/AzureResourceBranchDataProviderManager';
 import { DefaultAzureResourceBranchDataProvider } from './tree/azure/DefaultAzureResourceBranchDataProvider';
 import { registerAzureTree } from './tree/azure/registerAzureTree';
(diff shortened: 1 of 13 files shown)


src/services/DesktopSubscriptionProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export function createAzureAccountSubscriptionProviderFactory(): () => Promise<AzureSubscriptionProvider> {
+    return async () => {
+        const api = await getAzureAccountExtensionApi();
+        return new AzureAccountSubscriptionProvider(api);
+    }
+}
+
+export class AzureAccountSubscriptionProvider implements AzureSubscriptionProvider {
```
bwateratmsft, line 38: Does this need to be exported still?

src/extension.ts
```diff
@@ -64,7 +65,7 @@ export async function activate(context: vscode.ExtensionContext, perfStats: { lo
         activateContext.telemetry.measurements.mainFileLoad = (perfStats.loadEndTime - perfStats.loadStartTime) / 1000;
 
         setupEvents(context);
-        ext.subscriptionProvider = new VSCodeAzureSubscriptionProvider(context.globalState);
+        ext.subscriptionProviderFactory = ext.isWeb ? createWebSubscriptionProviderFactory(context) : createAzureAccountSubscriptionProviderFactory();
```
bwateratmsft, line 68:
```suggestion
        ext.subscriptionProviderFactory ??= ext.isWeb ? createWebSubscriptionProviderFactory(context) : createAzureAccountSubscriptionProviderFactory();
```

Might be useful for unit tests? I dunno.

bwateratmsft (approved):

src/extension.ts
alexweininger: For unit tests, what's worked best for me is adding a `ext.overrideBlahFactory`.

src/services/DesktopSubscriptionProvider.ts
alexweininger: No, it doesn't

bwateratmsft (approved):

nturinski:
I think you need to revert some of the `package.json` changes.

This error message: `12 verbose stack Invalid: lock file's @microsoft/vscode-azext-utils@0.5.1 does not satisfy @microsoft/vscode-azext-utils@0.4.3-alpha` is preventing it from building.

You can use 0.5.2-alpha.0. Should have all the changes of 0.5.1 + the dev stuff required.

alexweininger:
What's going on with this build? 😢

nturinski:
I've never seen this error before:
```
C:\Windows\system32\cmd.exe /D /S /C "C:\hostedtoolcache\windows\node\14.21.3\x64\npm.cmd ci"
npm ERR! assert not accessible from @microsoft/vscode-azext-dev
```

nturinski:
So annoying... I couldn't test it because of the dependency on the Azure Accounts extension in the `package.json`. I can manually change it, but should definitely do something about it on the scripting level.

Anyway, get these two errors when testing it out:

```
Cannot read properties of undefined (reading 'length')
this.subscriptionResultsTask is not a function
```

nturinski (approved):

PR #618, opened 2023-03-20
Refresh subscription instead of group

### LINKED ISSUE #617: Refreshing a group item doesn't behave as expected
Say I opened the Azure Resources view, then created a new Function App in the portal. The expected behavior of refreshing the Function Apps group in the resources view would be that the newly created Function App appears.

Currently, refreshing a group doesn't refetch the list of resources, and will not pick up newly created resources.

The easiest solution would be to make refreshing a group always do a root-refresh. A root-refresh is the only way we currently have to trigger refetching the resource list. 

Adding a way to refetch the resource list without having to refresh the entire tree might be a good thing to add in the future.

edit: Correction, a subscription refresh refreshes the resource list, so we don't have to do a root refresh.

alexweininger:
Fixes #617

### DIFF
--- a/src/commands/registerCommands.ts
+++ b/src/commands/registerCommands.ts
@@ -8,6 +8,7 @@ import { commands } from 'vscode';
 import { ext } from '../extensionVariables';
 import { BranchDataItemWrapper } from '../tree/BranchDataProviderItem';
 import { ResourceGroupsItem } from '../tree/ResourceGroupsItem';
+import { GroupingItem } from '../tree/azure/GroupingItem';
 import { clearActivities } from './activities/clearActivities';
 import { createResource } from './createResource';
 import { createResourceGroup } from './createResourceGroup';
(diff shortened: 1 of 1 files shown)


nturinski (commented):
Could we make the refresh implementation different on the node itself rather than having this weird edge case in the command defintion?

Like, if you call `GroupingTreeItem`'s refresh, it's implementation will handle passing up the subscription/parent node? Or do we not have that capability anymore?

alexweininger:
> Could we make the refresh implementation different on the node itself rather than having this weird edge case in the command defintion?
> 
> Like, if you call `GroupingTreeItem`'s refresh, it's implementation will handle passing up the subscription/parent node? Or do we not have that capability anymore?

We haven't done it that way for v2, however that's probably a good approach.

nturinski:
Is it possible to do that now or should I just go ahead and approve this for merge, and we can leave an issue for it later?

alexweininger:
> Is it possible to do that now or should I just go ahead and approve this for merge, and we can leave an issue for it later?

I will change it sometime today (hopefully). 😄

alexweininger:
Since this is sorta a special case, I think for now this implementation if fine. None of the solutions I have feel right since it's just for this one edge case.

nturinski:
Alright, definitely something to keep in mind when we convert a hefty boy like Functions.

nturinski (approved):

src/commands/registerCommands.ts
```diff
@@ -31,7 +32,12 @@ export function registerCommands(): void {
     // v1.5 client extensions attach these commands to tree item context menus for refreshing their tree items
     registerCommand('azureResourceGroups.refresh', async (context, node?: ResourceGroupsItem) => {
         await handleAzExtTreeItemRefresh(context, node); // for compatibility with v1.5 client extensions
-        ext.actions.refreshAzureTree(node);
+
+        if (node instanceof GroupingItem) {
+            ext.actions.refreshAzureTree(node.parent);
```
bwateratmsft, line 37: (Too late now but) can we add a comment explaining why?

PR #641, opened 2023-03-28
Only keep one object per id cached

### LINKED ISSUE #640: `pickAppResource` can return default item even if branch data provider is registered
How to reproduce:

1. Open a workspace
2. Expand the Azure resources view so that App Services are loading...
3. Run the "Deploy to Web App..." command and select a web app.
4. It will throw a "Cannot read properties of undefined (reading 'createClient')" error because the node returned by pickAppResource is the default item, and not the branch data provider item.

This behavior is causing https://github.com/microsoft/vscode-azureappservice/issues/2440

alexweininger:
Fixes #640, and a bunch of others including:
* "Cannot read properties of undefined (reading 'createClient')" errors.
* Issues related to things not being refreshed on the tree after running a command from the palette.

### Background

This following line was added in https://github.com/microsoft/vscode-azureresourcegroups/pull/510 as a band aid solution to try and fix issues where the item returned by the picker was not reference equal to the same item in the tree view. This made it so calling refresh with the item didn't refresh on the tree among other things.
https://github.com/microsoft/vscode-azureresourcegroups/blob/11e06179ad12c30fb01259bbe8bfc0a87304c8eb/src/api/compatibility/pickAppResource.ts#L19

The fix in #510 only worked sometimes, as seen by the numerous bugs reported by CTI surrounding things not refreshing on the tree when commands are invoked via command palette. It turns out this is because multiple items were being added to the cache with the same id. We were just returning the first one we found with a matching id when iterating through the cached items.

### Solution

The solution I implemented makes it so we only ever have one object in memory for each id. Before creating a new object, we check if we have an object in the cache with that id. 

In psuedocode:
```
if (cache.getItemForId(id)) {
    return cache.getItemForId(id)
} else {
    return createItem(id);
}
```

I've taken extra time on this PR to do a lot of manual testing with the client extensions to make sure this functionality works. I will have CTI do test passes on the clients as well.

### DIFF
--- a/src/tree/BranchDataProviderItem.ts
+++ b/src/tree/BranchDataProviderItem.ts
@@ -33,7 +33,7 @@ export class BranchDataItemWrapper implements ResourceGroupsItem, Wrapper {
     static readonly hasPortalUrlContextValue = 'hasPortalUrl';
 
     constructor(
-        private readonly branchItem: ResourceModelBase,
+        public branchItem: ResourceModelBase,
         private readonly branchDataProvider: BranchDataProvider<ResourceBase, ResourceModelBase>,
         private readonly itemCache: BranchDataItemCache,
         private readonly options?: BranchDataItemOptions) {
@@ -116,9 +116,11 @@ export type BranchDataItemFactory = (branchItem: ResourceModelBase, branchDataPr
 
 export function createBranchDataItemFactory(itemCache: BranchDataItemCache): BranchDataItemFactory {
     return (branchItem, branchDataProvider, options) => {
-        const cachedItem = itemCache.getItemForId(branchItem.id);
+        const cachedItem = itemCache.getItemForId(branchItem.id) as BranchDataItemWrapper | undefined;
         if (cachedItem) {
-            return cachedItem as BranchDataItemWrapper;
+            cachedItem.branchItem = branchItem;
+            itemCache.addBranchItem(branchItem, cachedItem);
+            return cachedItem;
         }
         return new BranchDataItemWrapper(branchItem, branchDataProvider, itemCache, options);
     }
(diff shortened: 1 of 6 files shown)


src/tree/BranchDataProviderItem.ts
```diff
@@ -110,5 +115,11 @@ export class BranchDataItemWrapper implements ResourceGroupsItem, Wrapper {
 export type BranchDataItemFactory = (branchItem: ResourceModelBase, branchDataProvider: BranchDataProvider<ResourceBase, ResourceModelBase>, options?: BranchDataItemOptions) => BranchDataItemWrapper;
 
 export function createBranchDataItemFactory(itemCache: BranchDataItemCache): BranchDataItemFactory {
-    return (branchItem, branchDataProvider, options) => new BranchDataItemWrapper(branchItem, branchDataProvider, itemCache, options);
+    return (branchItem, branchDataProvider, options) => {
+        const cachedItem = itemCache.getItemForId(branchItem.id);
```
bwateratmsft, line 119: Upon refresh what clears the caches?

alexweininger: When the extension host called `getChildren(undefined)` the cache is cleared.

https://github.com/microsoft/vscode-azureresourcegroups/blob/628de8437a48bfaff8f46e66dac647fb26c971f1/src/tree/wrapTreeForVSCode.ts#L15

bwateratmsft: So only a full refresh can clear the cache--no more partials?

alexweininger: Yes. And just to clarify, I haven't changed that behavior in this PR.

bwateratmsft (approved):

PR #651, opened 2023-03-31
Only run component detection on main branch

alexweininger:
I hope this will prevent CG alerts from appearing for branches other than main.


nturinski (approved):

bwateratmsft (approved):
I hope this works, but I suspect it will not actually. Component detection is injected automatically into all of our builds--and it skips if it's already been run.

PR #679, opened 2023-05-15
Focus feature v2

alexweininger:
Screen Recording 2023-05-15 at 2 55 49 PM

Note: I am doing some crude caching of the resource list operation. The cache is contained within the focus tree data provider and is reset when the refresh button is clicked. The cache is preserved through focus/unfocus commands so that the view updates quickly.

---

Client updates: Each client will need to add their commands to the `azureFocusView` view. Example: https://github.com/microsoft/vscode-azurecontainerapps/compare/main...alex/build-focus

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
     "name": "vscode-azureresourcegroups",
     "displayName": "Azure Resources",
     "description": "%azureResourceGroups.description%",
-    "version": "0.7.2-alpha.0",
+    "version": "0.7.4-alpha.0",
     "publisher": "ms-azuretools",
     "icon": "resources/resourceGroup.png",
     "aiKey": "0c6ae279ed8443289764825290e4f9e2-1a736e7c-1324-4338-be46-fc2a58ae4d14-7255",
--- a/package.json
+++ b/package.json
@@ -347,7 +347,7 @@
             "view/item/context": [
                 {
                     "command": "azureResourceGroups.focusGroup",
-                    "when": "view == azureResourceGroups && viewItem =~ /groupingItem/",
+                    "when": "view == azureResourceGroups && viewItem =~ /canFocus/",
                     "group": "inline"
                 },
                 {
--- a/package.json
+++ b/package.json
@@ -368,7 +368,7 @@
                 },
                 {
                     "command": "azureResourceGroups.deleteResourceGroupV2",
-                    "when": "view =~ /azureResourceGroups/ && viewItem =~ /azureResourceGroup/",
+                    "when": "view == azureResourceGroups && viewItem =~ /azureResourceGroup/",
                     "group": "1@1"
                 },
                 {
--- a/package.json
+++ b/package.json
@@ -65,6 +65,24 @@
         },
         "commands": [
             {
+                "command": "azureFavorites.refreshTree",
+                "title": "Refresh",
+                "category": "Azure",
+                "icon": "$(refresh)"
+            },
+            {
+                "command": "azureResourceGroups.addToFavorites",
+                "title": "Add to Favorites",
+                "category": "Azure",
+                "icon": "$(star-full)"
+            },
+            {
+                "command": "azureResourceGroups.removeFromFavorites",
+                "title": "Remove from Favorites",
+                "category": "Azure",
+                "icon": "$(close)"
+            },
+            {
                 "command": "azureResourceGroups.logIn",
                 "title": "%azureResourceGroups.logIn%",
                 "category": "Azure"
@@ -226,6 +244,12 @@
         "views": {
             "azure": [
                 {
+                    "id": "azureResourcesView;azureFavorites",
+                    "name": "Focused Resources",
+                    "visibility": "hidden",
+                    "when": "ms-azuretools.vscode-azureresourcegroups.hasFavorite == true"
+                },
+                {
                     "id": "azureResourceGroups",
                     "name": "Resources",
                     "visibility": "visible"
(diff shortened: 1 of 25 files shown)


src/tree/azure/CustomAzureResourceTreeDataProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+    azureExtensions
+        .map(e => e.resourceTypes)
+        .reduce((a, b) => a.concat(...b), []);
+
+export class CustomAzureResourceTreeDataProvider extends ResourceTreeDataProviderBase {
+    private subscriptionProvider: AzureSubscriptionProvider | undefined;
+    private filtersSubscription: vscode.Disposable | undefined;
+    private statusSubscription: vscode.Disposable | undefined;
+
+    constructor(
```
nturinski, line 32: I feel like a decent amount of this code is copied from 'AzureResourceTreeDataProvider`. We should be able to make a base class and extend it.

src/tree/azure/GroupingItem.ts
```diff
@@ -47,63 +50,103 @@ export class GroupingItem implements ResourceGroupsItem {
...
 
-        this.subscription = {
+        this.subscription = this.context ? {
             // for v1.5 compatibility
             ...this.context.subscriptionContext,
             ...this.context.subscription,
-        };
+        } : undefined;
+
+        if (this.resourceGroup) {
```
nturinski, line 59: Could you explain why this is the way it is?

src/tree/azure/GroupingItem.ts
```diff
@@ -47,63 +50,103 @@ export class GroupingItem implements ResourceGroupsItem {
...
+        const sortedResources = this.resources.sort((a, b) => {
+
+            if (a.subscription !== b.subscription) {
+                return a.subscription.name.localeCompare(b.subscription.name);
+            }
+
+            return a.name.localeCompare(b.name);
+        });
+
+        const groupMap = new Map<AzureSubscription, AzureResource[]>();
```
nturinski, line 86: Why do we need this groupMap now? Is it so we don't have repeated resource IDs in sorted resources?

package.json
```diff
@@ -265,6 +295,10 @@
                 "view": "azureWorkspace",
                 "contents": "No local workspace resource providers exist.",
                 "when": "azureWorkspace.state == 'noWorkspaceResourceProviders'"
+            },
+            {
+                "view": "azureFocusView",
+                "contents": "Select a resource to focus from the Resources view."
```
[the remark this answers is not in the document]
alexweininger: Welcome view shouldn't ever be shown anyway. I removed it.

package.json
```diff
@@ -225,6 +249,12 @@
         },
         "views": {
             "azure": [
+                {
+                    "id": "azureFocusView",
+                    "name": "Focused Resources",
+                    "visibility": "hidden",
+                    "when": "ms-azuretools.vscode-azureresourcegroups.hasFocusedGroup == true"
```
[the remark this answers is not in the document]
alexweininger: I actually came up with that myself 🤓

src/tree/azure/GroupingItem.ts
alexweininger: The purpose of this map is to add in the subscription separators (description-only tree items) if needed.

src/tree/azure/CustomAzureResourceTreeDataProvider.ts
```diff
@@ -0,0 +1,95 @@
...
+import { AzureResourceGroupingManager } from './AzureResourceGroupingManager';
+import { AzureResourceTreeDataProviderBase } from './AzureResourceTreeDataProviderBase';
+import { GroupingItem } from './GroupingItem';
+
+const supportedResourceTypes: AzExtResourceType[] =
+    azureExtensions
+        .map(e => e.resourceTypes)
+        .reduce((a, b) => a.concat(...b), []);
+
+export class CustomAzureResourceTreeDataProvider extends AzureResourceTreeDataProviderBase {
```
nturinski, line 26: nit: Are you going to change the name of this? It's pretty confusing as if what this is supposed to be for.

src/tree/azure/GroupingItem.ts
nturinski: I think it's worth writing a comment or changing the variable name to something more descriptive.

Like `subscriptionGroupingMap` or `resourcesGroupedBySubscriptionMap`

nturinski (dismissed):
Left some small comments, but overall looks good. I'll approve it since I don't want my being OOF to impact this getting in.

src/commands/focus/focusGroup.ts
```diff
@@ -0,0 +1,40 @@
...
+        ext.focusedGroup = {
+            kind: 'resourceType',
+            type: item.resourceType,
+        }
+    } else if (item.location) {
+        ext.focusedGroup = {
+            kind: 'location',
+            location: item.location,
+        }
+    }
```
bwateratmsft, line 28: No ARM tag group support?

alexweininger: For now I'm not adding ARM tag group support since it requires more refactoring than I'd like to do at this moment. The focus command will not appear on ARM tag groups.

src/tree/azure/GroupingItem.ts
alexweininger: `GroupingItem`s can now appear in both the focus view and the resources view. For the resources view, `id` property will behave the same as before.

However, in the focus view, these group items may not have a subscription and so I need to create an id without the subscription.

src/tree/azure/CustomAzureResourceTreeDataProvider.ts
alexweininger: Updated

nturinski (approved):

PR #684, opened 2023-05-17
web: Fix selected subscriptions setting value

### LINKED ISSUE #680: Tenant ID is missing from values in `azureResourceGroups.selectedSubscriptions`
These values should include the tenant ID. Blocks https://github.com/microsoft/vscode-azuretools/pull/1461.

alexweininger:
Important: I want to wait to merge this until after the focus feature is released.

Fixes #680

I had to add an automatic migration for old setting values. I'm hoping we can remove this migration logic in the future and replace it with some error handling.

I also added some detailed comments about what is happening when we list subscriptions for a specific tenant. I needed to make small changes in that code to ensure that subscriptions returned from the default tenant have their `tenantId` property set. For some reason, when subscriptions are fetched from the default tenant, the `tenantId` property is an empty string.

### DIFF
--- a/src/services/WebAzureSubscriptionProvider.ts
+++ b/src/services/WebAzureSubscriptionProvider.ts
@@ -6,7 +6,7 @@
 import * as arm from '@azure/arm-subscriptions';
 import { Environment } from '@azure/ms-rest-azure-env';
 import { uiUtils } from '@microsoft/vscode-azext-azureutils';
-import { callWithTelemetryAndErrorHandling, IActionContext } from '@microsoft/vscode-azext-utils';
+import { callWithTelemetryAndErrorHandling, IActionContext, nonNullValue } from '@microsoft/vscode-azext-utils';
 import * as vscode from 'vscode';
 import { AzureSubscription } from '../../api/src/index';
 import { AzureLoginStatus } from '../../azure-account.api';
@@ -94,6 +94,12 @@ class VSCodeAzureSubscriptionProvider extends vscode.Disposable implements Azure
 
         this.allSubscriptions = allSubscriptions;
 
+        // migrate setting values that are just a subscription id to `${tenantId}/${subscriptionId}`
+        const selectedSubscriptionIds = settingUtils.getGlobalSetting<string[] | undefined>('selectedSubscriptions');
+        if (selectedSubscriptionIds?.some(id => !id.includes('/'))) {
+            await this.updateSelectedSubscriptions(this.filters);
+        }
+
         return {
             status: session ? 'LoggedIn' : 'LoggedOut',
             allSubscriptions,
@@ -103,9 +109,11 @@ class VSCodeAzureSubscriptionProvider extends vscode.Disposable implements Azure
 
     get filters(): AzureSubscription[] {
         const selectedSubscriptionIds = settingUtils.getGlobalSetting<string[] | undefined>('selectedSubscriptions');
-        return this.allSubscriptions
-            .filter(s => selectedSubscriptionIds === undefined || selectedSubscriptionIds.includes(s.subscriptionId))
+        // values may be just subscription ids, or they may be in the (correct) form `${tenantId}/${subscriptionId}`
+        const subscriptions = this.allSubscriptions
+            .filter(s => selectedSubscriptionIds === undefined || selectedSubscriptionIds.includes(s.subscriptionId) || selectedSubscriptionIds.includes(`${s.tenantId}/${s.subscriptionId}`))
             .sort((a, b) => a.name.localeCompare(b.name));
+        return subscriptions;
     }
 
     async logIn(): Promise<void> {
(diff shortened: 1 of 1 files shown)


bwateratmsft (commented):
Is it only the web provider that uses that setting? Not on desktop?

alexweininger:
> Is it only the web provider that uses that setting? Not on desktop?

Yes, this setting is only used by the web provider. The Azure Account extension uses the `azure.resourceFilter` setting.

bwateratmsft:
> > Is it only the web provider that uses that setting? Not on desktop?
> 
> Yes, this setting is only used by the web provider. The Azure Account extension uses the `azure.resourceFilter` setting.

Gotcha. So going forward, using the new setting is the right move. We could potentially have a migration path from Azure Account -> new setting, but I don't think it's necessary. I might even argue it's better not to migrate that setting, and force users to go through it again--relearning that subscription filtering is an option, and one I want them to use more.

src/services/WebAzureSubscriptionProvider.ts
```diff
@@ -94,6 +94,12 @@ class VSCodeAzureSubscriptionProvider extends vscode.Disposable implements Azure
 
         this.allSubscriptions = allSubscriptions;
 
+        // migrate setting values that are just a subscription id to `${tenantId}/${subscriptionId}`
+        const selectedSubscriptionIds = settingUtils.getGlobalSetting<string[] | undefined>('selectedSubscriptions');
```
nturinski, line 98: Should we only do this in the constructor? `getSubscriptions` gets called every time `selectSubscriptions` does, which can be pretty often.

src/services/WebAzureSubscriptionProvider.ts
```diff
@@ -103,9 +109,11 @@ class VSCodeAzureSubscriptionProvider extends vscode.Disposable implements Azure
 
     get filters(): AzureSubscription[] {
         const selectedSubscriptionIds = settingUtils.getGlobalSetting<string[] | undefined>('selectedSubscriptions');
-        return this.allSubscriptions
-            .filter(s => selectedSubscriptionIds === undefined || selectedSubscriptionIds.includes(s.subscriptionId))
+        // values may be just subscription ids, or they may be in the (correct) form `${tenantId}/${subscriptionId}`
+        const subscriptions = this.allSubscriptions
```
nturinski, line 113: Shouldn't it be fixed by the time `filters` is returned? I don't know if we need to worry about the wrong formatting everywhere in code.

nturinski (commented):
Rather than having all this code to convert sub IDs that are malformed, I wonder if we would be better off just deleting anything that doesn't match the `tenant/subscription` format and making users select their subs again?

Since this is only a web issue, I don't think it's going to be that prevalent and the thought of having this workaround code in here for the indefinite future makes me bleh.

src/services/WebAzureSubscriptionProvider.ts
alexweininger: We need to `await` updating the setting value so I don't think it can be done in the constructor.

src/services/WebAzureSubscriptionProvider.ts
```diff
@@ -94,6 +94,12 @@ class VSCodeAzureSubscriptionProvider extends vscode.Disposable implements Azure
 
         this.allSubscriptions = allSubscriptions;
 
+        // clear setting value if there's a value that doesn't include the tenant id
```
bwateratmsft, line 97: Can we include a comment explaining why? e.g. linking to this PR or the bug

src/services/WebAzureSubscriptionProvider.ts
```diff
@@ -103,9 +109,10 @@ class VSCodeAzureSubscriptionProvider extends vscode.Disposable implements Azure
 
     get filters(): AzureSubscription[] {
         const selectedSubscriptionIds = settingUtils.getGlobalSetting<string[] | undefined>('selectedSubscriptions');
-        return this.allSubscriptions
-            .filter(s => selectedSubscriptionIds === undefined || selectedSubscriptionIds.includes(s.subscriptionId))
+        const subscriptions = this.allSubscriptions
+            .filter(s => selectedSubscriptionIds === undefined || selectedSubscriptionIds.includes(`${s.tenantId}/${s.subscriptionId}`))
```
bwateratmsft, line 113: I think we should treat `[]` as equivalent to `undefined`, i.e., nothing hidden by filter:
```suggestion
            .filter(s => selectedSubscriptionIds === undefined || selectedSubscriptionIds === [] || selectedSubscriptionIds.includes(`${s.tenantId}/${s.subscriptionId}`))
```

src/services/WebAzureSubscriptionProvider.ts
alexweininger: Ok, I added a `static async Create` method to the class that handles it.

bwateratmsft (approved):

PR #707, opened 2023-06-02
Use the new shared authentication provider package

bwateratmsft:
This ended up being more substantial than I anticipated, but I got all the automated tests passing, and a _very_ quick glance at features seems to be working fine.


src/commands/accounts/logOut.ts
bwateratmsft, line 1: `logOut` had to be removed. There is no command in the built-in auth provider for logging out--the user just has to do "Sign Out" from the account menu.

src/tree/azure/AzureResourceTreeDataProviderBase.ts
```diff
@@ -50,20 +48,32 @@ export abstract class AzureResourceTreeDataProviderBase extends ResourceTreeData
...
             return ext.testing.overrideAzureSubscriptionProvider();
         } else {
             if (!this.subscriptionProvider) {
                 this.subscriptionProvider = await ext.subscriptionProviderFactory();
-                await this.subscriptionProvider.waitForFilters();
             }
 
-            this.filtersSubscription = this.subscriptionProvider.onFiltersChanged(() => this.notifyTreeDataChanged());
-            this.statusSubscription = this.subscriptionProvider.onStatusChanged(() => this.notifyTreeDataChanged());
+            this.statusSubscription = vscode.authentication.onDidChangeSessions((evt: vscode.AuthenticationSessionsChangeEvent) => {
```
bwateratmsft, line 60: We could add an event to the shared package to mark sign in. I don't feel strongly about it being here.

alexweininger:
Just a note: I'm still having issues on sign-in when using an account that has subscriptions in a non-default tenant. I'll be working on fixing it next week.

src/commands/accounts/logOut.ts
```diff
@@ -1,12 +0,0 @@
-/*---------------------------------------------------------------------------------------------
-*  Copyright (c) Microsoft Corporation. All rights reserved.
-*  Licensed under the MIT License. See License.md in the project root for license information.
-*--------------------------------------------------------------------------------------------*/
-
-import { IActionContext } from '@microsoft/vscode-azext-utils';
-import { ext } from '../../extensionVariables';
-
-export async function logOut(_context: IActionContext): Promise<void> {
```
nturinski, line 9: Is the only way to logout through the VS Code accounts pane? It seems like it'd be nice to still support logging out via the command palette.

src/commands/accounts/selectSubscriptions.ts
```diff
@@ -4,9 +4,67 @@
...
+                    subscription
+                }))
+                .sort((a, b) => a.label.localeCompare(b.label));
+        }
+
+        const picks = await vscode.window.showQuickPick(
+            subscriptionQuickPickItems(),
+            {
+                canPickMany: true,
+                placeHolder: 'Select Subscriptions'
```
nturinski, line 35: nit: localize

nturinski (approved):

src/commands/accounts/logOut.ts
bwateratmsft: That's correct--logging out is not possible programmatically, so the command has to be removed.

nturinski:
The subscriptions don't automatically refresh after filtering them.

bwateratmsft:
> The subscriptions don't automatically refresh after filtering them.

Fixed!

alexweininger:
Being signed in with multiple Microsoft accounts at the same time seems to really throw the extension into a frenzy. It was having errors when I had 3 accounts. Signed out of 1, it loaded subs from one of the accounts. Then signed out of all of them and now it shows subs for a different account, but I'm completely signed out.

image

bwateratmsft:
That seems like it might be an issue with the auth provider...

bwateratmsft:
In fact, I wonder if it's this: https://github.com/microsoft/vscode/issues/179202

nturinski:
Might consider fixing this issue as well. https://github.com/microsoft/vscode-azureappservice/issues/2464

Whenever we do a request, we should probably verify with MS Auth provider whether or not the token we have is still valid.

bwateratmsft:
> Might consider fixing this issue as well. [microsoft/vscode-azureappservice#2464](https://github.com/microsoft/vscode-azureappservice/issues/2464)
> 
> Whenever we do a request, we should probably verify with MS Auth provider whether or not the token we have is still valid.

The authentication provider guarantees that it won't issue expired tokens and AFAIK we don't cache them anywhere (i.e., `getToken` returns the result from a call to `getSession`), so I don't think having expired creds is even possible.

nturinski:
It happens whenever you leave the VS Code for the Web window idle for like am hour or so. Your next request will have an expired token and you have to refresh the webpage to get the new one (but not required to sign in again). I don't think this should be unique to the web, it's just the only place we actually it's live right now though.

But that was an old implementation and maybe this current shared one wouldn't have the same issue.

bwateratmsft:
Looking closer, it looks like everything is substantially the same (in terms of what caching is happening) as the old implementation, except for where we have `subscriptionResultsTask` as a cached variable. But the implementation of the `TokenCredential` is still via `getSession`.

I don't think this will fix https://github.com/microsoft/vscode-azureappservice/issues/2464 but I also don't think it will be any worse.

package.json
```diff
@@ -74,11 +74,6 @@
                 "title": "%azureResourceGroups.logIn%",
                 "category": "Azure"
             },
-            {
-                "command": "azureResourceGroups.logOut",
```
nturinski, line 78: I think we should still expose these commands. Anyone familiar to using them via Azure Accounts may be wondering where they are. Otherwise the only entry points for logging out is through the accounts view.

nturinski: Never mind, forgot about the resolved command.

src/tree/azure/AzureResourceTreeDataProvider.ts
```diff
@@ -107,17 +110,6 @@ export class AzureResourceTreeDataProvider extends AzureResourceTreeDataProvider
                                 iconPath: new vscode.ThemeIcon('mortar-board')
                             }),
                     ];
-                } else {
```
nturinski, line 110: I liked having the loading~spin icon while the logging in was happening.

It can be confusing clicking that and it looks like nothing has changed. I've had times where I forgot that I had the "sign in with Azure" webpage opened, but that loading spinner reminded me to go to my browser.

src/tree/azure/AzureResourceTreeDataProviderBase.ts
```diff
@@ -50,20 +48,32 @@ export abstract class AzureResourceTreeDataProviderBase extends ResourceTreeData
...
+                if (evt.provider.id === 'microsoft' || evt.provider.id === 'microsoft-sovereign-cloud') {
+                    if (Date.now() > nextSessionChangeMessageMinimumTime) {
+                        nextSessionChangeMessageMinimumTime = Date.now() + sessionChangeMessageInterval;
+                        // This event gets HEAVILY spammed and needs to be debounced
+                        // Suppress additional messages for 1 second after the first one
+                        this.notifyTreeDataChanged();
+                    } else {
+                        console.log("Too soon, ignoring this event");
```
nturinski, line 68: Is this supposed to be here still?

bwateratmsft: Nope

src/tree/azure/AzureResourceTreeDataProvider.ts
bwateratmsft: Turns out to be _much_ more complicated...

Long story short, if you abandon a sign-in process without doing the browser part, `subscriptionProvider.isSignedIn()` will _block_. So the tree view will spin indefinitely since `getChildren()` cannot return until `isSignedIn()` unblocks--it's not possible to put a spinner _in_ the tree view.

nturinski (approved):
Figured I'd rather unblock Brandon and get CTI using this to start testing as well. Any comments/concerns, we can always address later since nothing was blocking anyway.

src/tree/azure/AzureResourceTreeDataProviderBase.ts
bwateratmsft: Fixed in #728

src/tree/azure/AzureResourceTreeDataProvider.ts
bwateratmsft: Fixed in #728

PR #718, opened 2023-06-15
Migrate to Track 2 SDK

alexweininger:


bwateratmsft (approved):

bwateratmsft:
Will we need to update all the core extensions to track-2 only before we update RGs to track-2 only?

src/utils/v2/credentialsUtils.ts
```diff
@@ -28,9 +27,6 @@ export function createCredential(getSession: (scopes?: string[]) => vscode.Provi
             } else {
                 return null;
             }
-        },
-        signRequest: async () => {
-            throw new Error((localize('signRequestError', 'Track 1 credentials are not (currently) supported.')));
```
alexweininger, line 33: @bwateratmsft 
> Will we need to update all the core extensions to track-2 only before we update RGs to track-2 only?

Since this whole time we were only stubbing this method to satisfy the interface I think it's fine. Anyt extensions relying upon Track 1 credentials should be throwing errors with v2. 

So my answer is no, we shouldn't have to update and release all clients to track 2 before we update RGs to track 2 only since it seems like Track 1 was already not being supported by RGs v2.

PR #735, opened 2023-08-09
Support icons next to create commands in quick pick

alexweininger:
Added it to Functions to try out:

image

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -140,6 +140,7 @@ commands?: {
     command: string;
     title: string;
     detail: string;
+    type?: string; // Optional: resource type associated with the command. Used to show an icon next to the command.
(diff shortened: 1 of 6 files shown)


nturinski (approved):
Should we have an `iconPath` property too so that client extensions can bring in their own icon as well? I guess since it's only for 'create azure resource' commands, we should theoretically have them already though.

alexweininger:
> Should we have an `iconPath` property too so that client extensions can bring in their own icon as well? I guess since it's only for 'create azure resource' commands, we should theoretically have them already though.

I thought about that but I think specifying icon paths across extensions could be weird. I haven't tried it though, so maybe it _just works_.

PR #742, opened 2023-08-24
Add `ArcEnabledServers` resource to `AzExtResourceType` enum 

andyleejordan:
We'll map it to an upcoming extension later, when it's ready.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -144,11 +144,11 @@ commands?: {
 }[];
 ```
 
-> [View defintion of `AzExtResourceType`](../src/AzExtResourceType.ts)
+> [View definition of `AzExtResourceType`](../src/AzExtResourceType.ts)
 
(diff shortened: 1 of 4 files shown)


andyleejordan:
The [documentation](https://learn.microsoft.com/en-us/azure/azure-arc/overview) on Azure Arc if we want to figure out a better name. Also /cc @maertendMSFT

andyleejordan:
Ok so "Azure Arc-enabled Servers" is the official name. Since "Azure" is implied by the type and this is just implementation, we're dropping that. We'll name the upcoming extension `vscode-azurearcenabledservers`. Not worried about localization or the `legacyTypeMap`.

alexweininger (approved):

alexweininger (approved):

PR #746, opened 2023-09-01
Sort workspace items by id

alexweininger:


src/tree/workspace/WorkspaceResourceTreeDataProvider.ts
```diff
@@ -47,7 +47,7 @@ export class WorkspaceResourceTreeDataProvider extends ResourceTreeDataProviderB
 
             // NOTE: Returning zero children indicates to VS Code that is should display a "welcome view".
             //       The one chosen for display depends on the context set above.
-            return children;
+            return children.sort((a, b) => a.id.localeCompare(b.id));
```
nturinski, line 50: I was wondering-- should we separate the workspace items by which extension contributed them? Alphabetical order may not make that much sense,

alexweininger: I was thinking that sorting by id would be most flexible. We can manage the ids as needed to sort it how we want.

nturinski: That's fair enough.

nturinski (approved):

PR #761, opened 2023-11-09
Add telemetry for subscription information

alexweininger:
The Azure Account extension is no longer used by the Azure Resources extension and the `updateSubscriptionsAndTenants` event it had is relied upon to generate Azure-subscription-related telemetry. We need to mimic the old event so that Kusto queries can be minimally changed.

### DIFF
--- a/src/tree/azure/AzureResourceTreeDataProvider.ts
+++ b/src/tree/azure/AzureResourceTreeDataProvider.ts
@@ -78,7 +78,7 @@ export class AzureResourceTreeDataProvider extends AzureResourceTreeDataProvider
                         }
                     )];
                 } else if (await subscriptionProvider.isSignedIn()) {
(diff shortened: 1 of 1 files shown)


src/tree/azure/AzureResourceTreeDataProvider.ts
```diff
@@ -135,4 +136,34 @@ export class AzureResourceTreeDataProvider extends AzureResourceTreeDataProvider
 
         return undefined;
     }
+
+    private hasSentSubscriptionTelemetry = false;
+    private async sendSubscriptionTelemetryIfNeeded(): Promise<void> {
```
bwateratmsft, line 141: A tiny performance improvement would be to make this method not-async

bwateratmsft (dismissed):

bwateratmsft (approved):

PR #825, opened 2024-03-01
Throw an error for any item that resolves as undefined

nturinski:
Mitigates https://github.com/microsoft/vscode-azurefunctions/issues/3851 and other issues, I'm sure

We see this error floating around quite often. While this doesn't actually fix it, at least the error makes a little more sense to the user.

### DIFF
--- a/package.json
+++ b/package.json
@@ -571,7 +571,7 @@
     },
     "scripts": {
         "vscode:prepublish": "npm run cleanReadme",
-        "build": "tsc && gulp webpack-prod",
+        "build": "tsc",
         "cleanReadme": "gulp cleanReadme",
         "compile": "tsc -watch",
         "package": "vsce package --githubBranch main",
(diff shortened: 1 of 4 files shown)


bwateratmsft (dismissed):

package.json
```diff
@@ -571,7 +571,7 @@
     },
     "scripts": {
         "vscode:prepublish": "npm run cleanReadme",
-        "build": "tsc && gulp webpack-prod",
+        "build": "tsc",
```
bwateratmsft, line 574: Is this intentionally included?

nturinski: Yeah, the `npm run build` shouldn't run webpack. It wasn't actually changing the line `return await extension.activate(ctx, perfStats, true);` either (where the true is ignoreBundle) so it wasn't actually prepping it to be used with webpack.

alexweininger: Can confirm that this matches our other extensions

bwateratmsft (dismissed):

package.json
nturinski: Yeah, agreed: https://github.com/microsoft/vscode-azureresourcegroups/issues/826

alexweininger (dismissed):

alexweininger (approved):

PR #836, opened 2024-04-09
Add maintainCloudShellConnection command

alexweininger:
* Update Node typings so we can use the built in fetch
  * Update chokidar types to solve a typing error caused by updating the node types

Added the maintain cloud shell connection command, which is only to be called by vscode.dev. And isn't shown in the palette.


src/commands/cloudShell.ts
```diff
@@ -0,0 +1,124 @@
...
+        } catch (e) {
+            ext.outputChannel.error(parseError(e).message);
+        }
+
+        await delay(1000 * 60 * 10 /* 10 minutes */);
+    }
+}
+
+export async function refreshDevTunnelAccessToken() {
+    const devTunnelsProdAppId = '46da2f7e-b5ef-422a-88d4-2a7f9de6a0b2';
```
nturinski, line 51: This isn't sensitive info?

src/commands/cloudShell.ts
```diff
@@ -0,0 +1,124 @@
...
+
+interface MaintainCloudShellConnectionOptions {
+    consoleUri: string;
+    terminalId: string;
+}
+
+export async function maintainCloudShellConnection(_context: IActionContext, options: MaintainCloudShellConnectionOptions) {
+    while (true) {
+        try {
+            await refreshDevTunnelAccessToken();
```
nturinski, line 29: If one of these fails, is it still a good idea to do the other two functions?

src/commands/cloudShell.ts
```diff
@@ -0,0 +1,124 @@
...
+            ext.outputChannel.error(parseError(e).message);
+        }
+
+        try {
+            await cloudShellSize(options.consoleUri, options.terminalId);
+        } catch (e) {
+            ext.outputChannel.error(parseError(e).message);
+        }
+
+        await delay(1000 * 60 * 10 /* 10 minutes */);
```
nturinski, line 46: Is it documented how long it takes for the connection to die?

alexweininger: No, the Cloud Shell engineers say "about every 10 minutes should work"

src/commands/cloudShell.ts
```diff
@@ -0,0 +1,124 @@
...
+    return new Promise(resolve => setTimeout(resolve, ms))
+}
+
+interface MaintainCloudShellConnectionOptions {
+    consoleUri: string;
+    terminalId: string;
+}
+
+export async function maintainCloudShellConnection(_context: IActionContext, options: MaintainCloudShellConnectionOptions) {
+    while (true) {
```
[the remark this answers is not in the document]
alexweininger: I also have an aversion to `while(true)`. I might change it if I get around to it.

src/commands/cloudShell.ts
alexweininger: I think so. Not all these operations are strictly required to keep it alive, so it's best to do what we can even if one fails. For example, refreshing the dev tunnels access token is only so that if the server restarts for some reason, it can reconnect to the same dev tunnel.

nturinski (approved):

PR #848, opened 2024-05-02
☁️ 🐚 Migrate Azure Cloud Shell feature

alexweininger:
Purpose: In order to deprecate Azure Account, we need move the Azure Cloud Shell feature to the Azure Resources extension.

I will be making updates to the code later but wanted to perform an initial migration without changing the code too much.

* Removed webworker webpack config. We no longer ship a web version of Azure Resources, so it should've been removed a while ago. This config is not compatible with the Azure Cloud Shell feature.
* My process was copying over the files and adding comments. I tried to keep my changes minimal.
* Replaced the reliance on the Azure Account auth with our subscription provider. If a user has multiple tenants, we will prompt the user to select one every time they launch cloud shell. I think we can improve this later, but it's fine for now.
* The feature relies on `node-fetch` and `request-promise`, in a future update I hope to replace both of these with built-in fetch.

This feature is fairly complex and does a lot of out of the ordinary stuff to make it work. Please ask questions. I'm tentatively planning to write a document in the repo explaining the workings of the feature.

edit: Also I purposely left two prominent bugs in the code from AA. I will fix those in follow up PRs. The fixes for those are in https://github.com/microsoft/vscode-azureresourcegroups/pull/851, so use that for testing.


package.json
```diff
@@ -626,13 +644,17 @@
     "dependencies": {
         "@azure/arm-resources": "^5.2.0",
         "@azure/arm-resources-profile-2020-09-01-hybrid": "^2.1.0",
+        "@azure/ms-rest-js": "^2.7.0",
         "@microsoft/vscode-azext-azureauth": "^2.3.0",
         "@microsoft/vscode-azext-azureutils": "^2.0.0",
         "@microsoft/vscode-azext-utils": "^2.4.0",
         "buffer": "^6.0.3",
         "jsonc-parser": "^2.2.1",
+        "node-fetch": "^3.3.2",
```
nturinski, line 653: Shouldn't this be native?

tsconfig.json
```diff
@@ -4,14 +4,16 @@
         "target": "es6",
         "outDir": "out",
         "lib": [
-            "es6"
+            "es6",
+            "DOM"
```
nturinski, line 8: 👀

webpack.config.js
```diff
@@ -44,18 +30,18 @@ const webConfig = dev.getDefaultWebpackConfig({
         '../build/default/validation': 'commonjs ../build/default/validation',
         '../build/Release/bufferutil': 'commonjs ../build/Release/bufferutil',
         '../build/default/bufferutil': 'commonjs ../build/default/bufferutil',
+
+        // required for cloud shell feature
+        bufferutil: 'commonjs bufferutil',
```
nturinski, line 35: This is the scary stuff 😬

nturinski (approved):
Don't see anything obviously wrong that stands out to me.

Only thing that I don't see is the equivalent of `uploadFile` being migrated over.
https://github.com/microsoft/vscode-azure-account/pull/958/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626L112

alexweininger:
> Only thing that I don't see is the equivalent of `uploadFile` being migrated over.
https://github.com/microsoft/vscode-azure-account/pull/958/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626L112

https://github.com/microsoft/vscode-azureresourcegroups/pull/852

tsconfig.json
alexweininger: I know. I'm not happy we have to add that but I hope to be able to remove it once I move this whole feature onto built-in fetch.

webpack.config.js
alexweininger: Honestly I don't know what this does

nturinski: lmao I'm pretty sure it's polyfilling. Probably has to do with cloud shell stuff running on its own node process.

src/cloudConsole/cloudShellChildProcess/cloudConsoleLauncher.ts
```diff
@@ -0,0 +1,217 @@
...
+    };
+}
+
+let resizeToken = {};
+async function resize(accessToken: string, terminalUri: string) {
+    const token = resizeToken = {};
+    await delay(300);
+
+    for (let i = 0; i < 10; i++) {
+        if (token !== resizeToken) {
```
MicroFish91, line 54: Q: Is it possible for `token` to not equal `resizeToken` here?

PR #852, opened 2024-05-07
Migrate "Upload to Cloud Shell" command

alexweininger:
This PR migrates the "Upload to Cloud Shell" command from Azure Account, to the Resources extension. The command allows users to upload files to Azure Cloud Shell. 

The entry points are:
1. File explorer context menus on files
2. Open file tab context menu
3. Command palette (file picker is shown)

The command requires users to have a cloud shell terminal open before running the command. To do this, we track open cloud shell terminals with a context value, and only show the command entry points when the value is greater than 0. I made sure to change the context key from what was used in Accounts to avoid any weirdness.


src/cloudConsole/uploadFileToCloudShell.ts
```diff
@@ -0,0 +1,57 @@
...
+    return undefined;
+}
+
+
+export async function uploadFileToCloudShell(_context: IActionContext, uri?: Uri) {
+    (async () => {
+        if (!workspace.isTrusted) {
+            throw new Error(localize('uploadingRequiresTrustedWorkspace', 'File upload only works in a trusted workspace.'));
+        }
+        let shell = shells[0];
```
MicroFish91, line 31: Does this mean that you can have multiple local shell instances but they all are connected to the same remote instance?  I notice if I upload a file, they appear in all cloud shell terminals when I list contents

MicroFish91 (approved):

src/cloudConsole/uploadFileToCloudShell.ts
alexweininger: Not exactly sure what you mean, but every user just has one cloud shell instance in Azure. So if you spin up multiple clients, the will all be connected to the same container.

nturinski (approved):

src/cloudConsole/uploadFileToCloudShell.ts
MicroFish91: Okay cool, you covered what I was trying to confirm, I probably wasn't referring to each entity using the correct name which could make it confusing to understand.

alexweininger:
The base changed. I'm going to redo this PR.

nturinski (approved):

PR #854, opened 2024-05-08
Use `process.execPath` instead of `process.argv0` to fix launching Cloud Shell on Linux

alexweininger:
This fix addresses https://github.com/microsoft/vscode-azure-account/issues/719

For some reason only on Linux (in my case Ubuntu desktop 22.04.2) `process.execPath` and `process.argv0` are different. We are using the `process.argv0` value as the shell args that start up the child process to control the custom cloud shell shell. And for linux, `/proc/self/exe` wasn't working, but when I change it to `process.execPath` it's happy.

I'd love for some info on this if anyone knows what's going on here.

on Linux:
Pasted Graphic

on macOS:
Pasted Graphic 1


MicroFish91 (approved):
No idea either, all I could find was this:
https://stackoverflow.com/questions/383973/is-args0-guaranteed-to-be-the-path-of-execution

nturinski (approved):

PR #856, opened 2024-05-09
Use new provider and add create/delete resource group tests

nturinski:
I'll come back and add some comments tomorrow explaining some weirdness/gaps in these tests.


extension.bundle.ts
```diff
@@ -15,6 +15,7 @@
 // The tests should import '../extension.bundle'. At design-time they live in tests/ and so will pick up this file (extension.bundle.ts).
 // At runtime the tests live in dist/tests and will therefore pick up the main webpack bundle at dist/extension.bundle.js.
 export * from '@microsoft/vscode-azext-utils';
+export * from '@microsoft/vscode-azext-azureutils';
```
nturinski, line 18: I'm pretty sure that you have to make sure that this export is below `azext-utils`. The reason being, `azureutils` relies on the UI extension variables being registered, so if you try to export this first, it breaks.

This is annoying because the `source.organizeImports` automatically brings it up to the top if you save with formatting on... 🤦

package.json
```diff
@@ -502,7 +502,8 @@
                             "%azureResourceGroups.deleteConfirmation.EnterName%",
                             "%azureResourceGroups.deleteConfirmation.ClickButton%"
                         ],
-                        "default": "EnterName"
+                        "default": "EnterName",
+                        "scope": "machine"
```
nturinski, line 506: I changed these scopes to be only machine/user scope. The reason being, you'd get some weird behavior if there was a workspace setting. The command that we have that changes the groupBy setting would set it on the global level however, so it was making it not actually do anything.

I think it makes more sense to have it on the user level since it impacts your Azure Resources view, which really shouldn't be workspace dependent anyway.

package.json
```diff
@@ -512,7 +513,8 @@
                     "azureResourceGroups.groupBy": {
                         "type": "string",
                         "description": "%azureResourceGroups.groupBy%",
-                        "default": "resourceType"
+                        "default": "resourceType",
+                        "scope": "machine"
```
nturinski, line 517: I changed these scopes to be only machine/user scope. The reason being, you'd get some weird behavior if there was a workspace setting. The command that we have that changes the groupBy setting would set it on the global level however, so it was making it not actually do anything.

I think it makes more sense to have it on the user level since it impacts your Azure Resources view, which really shouldn't be workspace dependent anyway.

src/extensionVariables.ts
```diff
@@ -47,10 +47,6 @@ export namespace ext {
 
     export let subscriptionProviderFactory: () => Promise<AzureSubscriptionProvider>;
 
-    // When debugging thru VS Code as a web environment, the UIKind is Desktop. However, if you sideload it into the browser, you must
-    // change this to UIKind.Web and then webpack it again
-    export const isWeb: boolean = env.uiKind === UIKind.Web;
```
nturinski, line 52: Shouldn't need this anymore since we don't deliver the web extension anymore.

test/nightly/crud.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    });
+
+    test('Delete Resource (EnterName)', async () => {
+        await settingUtils.updateGlobalSetting('deleteConfirmation', 'EnterName');
+        await runWithTestActionContext('Delete Resource', async context => {
+            await context.ui.runWithInputs([rgName, rgName], async () => {
+                await deleteResourceGroupV2(context);
+            });
+
+            // ext.azureResourceProvider.onDidChangeResource!(async e => {
```
nturinski, line 98: Because the `deleteResourceGroupV2` command just starts the delete and doesn't actually await it, we don't get any confirmation that the delete happened. That makes it difficult for me to check to see if it actually deleted.

It emits a tree state change, so does anyone know how to watch for that event?

extension.bundle.ts
nturinski: Since I'm only importing `LocationListStep` now, the export being on top doesn't break anything, it seems.

MicroFish91: Whoa, I get it, but that's super weird.  Is this the first time we've run into this issue in `extension.bundle.ts` in an extension??

test/nightly/crud.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        }));
+    });
+
+    test('Get Resources', async () => {
+        const subscriptionTreeItems = await ext.appResourceTree.getChildren();
+        assert.ok(subscriptionTreeItems.length > 0);
+        for (const subscription of subscriptionTreeItems) {
+            const groupTreeItems = await ext.appResourceTree.getChildren(subscription as AzExtParentTreeItem);
+            await Promise.all(groupTreeItems.map(async g => {
+                const children = await ext.appResourceTree.getChildren(g as AzExtParentTreeItem);
```
MicroFish91, line 68: Is it possible for an error to be thrown when calling `getChildren`?  Will it mess up the rest of the tests if an error is thrown here?

test/nightly/crud.test.ts
```diff
@@ -0,0 +1,127 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { type Location } from '@azure/arm-resources-subscriptions';
+import { createTestActionContext, runWithTestActionContext } from '@microsoft/vscode-azext-dev';
+import { AzExtParentTreeItem, createResourceClient, createResourceGroup, deleteResourceGroupV2, ext, IActionContext, LocationListStep, randomUtils, settingUtils, SubscriptionItem } from '../../extension.bundle';
+import { longRunningTestsEnabled } from "../global.test";
+import assert = require("assert");
```
MicroFish91, line 10:
```suggestion
import * as assert from 'assert';
```

src/services/VSCodeAzureSubscriptionProvider.ts
```diff
@@ -15,6 +16,13 @@ export function createVSCodeAzureSubscriptionProviderFactory(): () => Promise<VS
     }
 }
 
+export function createAzureDevOpsSubscriptionProviderFactory(): () => Promise<AzureDevOpsSubscriptionProvider> {
+    return async (): Promise<AzureDevOpsSubscriptionProvider> => {
+        azureDevOpsSubscriptionProvider ??= new AzureDevOpsSubscriptionProvider();
+        return azureDevOpsSubscriptionProvider;
+    }
+}
+
```
hossam-nasr, line 25: This shouldn't be necessary since it's already exported from the auth package, shouldn't duplicate work here

nturinski:
Closing in favor of https://github.com/microsoft/vscode-azureresourcegroups/commit/e0560951b1f46fd94e35ed33d34c6e51407ebb73

PR #857, opened 2024-05-11
Improve tenant selection for cloud shell

alexweininger:
The diff is misleading. There aren't that many changes here. I think because of the whitespace changes...

I'll do my best to explain what's going on here:

The Azure Account behavior was to get all authenticated tenants that have >0 subscriptions and let the user pick if there's more than one. The current behavior is to get all tenants and let the user pick if there's more than one. I'd like to make the experience mimic that of Azure Account.

The new behavior is the same as Azure Account, however a shortcut step 1. has been added that wasn't present before.

1. We list all tenants, if we find only one, we select it.
2. If they have multiple tenants, we check which are authenticated and have at least one subscription, if there's only one, we select it. (If there are zero, we still select the first tenant because later it'll show they need to setup their account if they have no subs)
3. If we get here, it means they have multiple authenticated tenants, we ask the user to select one.

In the future we might want to simplify this and just check if the tenant is authenticated and not if it has subscriptions. But for now I want to keep the experience unchanged from Azure Account to avoid any unnecessary UX changes for users.

### DIFF
--- a/src/cloudConsole/cloudConsole.ts
+++ b/src/cloudConsole/cloudConsole.ts
@@ -362,7 +362,8 @@ export function createCloudConsole(subscriptionProvider: AzureSubscriptionProvid
                 serverQueue.push({ type: 'log', args: [localize('foundTenants', `Found ${tenantsWithSubs.length} authenticated tenant${tenants.length > 1 ? 's' : ''}. Please use the "Sign in to directory..." command to sign in to additional tenants.`)] });
 
                 if (tenantsWithSubs.length <= 1) {
-                    selectedTenant = tenantsWithSubs[0];
+                    // If they have only one tenant with subscriptions, use it. If there's no tenant with subscriptions, use the first tenant.
+                    selectedTenant = tenantsWithSubs[0] ?? tenants[0];
                 } else {
                     // Multipe tenants with subscriptions, user must pick a tenant
                     serverQueue.push({ type: 'log', args: [localize('selectingTenant', `Selecting tenant...`)] });
--- a/src/cloudConsole/cloudConsole.ts
+++ b/src/cloudConsole/cloudConsole.ts
@@ -5,7 +5,7 @@
 
 import { TenantIdDescription } from '@azure/arm-resources-subscriptions';
 import { AzureSubscriptionProvider, getConfiguredAzureEnv } from '@microsoft/vscode-azext-azureauth';
-import { IActionContext, IAzureQuickPickItem, IParsedError, callWithTelemetryAndErrorHandlingSync, parseError } from '@microsoft/vscode-azext-utils';
+import { IActionContext, IAzureQuickPickItem, IParsedError, callWithTelemetryAndErrorHandlingSync, nonNullProp, parseError } from '@microsoft/vscode-azext-utils';
 import * as cp from 'child_process';
 import * as FormData from 'form-data';
 import { ReadStream } from 'fs';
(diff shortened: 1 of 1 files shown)


nturinski:
> The diff is misleading. There aren't that many changes here. I think because of the whitespace changes...

https://github.com/microsoft/vscode-azureresourcegroups/pull/857/files?diff=split&w=1

src/cloudConsole/cloudConsole.ts
```diff
@@ -343,36 +343,54 @@ export function createCloudConsole(subscriptionProvider: AzureSubscriptionProvid
...
+            } else {
+                // If the user has multiple tenants, then we check which tenants have subscriptions.
+                // This also checks if this tenant is authenticated.
+                // If a tenant is not authenticated, users will have to use the "Sign in to Directory..." command before launching cloud shell.
+                const tenantsIdsWithSubs = new Set<string>();
+                const subscriptions = await subscriptionProvider.getSubscriptions(false);
+                subscriptions.forEach((sub) => {
+                    tenantsIdsWithSubs.add(sub.tenantId);
                 });
+                const tenantsWithSubs = tenants.filter(tenant => tenantsIdsWithSubs.has(nonNullProp(tenant, 'tenantId')));
```
nturinski, line 361: Couldn't you just create an array with the set instead of filtering the tenants array?

`const tenantsWithSubs = Array.from(tenantsIdsWithSubs);`

src/cloudConsole/cloudConsole.ts
```diff
@@ -343,36 +343,54 @@ export function createCloudConsole(subscriptionProvider: AzureSubscriptionProvid
...
-                        description: defaultDomainName,
-                        data: tenant,
-                    };
-                }).sort((a, b) => a.label.localeCompare(b.label));
-
-                const pick = await window.showQuickPick<IAzureQuickPickItem<TenantIdDescription>>(picks, {
-                    placeHolder: localize('selectDirectoryPlaceholder', "Select tenant"),
-                    ignoreFocusOut: true // The terminal opens concurrently and can steal focus (https://github.com/microsoft/vscode-azure-account/issues/77).
+
+            if (tenants.length <= 1) {
```
nturinski, line 348: Is there any reason to have this conditional? It seems like you basically end up doing the same thing in `if (tenantsWithSubs.length <= 1)`

alexweininger: This is step 1 in the steps I outlined in the description.

I think it's very common to only have one tenant. Knowing that, I'd like to avoid making extra ARM calls wherever possible.

I could remove this shortcut case to simplify the code, but I think it's worth keeping since it speeds up the operation and avoids an ARM call for the common case.

alexweininger:
> > The diff is misleading. There aren't that many changes here. I think because of the whitespace changes...
> 
> https://github.com/microsoft/vscode-azureresourcegroups/pull/857/files?diff=split&w=1

This doesn't change it for me. I think it's because I added a new level of indentation because there's an extra if statement.

src/cloudConsole/cloudConsole.ts
nturinski: > I think it's very common to only have one tenant. Knowing that, I'd like to avoid making extra ARM calls wherever possible.

Yeah, I figured that one ARM call for one tenant wouldn't be that bad to make the code simpler, but your reasoning makes sense to me too.

nturinski (approved):
Weird, it makes the diff simpler for me. Basically anything that has to do with the `serverQueue` doesn't look changed at al..

PR #864, opened 2024-05-23
Enable long running tests and use new federated credentials for testing

hossam-nasr:
This PR uses the newly created `AzureDevOpsSubscriptionProvider` from the most recent release of the `auth` package, and creates some simple CRUD E2E tests. In addition, it also:

- Removes some references to the deprecated azure-account extension, and removes installing it as a pretest step
- Adds `webpack` as a pretest step
- Removes no-longer-needed references to web UI


.azure-pipelines/1esmain.yml
```diff
@@ -27,3 +27,5 @@ resources:
 # Use those templates
 extends:
   template: azure-pipelines/1esmain.yml@azExtTemplates
+  parameters:
+    useAzureFederatedCredentials: true
```
hossam-nasr, line 31: If you look at the rest of the file, this pipeline file is only run when scheduled for nightly builds. Normal build pipelines that occur on PRs are done through GitHub Actions and not through this file. This is why `useAzureFederatedCredentials` is always set to `true`. For completeness sake, we _could_ also set it to only be true if the build reason is `Schedule`, but that would make it harder to test manually for any reason

.vscode/extensions.json
```diff
@@ -1,6 +1,5 @@
 {
     "recommendations": [
         "dbaeumer.vscode-eslint",
-        "ms-vscode.azure-account"
```
hossam-nasr, line 4: no longer need this recommendation

gulpfile.ts
```diff
@@ -50,6 +50,5 @@ async function cleanReadme(): Promise<void> {
 
 exports['webpack-dev'] = gulp.series(prepareForWebpack, () => gulp_webpack('development'));
 exports['webpack-prod'] = gulp.series(prepareForWebpack, () => gulp_webpack('production'));
-exports.preTest = gulp_installAzureAccount;
```
hossam-nasr, line 53: We no longer need to install Azure Account before tests

test/global.test.ts
```diff
@@ -22,6 +22,20 @@ suiteSetup(async function (this: Mocha.Context): Promise<void> {
...
         context.ui = new TestUserInput(vscode);
     });
+    const groupBySetting = settingUtils.getWorkspaceSetting('groupBy')
+    userSettings.push({ key: 'groupBy', value: groupBySetting });
+
+    const deleteConfirmationSetting = settingUtils.getWorkspaceSetting('deleteConfirmation');
+    userSettings.push({ key: 'deleteConfirmation', value: deleteConfirmationSetting });
+
+    // TODO: Set to true for testing purposes
+    longRunningTestsEnabled = true; //!/^(false|0)?$/i.test(process.env.ENABLE_LONG_RUNNING_TESTS || '');
```
hossam-nasr, line 32: If this is only being used for tests, should this always be set to `true`?

test/runTest.ts
```diff
@@ -3,27 +3,15 @@
...
-        cp.spawnSync(
-            cli,
-            [
-                ...args,
-                '--install-extension', 'ms-vscode.azure-account',
-            ],
-            {
-                encoding: 'utf-8',
-                stdio: 'inherit'
-            });
```
hossam-nasr, line 24: No longer need to install Azure Account before running tests

src/extension.ts
```diff
@@ -69,7 +70,32 @@ export async function activate(context: vscode.ExtensionContext, perfStats: { lo
         activateContext.telemetry.properties.isActivationEvent = 'true';
         activateContext.telemetry.measurements.mainFileLoad = (perfStats.loadEndTime - perfStats.loadStartTime) / 1000;
 
-        ext.subscriptionProviderFactory = createVSCodeAzureSubscriptionProviderFactory();
+        // if this for a nightly test, we want to use the test subscription provider
+        const useAzureFederatedCredentials: boolean = !/^(false|0)?$/i.test(process.env['AzCode_UseAzureFederatedCredentials'] || '')
+        if (useAzureFederatedCredentials) {
+            // when running tests, ensure we throw the errors and they aren't silently swallowed
```
nturinski, line 76: Can you pull this out of the `extension.ts` code and make a new file for it? Just to help clean up our activate command.

hossam-nasr: Done 🫡

src/extension.ts
```diff
@@ -69,7 +69,12 @@ export async function activate(context: vscode.ExtensionContext, perfStats: { lo
         activateContext.telemetry.properties.isActivationEvent = 'true';
         activateContext.telemetry.measurements.mainFileLoad = (perfStats.loadEndTime - perfStats.loadStartTime) / 1000;
 
-        ext.subscriptionProviderFactory = createVSCodeAzureSubscriptionProviderFactory();
+        // if this for a nightly test, we want to use the test subscription provider
+        const useAzureFederatedCredentials: boolean = !/^(false|0)?$/i.test(process.env['AzCode_UseAzureFederatedCredentials'] || '')
```
nturinski, line 73: very very very small nit, but this could just go into `getSubscriptionProviderFactory`.

nturinski (approved):

src/extension.ts
hossam-nasr: I only kept it here because it's also used to set `activateContext.errorHandling.rethrow `. Should this be set in the `getSubscriptionProviderFactory` instead? Should I create the `useAzureFederatedCredentials` boolean in both?

nturinski: Ah, in that case, I think you could just pass the activateContext into `getSubscriptionProviderFactory` instead and set it in there?

hossam-nasr: Sure, done

nturinski (approved):

Issue #849, opened 2024-05-02
Tenants view


alexweininger:
A new view for viewing and managing tenants. It would be collapsed by default and below the Resources and Workspace views.

<img width="485" alt="Screenshot 2024-05-01 at 2 19 43 PM" src="https://github.com/microsoft/vscode-azureresourcegroups/assets/12476526/9d01b58b-2f3d-4651-8b4d-45d87635e69c">

1. Be able to see all tenants for the signed in account, and possible multiple accounts if we can support that
1. Select and deselect tenants. Only subscriptions from selected tenants should show up in the Resources view.
1. See when a tenant requires additional authentication and allow the user to start the auth process for those tenants

I also want to think about how this would work if other Azure (non v2 api) extensions want to leverage a similar view.

PR #869, opened 2024-06-03
Add Tenants view 

motm32:
Completes #849 

Here is how the view will look: 
image

**Detailed changes**

_Related to accounts:_ 
- This view will contain information related to accounts and tenants associated with those accounts. With this we also allow support users being signed into multiple accounts
- If the user is not signed into any azure account the Tenants view will include a "Sign in" button similar to the Resources view. 
- With Multi-Account support there is the possibility that duplicate subscriptions can show up if the same subscription belongs to multiple tenants. In this case we will add in the description of the subscription item in the resources view the associated tenant so users know which tenant they are making changes under. This will look like this 
image

_Related to tenants:_
- Checking and unchecking tenants within the view will update the subscriptions shown in the Resources view and the subscription filter. Along with this if the checked tenants or if no tenants match any subscriptions then by default all subscriptions will be shown. 
- If a tenant needs further authentication the tenant will default to being unchecked and a button will appear if a user hovers over the unauthenticated tenants which users can press to authenticate the tenant. Users can also right click on the unauthenticated tenant and click `Sign in to Tenant`. Finally if users check an unauthenticated tenant they will automatically be prompted to sign into that tenant. image

To do:

- [x] Changes to the `Sign out of Azure` code so the tenants view also gets refreshed
- [x] Integrate VSCode api changes to getsessions() once they are released
- [x] Add multi-account support


package.json
```diff
@@ -216,6 +216,24 @@
...
                 "category": "Azure"
+            },
+            {
+                "command": "azureTenant.refresh",
+                "title": "%azureResourceGroups.refresh%",
+                "category": "Azure",
+                "icon": "$(refresh)"
+            },
+            {
+                "command": "azureTenant.refreshTree",
```
bwateratmsft, line 227: What's the difference between this command and the other one?

src/commands/registerCommands.ts
```diff
@@ -60,6 +62,15 @@ export function registerCommands(): void {
         ext.actions.refreshFocusTree(node);
     });
 
+    registerCommand('azureTenant.refresh', async (context, node?: ResourceGroupsItem) => {
+        await handleAzExtTreeItemRefresh(context, node); // for compatibility with v1.5 client extensions
```
bwateratmsft, line 66: Are any of the v1.5 extensions going to be interacting with the new tenant tree?

src/tree/tenants/TenantResourceTreeDataProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    async onGetChildren(element?: ResourceGroupsItem | undefined): Promise<ResourceGroupsItem[] | null | undefined> {
+        if (element) {
+            return await element.getChildren();
+        } else {
+            const subscriptionProvider = await getAzureSubscriptionProvider(this);
+            const children: ResourceGroupsItem[] = await OnGetChildrenBase(subscriptionProvider);
+
+            if (children.length === 0) {
+                const session = await vscode.authentication.getSession('microsoft', ['https://management.azure.com/.default']); // Will not work for sovereign clouds
```
bwateratmsft, line 52: Is this a TODO?

motm32: Somewhat the VSCode team is making some changes to how getSession works so it's likely this line will look different. We will have to see once they release it on insiders hopefully in the next couple of weeks 😊

src/commands/registerCommands.ts
motm32: Not exactly sure if there will be any direct interactions but Alex did mention in the issue he created (https://github.com/microsoft/vscode-azureresourcegroups/issues/849) he wanted it compatible with the non v2 extensions which is why I kept this in.

package.json
motm32: I mostly took this formatting from the way workspace does refresh but I can probably remove one of them.

src/tree/tenants/TenantResourceTreeDataProvider.ts
alexweininger: See https://github.com/microsoft/vscode/issues/213947. Tyler is working on updating/adding APIs to make interacting with multiple accounts easier.

package.json
bwateratmsft: That makes sense, thanks!

alexweininger:
Note for later: lets update the people in this issue once this is merged - https://github.com/microsoft/vscode-azure-account/issues/901

motm32:
Also just a note for reviewing. This includes a lot of code at the moment that is from the auth tools package for testing purposes. I will remove that code once we are closer to being done.

vscode.proposed.authGetSessions.d.ts
alexweininger, line 1: Can you verify if we still need this? I think it has been released and included in stable API

src/commands/accounts/getSessionFromVSCode.ts
alexweininger, line 1: Can you mark which files were copied from azuretools? Did you need to modify any of them?

api/src/index.ts
alexweininger, line 1: I'd like to avoid making any changes to the resources api, even if just adding types.

Why do we need to add this to the API if we aren't expecting anyone else to contribute to the view?

motm32: I think I added it in the beginning when we weren't completely sure of the scope of this. I can remove this now since I don't think it will be needed.

vscode.proposed.authGetSessions.d.ts
motm32: Yup I'll check it out!

src/commands/accounts/getSessionFromVSCode.ts
motm32: We do need to modify in a couple of different places I'll point those out in another comment. I think I put a comment at some of the files from the azuretools package but I'll go through and make sure each of them have it.

alexweininger: Ok, I think overall the changes look good. So it might be time to get those changes into the auth package so I can see those in a PR too.

src/tree/azure/grouping/GroupingItem.ts
```diff
@@ -56,7 +56,7 @@ export class GroupingItem implements ResourceGroupsItem {
         } : undefined;
 
         if (this.context?.subscription) {
-            this.id = `/subscriptions/${this.context?.subscriptionContext.subscriptionId}/groupings/${this.label}`;
+            this.id = `/subscriptions/${this.context?.subscriptionContext.subscriptionId}/account/${this.context?.subscription.account?.id}/groupings/${this.label}`;
```
alexweininger, line 59: Have we tested reveal to make sure it still works with this change?

src/tree/tenants/TenantResourceTreeDataProvider.ts
alexweininger, line 1: A lot of the auth and tenant related helper functions in this file should be removed in favor of existing implementations.

src/tree/tenants/registerTenantTree.ts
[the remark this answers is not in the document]
motm32: I can remove it because I don't think I am using it but `createTreeView` requires an itemCache. I could maybe modify that or use something else but that makes that a little tricky to remove. 

The item state store should be easy to remove so I will go ahead and do that.

src/commands/accounts/selectSubscriptions.ts
```diff
@@ -19,6 +19,17 @@ export async function selectSubscriptions(context: IActionContext): Promise<void
 
             const allSubscriptions = await provider.getSubscriptions(false);
```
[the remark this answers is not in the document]
motm32: The intended behavior is that if no tenants are selected we would default to showing all subscriptions regardless of the tenant. At first I had it so that no subs would show but that seemed very weird to me. Not sure if you have a different opinion on that. 

I'll add a comment explaining that otherwise.

src/tree/azure/grouping/GroupingItem.ts
motm32: I have not tested it but I am pretty sure based on discussions with Nathan it would break things. I'll test it out and see.

alexweininger: Ok, we'll need to sort that out before merging. I can help with that too

motm32: Just tested it out by creating a storage account and clicking "click to view resource" in the activity log and nothing happened so seems to be broken. Before releasing this we definitely want to fix that.

src/tree/tenants/TenantResourceTreeDataProvider.ts
motm32: All the helper functions here are just changes from the tools package. I have moved them over to the auth package (see tagged auth PR).

src/tree/tenants/TenantResourceTreeDataProvider.ts
```diff
@@ -0,0 +1,199 @@
...
+            return await element.getChildren();
+        } else {
+            const subscriptionProvider = await getAzureSubscriptionProvider(this);
+            const children: ResourceGroupsItem[] = await OnGetChildrenBase(subscriptionProvider);
+
+            if (children.length === 0) {
+                const accounts = await vscode.authentication.getAccounts(getConfiguredAuthProviderId());
+                for (const account of accounts) {
+                    const session = await vscode.authentication.getSession(getConfiguredAuthProviderId(), getScopes(undefined), { account: account });
+                    const tenants = await this.getTenants(session);
```
alexweininger, line 60: See https://github.com/microsoft/vscode-azuretools/pull/1789/files#r1752557190

package.json
```diff
@@ -236,6 +236,24 @@
...
+                "title": "%azureResourceGroups.refresh%",
+                "category": "Azure",
+                "icon": "$(refresh)"
+            },
+            {
+                "command": "azureTenant.signInToTenant",
+                "title": "%azureTenant.signInToTenant%",
+                "category": "Azure",
+                "icon": "$(sign-in)"
```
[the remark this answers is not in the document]
motm32: They are referring to seperate commands. I just prefixed the same way that  "azureWorkspace.refreshTree" and "azureWorkspace.refresh" are done.

src/commands/accounts/selectSubscriptions.ts
```diff
@@ -73,3 +86,25 @@ export async function getSelectedTenantAndSubscriptionIds(): Promise<string[]> {
 async function setSelectedTenantAndSubscriptionIds(tenantAndSubscriptionIds: string[]): Promise<void> {
     await settingUtils.updateGlobalSetting('selectedSubscriptions', tenantAndSubscriptionIds);
 }
+
+// This function is also used to filter subscription tree items in AzureResourceTreeDataProvider
+export function getTenantFilteredSubscriptions(allSubscriptions: AzureSubscription[]): AzureSubscription[] | undefined {
+    const tenants = ext.context.globalState.get<string[]>('unselectedTenants');
```
alexweininger, line 92: Somewhere we should store this string as a constant.

motm32:
Hi @slavizh. Here is the custom vsix you can use to test Lighthouse! 
[ResourceGroupsArtifacts.zip](https://github.com/user-attachments/files/17996981/ResourceGroupsArtifacts.zip)


On VS Code you can open the command prompt (ctrl + shift + P) and use the command `Extensions: Install from VSIX...` to install this version of the Resource Groups extension. You may need to reload your window for the view to appear. If you run into any issues please file them on this repo and if you have any questions please let us know!

src/commands/accounts/selectSubscriptions.ts
```diff
@@ -31,14 +32,15 @@ export async function selectSubscriptions(context: IActionContext, options?: Sel
         let subscriptionsShownInPicker: string[] = [];
 
         const subscriptionQuickPickItems: () => Promise<IAzureQuickPickItem<AzureSubscription>[]> = async () => {
-
+            // If there are no tenants selected by default all subscriptions will be shown.
```
alexweininger, line 35: Not a blocker, but I want to discuss ways we could convey this behavior to users. When you only have a single sub it seems like checking/unchecking tenants does nothing (since the single sub will always be shown).

src/tree/tenants/TenantResourceBranchDataProviderManager.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { BranchDataProvider, ResourceModelBase } from '../../../api/src/index';
+import { ResourceBranchDataProviderManagerBase } from "../ResourceBranchDataProviderManagerBase";
+import { TenantResource } from './tenant';
+
+export class TenantResourceBranchDataProviderManager extends ResourceBranchDataProviderManagerBase<string, BranchDataProvider<TenantResource, ResourceModelBase>> {
+    constructor(
+        defaultProvider: BranchDataProvider<TenantResource, ResourceModelBase>
+    ) {
+        super(defaultProvider, () => { return 'tenant' });
```
alexweininger, line 14: Is this 'tenant' value used anywhere else? If so it should be a constant.

src/tree/OnGetChildrenBase.ts
```diff
@@ -0,0 +1,80 @@
...
+import { isLoggingIn } from "../commands/accounts/logIn";
+import { ext } from "../extensionVariables";
+import { localize } from "../utils/localize";
+import { GenericItem } from "./GenericItem";
+import { ResourceGroupsItem } from "./ResourceGroupsItem";
+import { AzureResourceTreeDataProvider } from "./azure/AzureResourceTreeDataProvider";
+import { AzureResourceTreeDataProviderBase } from "./azure/AzureResourceTreeDataProviderBase";
+import { TenantResourceTreeDataProvider } from "./tenants/TenantResourceTreeDataProvider";
+
+export async function OnGetChildrenBase(subscriptionProvider: AzureSubscriptionProvider, tdp?: AzureResourceTreeDataProvider): Promise<ResourceGroupsItem[]> {
```
alexweininger, line 17: Btw, I really like this shared approach.

src/tree/tenants/registerTenantTree.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    if (!value || !Array.isArray(value)) {
+        return [];
+    }
+
+    // remove any duplicates
+    return removeDuplicates(value);
+}
+
+export function isTenantFilteredOut(tenantId: string, accountId: string): boolean {
```
alexweininger, line 93: Non blocker, but we should plan to convert this to use a setting

src/tree/tenants/TenantTreeItem.ts
```diff
@@ -0,0 +1,43 @@
...
+import { ResourceGroupsItem } from '../ResourceGroupsItem';
+
+export interface TenantItemOptions extends GenericItemOptions {
+    readonly checkboxState?: vscode.TreeItemCheckboxState;
+}
+
+export class TenantTreeItem implements ResourceGroupsItem {
+    constructor(public readonly label: string, public tenantId: string, public readonly account: vscode.AuthenticationSessionAccountInformation, private readonly options?: TenantItemOptions) {
+    }
+
```
alexweininger, line 17: Nit: maybe this was a design choice, but you could pass the whole tenant object here instead of requesting the label and tenantId here. Then you can set tenantId and label based on the tenant object in the contructor.

src/tree/azure/AzureResourceTreeDataProvider.ts
```diff
@@ -65,75 +66,87 @@ export class AzureResourceTreeDataProvider extends AzureResourceTreeDataProvider
...
-                        ) {
-                            // Subscriptions might exist in an unauthenticated tenant. Show welcome view.
-                            await vscode.commands.executeCommand('setContext', 'azureResourceGroups.needsTenantAuth', true);
-                            return [];
-                        } else {
-                            return [new GenericItem(localize('noSubscriptions', 'Select Subscriptions...'), {
-                                commandId: 'azureResourceGroups.selectSubscriptions'
-                            })]
-                        }
+            if (children.length === 0) {
```
alexweininger, line 72: For code reading purposes, it might be nice to note here in what cases `OnGetChildrenBase` will return no children. Aka when the user is signed in.

src/tree/azure/grouping/GroupingItem.ts
alexweininger: Bump

src/tree/tenants/TenantResourceBranchDataProviderManager.ts
motm32: Seems like this is the only place this is used.

src/tree/azure/AzureResourceTreeDataProvider.ts
```diff
@@ -67,6 +67,7 @@ export class AzureResourceTreeDataProvider extends AzureResourceTreeDataProvider
             return await element.getChildren();
         } else {
             const subscriptionProvider = await getAzureSubscriptionProvider(this);
+            // When a user is signed in 'OnGetChildrenBase' will return the no children
```
alexweininger, line 70:
```suggestion
            // When a user is signed in 'OnGetChildrenBase' will return no children
```

alexweininger (dismissed):

slavizh:
@motm32  I did not spot any major issues but I would like to suggest some improvements:
* When you select subscription it takes some time to display. Seems to me if that experience is faster will be better.
*  When you click on F1 select subscriptions the subscriptions takes some time to load when you have a lot of subscriptions. It will be better if this can be faster.
* When I do F1 select tenant I have to sign in every time I choose a different tenant for the same account. Would be better if I can just sign in once and being able to switch tenants without having to sing in all over again.
* When I do F1 select tenant and after that F1 select subscriptions regardless of the tenant chosen I see the same subscriptions. Either select tenant does not work properly or somehow it is useless as you see all subscriptions from all tenants. I cannot verify the latter with my account as I do not have subscriptions that are seen only in one of the tenants my account logs.
* Specific to Lighthouse: The subscriptions I have access via directly on the tenant and via Lighthouse with the same account are visible only once when you do F1 select subscriptions. It is unclear to me from which access (RBAC or Lighthouse) I am viewing the subscriptions. I cannot verify as well which as from RBAC I have Owner and via Lighthouse I have contributor. May be this is related to the above issue if somehow select tenant is broken or all subscriptions are listed regarding of the tenant selected. I think it is useful to be able to see such subscriptions twice and being able to see which access you are using. For example may be there could be some grouping that shows available subscriptions per tenant in F1 select subscriptions.
* Via Lighthouse I have access to multiple subscriptions from multiple tenants. When doing F1 select subscriptions I can see all these subscriptions but would be nice if I can somehow group the subscriptions by their tenant owner so I can easily select all subscriptions that belong to certain tenant. There is similar experience in Azure portal for Lighthouse. May be even can be done via some command like Select Subscriptions by Tenant Owner or something where you first choose the tenant owner and after that the subscriptions. Would be nice if also allows selecting multiple tenant owners before being able to select the subscriptions. 

I hope this will be helpful.

slavizh:
@motm32 ignore the last reply. I did not saw the new pane so will come back with new feedback after testing that.

slavizh:
@motm32 My feedback after testing the accounts and tenant view:
* Confirming that doing things like select subscriptions, selecting/deselecting tenants is slow. I have a lot of subscriptions but at the same time it feels like like I have to wait more than 15 seconds for resources pane to load after selecting/deselecting tenant. Same for subscriptions selection. Sometimes even shows loading for minutes.
* For subscriptions for which I have access via Lighthouse and RBAC I can see them listed once when F1 select subscriptions. So I do not know which access I use to select them. From what I have tested the tenant where I am guest account and have direct access to the subscriptions via RBAC these subscriptions are missing, I can only see them via the tenant where the subscriptions are projected via Lighthouse. To verify that I have selected both tenants, selected those subscriptions, after deselecting the tenant where I have access to the subscriptions via RBAC the subscriptions disappear from Resources pane. While writing this and continue testing I noticed that sometimes also the same subscriptions appear no matter which tenant is selected. There is some inconstancy here which prevents identifying the correct behavior that is implemented.
* When I deselect a tenant do F1 select subscriptions I see the subscriptions of the deselected tenant. That I think should not happen. Especially because when the tenant is deselected the subscriptions will not appear in Resources pane even when selected.
* When you do F1 select subscriptions I think it is better to show the tenant name rather the tenant ID. In most cases the tenant ID does not meany anything to you but the tenant name/domain is something that is recognized.
* I am seeing a lot of issues with selecting/deselecting tenants I deselect tenant in one account and in Resources pane the subscriptions of other account and tenant appear along the subscription of the deselected tenant stays. It is very strange.


This feedback that was from previous reply stays:
* Specific to Lighthouse: The subscriptions I have access via directly on the tenant and via Lighthouse with the same account are visible only once when you do F1 select subscriptions. It is unclear to me from which access (RBAC or Lighthouse) I am viewing the subscriptions. I cannot verify as well which as from RBAC I have Owner and via Lighthouse I have contributor. May be this is related to the above issue if somehow select tenant is broken or all subscriptions are listed regarding of the tenant selected. I think it is useful to be able to see such subscriptions twice and being able to see which access you are using. For example may be there could be some grouping that shows available subscriptions per tenant in F1 select subscriptions.
* Via Lighthouse I have access to multiple subscriptions from multiple tenants. When doing F1 select subscriptions I can see all these subscriptions but would be nice if I can somehow group the subscriptions by their tenant owner so I can easily select all subscriptions that belong to certain tenant. There is similar experience in Azure portal for Lighthouse. May be even can be done via some command like Select Subscriptions by Tenant Owner or something where you first choose the tenant owner and after that the subscriptions. Would be nice if also allows selecting multiple tenant owners before being able to select the subscriptions.


FYI I have over 700 subscriptions.

I hope this helps.

motm32:
@slavizh Thank you so much for your feedback it is very helpful 😊

motm32:
If possible @slavizh could you provide a number of how many tenants you had in the view? For instance were the 700 subscriptions you mentioned were belonging to one tenant or spread across multiple?

slavizh:
@motm32 the subscriptions belong to multiple tenants because they are project via Lighthouse but I think in this case you are asking mostly if the subscriptions are seen from multiple tenants or from single one. In my case I have tenant where I have over 700 subscriptions, another tenant with 8 subscriptions. Those 8 are also seen on the other tenant via Lighthouse. These two tenants are under the same account. I also have my personal account where I have 2 subscriptions under one tenant. Over my personal account I also have 3 other tenants visible but there are no subscriptions for those.

PR #883, opened 2024-06-28
Use a promise.allSettle instead of a for loop that resolves one at a time

nturinski:
We talked about this in stand up and I wanted to see if the performance boost was significant.  Not surprisingly, it's an exponential performance boost, as in, the more resources you have, the better the improvement.

Functions has a weird bug right now that is hitting performance harder that I've fixed and will be putting up a PR for. It's making it take even longer, but it illustrates how much faster the promises are compared to the loop.

EDIT: left is the changes, right is the current release.

resources promiseAllSettleShort

### DIFF
--- a/src/tree/azure/grouping/GroupingItem.ts
+++ b/src/tree/azure/grouping/GroupingItem.ts
@@ -5,6 +5,7 @@
 
 import { createContextValue, ISubscriptionContext, TreeItemIconPath } from '@microsoft/vscode-azext-utils';
 import * as vscode from 'vscode';
+import { localize } from 'vscode-nls';
 import { AzExtResourceType, AzureResource, AzureResourceBranchDataProvider, AzureResourceModel, AzureSubscription } from '../../../../api/src/index';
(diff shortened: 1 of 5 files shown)


src/tree/azure/grouping/GroupingItem.ts
```diff
@@ -125,7 +125,8 @@ export class GroupingItem implements ResourceGroupsItem {
                     } catch (e) {
```
alexweininger, line 125: Doesn't have to be this PR but:
* We should log this error.
* And make sure it gets reported to telemetry.

alexweininger (dismissed):

src/tree/azure/grouping/GroupingItem.ts
nturinski: Oddly, this command doesn't pass in a context. It's a slightly bigger change to start pushing in a context everywhere for error telemetry, so I'll just log to output channel for now.

alexweininger (dismissed):

alexweininger (approved):

PR #887, opened 2024-07-25
Use cache to reveal resource

nturinski:
There was already an item cache being used for storing the relationship data between parent and child. I was able to just leverage that to make reveal a lot faster if something has already been loaded before.

The next step to improving performance is to the do following across extensions:
1. Make resources resolve faster by not requiring the Azure API call until _after_ it is expanded
2. Change all extensions that resolve the resources with multi
ple `GET` calls use the `LIST` call. There is some concern about the caching that messes things up that we should consider, however, such as deleting too many resources at once and the cached list doesn't update to reflect that.

reveal_cache


alexweininger:
Do we need to reevaluate the need for this now that we've fixed the resolver caching?

nturinski:
Hmm, I don't really see the harm in merging this considering the itemCache already existed. Might as well leverage it if we had already loaded the child once before.

Other than using the cache, I don't think that this changes the behavior of anything else.

PR #890, opened 2024-08-02
Fix tree item descriptions

alexweininger:
Descriptions on subscription items weren't showing up for Megan and I found out why.


motm32 (approved):

bwateratmsft (approved):

alexweininger:
Note: this isn't impacting users since until now we never had descriptions on items owned by RGs.

PR #926, opened 2024-09-17
Add shortcut to switching to sovereign cloud within the tenant view 

motm32:
This allows users to quickly change their sovereign cloud setting without having to go into the user settings. 

The shortcut lives here under the gear icon: 
image

There is also a command palette command titiled: `Azure: Configure Sovereign Cloud`

This brings up the below quick pick. The names correspond with the names on the setting.  When an item is selected the setting `microsoft-sovereign-cloud.environment` is changed to the corresponding value.

(I included the generic azure one so user could switch back easily)

image

I tested this by switching between an Azure China account and a default Azure account and had no issues.


alexweininger:
does this bring up a quick pick or something

motm32:
> Can you put a screenshot of the quick pick please

Yup I'll edit the description to add that

alexweininger:
just some quick feedback: Maybe we can put (Default) in the description of the "Azure" normal option?

motm32:
> just some quick feedback: Maybe we can put (Default) in the description of the "Azure" normal option?

Yup I'll do that!

alexweininger:
Also I think we'll merge this after we merge the tenant view PR. There might be other small features like this on top of the tenant view that we want to add.

motm32:
> Also I think we'll merge this after we merge the tenant view PR. There might be other small features like this on top of the tenant view that we want to add.

Sounds good I'll change the branch to main before we merge this.

src/commands/sovereignCloud/SovereignCloudListStep.ts
```diff
@@ -0,0 +1,24 @@
...
+import { AzureWizardPromptStep, IAzureQuickPickItem } from "@microsoft/vscode-azext-utils";
+import { ConfigureSovereignCloudContext } from "./ConfigureSovereignCloudContext";
+
+export class SovereignCloudListStep extends AzureWizardPromptStep<ConfigureSovereignCloudContext> {
+    public async prompt(context: ConfigureSovereignCloudContext): Promise<void> {
+        const picks: IAzureQuickPickItem<string>[] = [
+            { label: 'Azure (Default)', data: '' },
+            { label: 'Azure China', data: 'ChinaCloud' },
+            { label: 'Azure US Government', data: 'USGovernment' },
+            { label: 'A custom Microsoft Sovereign Cloud', data: 'custom' }
```
alexweininger, line 15: If they pick custom, should we take them to the setting page so they can actually configure it? We can't make this change later though.

motm32: Yah we should change the custom since it will cause an error at the moment

alexweininger (approved):

alexweininger (approved):

alexweininger (approved):

PR #934, opened 2024-10-03
Add feature to scope select subscriptions prompt by tenant

alexweininger:
This adds the ability to scope the select subscriptions prompt by tenant. It also adds the option to filter by account, but that's not implemented in this PR.

If a prompt is scoped to a tenant, it won't change any selected subscriptions that aren't in that tenant. Only subscriptions shown in the picker will be removed or added to the selected subscriptions setting.

### DIFF
--- a/src/commands/accounts/selectSubscriptions.ts
+++ b/src/commands/accounts/selectSubscriptions.ts
@@ -12,7 +12,7 @@ import { settingUtils } from "../../utils/settingUtils";
 
 interface SelectSubscriptionOptions {
     /**
-     * If provided, only subscriptions in this tenant will be shown in the picker. Only subscriptions shown in the picker will be rempved or added to the selected subscriptions setting.
+     * If provided, only subscriptions in this tenant will be shown in the picker. Only subscriptions shown in the picker will be removed or added to the selected subscriptions setting.
      */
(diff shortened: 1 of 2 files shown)


bwateratmsft (approved):

motm32 (approved):

alexweininger:
A further improvement to the select subscription picker would be to group the picks by tenant/account

alexweininger:
> A further improvement to the select subscription picker would be to group the picks by tenant/account

Foiled by this 
image

bwateratmsft (dismissed):

bwateratmsft (approved):

nturinski:
This error is being thrown by us. I think there's no actual limitation with VS Code's quick picker that prevents us from doing it. As far as I know, the grouping is just us sorting everything by group and using the `QuickPickItemKind.Separator` in between each group.

MRayermannMSFT (approved):

alexweininger:
> This error is being thrown by us. I think there's no actual limitation with VS Code's quick picker that prevents us from doing it. As far as I know, the grouping is just us sorting everything by group and using the `QuickPickItemKind.Separator` in between each group.

Yeah I was tempted to @ you fix it

PR #940, opened 2024-10-11
Add walkthrough for /azure

alexweininger:
Only enabled in vscode for the web


package.json
```diff
@@ -560,6 +560,31 @@
...
+                            "svg": "resources/vscode-dev-walkthrough/azdUp.svg",
+                            "altText": "Right click azure.yaml to deploy"
+                        }
+                    },
+                    {
+                        "id": "environment",
+                        "title": "Get to know your environment",
+                        "description": "The environment includes:\nAzure Developer CLI - Run, debug, and deploy Azure applications directly within the browser. [Learn more](https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/overview)\nGitHub Copilot - Access GitHub Copilot and manage your projects through the AI Assistant.\n[Open GitHub Copilot](command:workbench.panel.chat.view.copilot.focus)\nTip: You can use [GitHub Copilot for Azure](https://microsoft.qualtrics.com/jfe/form/SV_4OCHtDpnNfB1Meq) (``@azure``) to view and manage your Azure resources from within GitHub Copilot.\nVS Code Azure Extensions - The environment is fully equipped with Azure VS Code extensions. These extensions make Azure development easier to create, run and deploy your applications.\n[View Azure Extensions](command:workbench.view.extension.azure)",
```
jinglouMSFT, line 581: The URL for GHCP for Azure likely will need to be changed to point to the Marketplace when we launch vscode.dev/azure.  But good for now.

jinglouMSFT (approved):

package.json
alexweininger: Ok, I didn't think of that. I'll make it a aka.ms link so we can change it later easily.

PR #942, opened 2024-10-17
Add a toast for GHCP4A

### LINKED ISSUE #941: Toast notification for GHCP4A
Need to release on 10/29, shortly after GHCP4A preview. 

When to toast?
1. When Azure Tool and VS Code Copilot extensions are installed, but GHCP4A is not installed
2. When individual Azure extension we display resources in the resource tree is installed and VS Code Copilot extension is installed, but GHCP4A is not installed. This includes Logic Apps extension, IoT extension, Web PubSub, etc. 

What's out of scope?
1. Resources that are not shown in the resource group view.  eg. the AKS extension

Experiences:
When VS Code is launched, the toast notification should be shown. Likely done on extension activation.
Need telemetry on the button clicks

UI and strings:
Image

MicroFish91:
Closes #941 

Example [demo](https://microsoft-my.sharepoint.com/:v:/p/matfisher/EXM8bPc2wvFFqAV-dhJqPJ0BrVvdtOtBtFgicd9vUdzEEA?nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJPbmVEcml2ZUZvckJ1c2luZXNzIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXciLCJyZWZlcnJhbFZpZXciOiJNeUZpbGVzTGlua0NvcHkifX0&e=Gmihax) of implementation using my own a11y-kit extension for now


src/gitHubCopilotForAzure.ts
```diff
@@ -0,0 +1,69 @@
...
+const ghcpChatExtensionId = 'github.copilot-chat';
+const ghcp4aExtensionId = 'ms-azuretools.vscode-azure-github-copilot';
+const ghcp4aLearnPage = 'https://aka.ms/GetGitHubCopilotForAzure';
+const dontShowKey = 'ghcp4a/dontShow';
+
+/**
+ * Conditionally show an install toast for the GitHub Copilot for Azure extension
+ */
+export function gitHubCopilotForAzureToast({ globalState }: ExtensionContext): void {
+    void callWithTelemetryAndErrorHandling('ghcp4aToast', async (context: IActionContext) => {
```
bwateratmsft, line 20: Since this telemetry event is getting sent for all users on every activation, we should do `isActivationEvent = 'true'`

src/gitHubCopilotForAzure.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        const dontShow: boolean = globalState.get<boolean>(dontShowKey, false);
+        if (dontShow) {
+            return;
+        }
+
+        const install = {
+            title: localize('install', 'Install'),
+            run: async () => {
+                context.telemetry.properties.install = 'true';
```
bwateratmsft, line 34: Instead of three telemetry properties (`install`, `learnMore`, and `dontShowAgain`), I'd make this one property with three possible results.

src/gitHubCopilotForAzure.ts
MicroFish91: For this: `Otherwise we should suppress it anytime the notification is not shown`

Should I test for early return before calling `callWithTelemetryAndErrorHandling` then, or is the way I updated it now fine?

bwateratmsft (approved):

PR #975, opened 2024-12-06
Fix the subscription filter to match the checkboxes in the tenant view 

motm32:
When a tenant is unchecked those subscriptions don't appear in the resources view so they also should not appear in the subscription filter.


alexweininger:
Can you make sure the behavior from #977 is preserved?

alexweininger (approved):

PR #979, opened 2024-12-06
Add Tenant view telemetry and sign in button on tenant view 

motm32:
Here is how the button looks: 
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,12 @@
                 "category": "Azure"
             },
             {
+                "command": "azureResourceGroups.logInButton",
+                "title": "%azureResourceGroups.logInButton%",
+                "category": "Azure",
+                "icon": "$(sign-in)"
+            },
+            {
                 "command": "azureResourceGroups.selectSubscriptions",
                 "title": "%azureResourceGroups.selectSubscriptions%",
                 "category": "Azure",
@@ -392,6 +398,11 @@
                     "submenu": "azureWorkspaceCreate",
                     "when": "view == azureWorkspace",
                     "group": "navigation@1"
+                },
+                {
+                    "command": "azureResourceGroups.logInButton",
+                    "when": "view == azureTenantsView",
+                    "group": "navigation@1"
                 }
             ],
             "view/item/context": [
@@ -510,6 +521,10 @@
                 },
                 {
                     "command": "azureResourceGroups.installExtension",
+                    "when": "never"
+                },
+                {
+                    "command": "azureResourceGroups.logInButton",
                     "when": "never"
                 }
             ],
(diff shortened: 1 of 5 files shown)


alexweininger:
We can merge, but I think the button icon needs discussion. I don't think that icon is clear, I might prefer using no icon and just using text that says "Add Account".

package.json
```diff
@@ -105,6 +105,12 @@
                 "title": "%azureResourceGroups.logIn%",
                 "category": "Azure"
             },
+            {
+                "command": "azureResourceGroups.logInButton",
```
alexweininger, line 109: Should we prefix this command like: `azureTenantsView.logInButton`?

src/tree/tenants/TenantResourceTreeDataProvider.ts
```diff
@@ -43,36 +43,39 @@ export class TenantResourceTreeDataProvider extends ResourceTreeDataProviderBase
     }
 
     async onGetChildren(element?: ResourceGroupsItem | undefined): Promise<ResourceGroupsItem[] | null | undefined> {
-        if (element) {
-            return await element.getChildren();
-        } else {
-            const subscriptionProvider = await getAzureSubscriptionProvider(this);
-            const children: ResourceGroupsItem[] = await OnGetChildrenBase(subscriptionProvider);
+        return await callWithTelemetryAndErrorHandling('tenantResourceTreeDataProvider.onGetChildren', async (context: IActionContext) => {
```
alexweininger, line 46: Nathan might have input here, but do we want to expose our implementation details in telemetry?

A better name might be `azureTenantsView.getChildren`?

src/tree/tenants/registerTenantTree.ts
```diff
@@ -56,8 +56,10 @@ async function updateTenantsSetting(_context: IActionContext, tenants: vscode.Tr
 
     for (const [tenantTreeItem, state] of tenants.items) {
         if (state === vscode.TreeItemCheckboxState.Unchecked) {
+            _context.telemetry.properties.uncheckedTenant = 'true';
```
alexweininger, line 59: We use this now so remove the underscore.

```suggestion
            context.telemetry.properties.uncheckedTenant = 'true';
```

package.json
```diff
@@ -106,10 +106,10 @@
                 "category": "Azure"
             },
             {
-                "command": "azureResourceGroups.logInButton",
-                "title": "%azureResourceGroups.logInButton%",
+                "command": "azureTenantsView.logInButton",
```
alexweininger, line 109: Should the title now be "addAccount"?

alexweininger (approved):

PR #985, opened 2024-12-10
Sort children in tenants view

alexweininger:


nturinski (approved):
This is fine, but should we have a `sortChildrenImpl` that we just have on the providers?

alexweininger:
> This is fine, but should we have a `sortChildrenImpl` that we just have on the providers?

I think it's easier to just handle it case by case. I realize that it duplicates some code, however it allows us total freedom and keeps the implementation super simple.

PR #1014, opened 2025-01-10
Add Durable Task Scheduler resource type

philliphoff:
Adds a resource type for the new Durable Task Scheduler resource.


alexweininger:
Is this in preparation for changes to the Functions extension?

philliphoff:
Yes, it'd be to allow the Functions extension to extend DTS resources shown in the Azure tree (which currently show up in the "Unknown" group).  I assume it's still the case that we can't add support without this first step?

alexweininger:
> Is there a reasonable way that I can develop/test the Functions extension support without having this committed first? I seem to vaguely recall a way that I can run multiple in-development extensions together in the same test instance, but it's been a long minute.

Two ways.

First is to package Azure Resources into a vsix and install it then debug Functions.

Second is to use this Launch config which launches both Azure Resources and Azure Functions. But you'll only be able to set breakpoints in the currently opened extension (in this case Functions) https://github.com/microsoft/vscode-azurefunctions/blob/fe58f7faca18e3e52e4bef833db2a9d068dd3746/.vscode/launch.json#L22-L39

philliphoff:
> Second is to use this Launch config which launches both Azure Resources and Azure Functions. But you'll only be able to set breakpoints in the currently opened extension (in this case Functions) https://github.com/microsoft/vscode-azurefunctions/blob/fe58f7faca18e3e52e4bef833db2a9d068dd3746/.vscode/launch.json#L22-L39

@alexweininger That's super useful, thanks; I've used that method to prepare proposed changes to the Functions extension.  I've switched this to a draft PR for now; I'll switch it back when we're ready for both to go in.

philliphoff:
@alexweininger We're ready for the Functions extensions changes to go it, so it'd be great to get this committed whenever convenient.

alexweininger (approved):

alexweininger:
We're releasing tomorrow, do you want these changes included?

philliphoff:
@alexweininger If possible, that'd be great.  If not, when's the next window?

alexweininger:
@philliphoff I can get it in today

PR #1023, opened 2025-01-22
Update version and changelog for v0.10.3

alexweininger:


alexweininger:
Why is the license check hanging 😢

PR #1030, opened 2025-01-29
Add API for getting activities

MRayermannMSFT:


bwateratmsft:
Should this be implemented as an LM tool instead?

MRayermannMSFT:
> Should this be implemented as an LM tool instead?

Long term, yep, this should probably end up as a tool. Also I can close this PR for now. Was just up for our `@azure` hackathon week. :)

PR #1034, opened 2025-01-31
Implement stand-in chat provider for `@azure`

### LINKED ISSUE #1033: Add a faux `@azure` chat participant
This extension should add a faux `@azure` chat participant whose job is to enable discovery of [GitHub Copilot for Azure](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azure-github-copilot).

This faux chat participant should:
- Have `disambiguations` which are copied from/the same as `@azure`'s.
- Be conditionally contributed based on a configuration value which enables/disables the faux chat participant.
- Have an implementation which:
    1. If GitHub Copilot for Azure is not installed then the faux chat participant implementation should: 
       - Tell the user that to use `@azure` for help with Azure questions in Copilot Chat, they need to install GitHub Copilot for Azure.
       - Offer a button which will:
          - Install/open extension page for GitHub Copilot for Azure.
          - Disable the faux chat participant.
       - Offer a link (AKA less prominent UX) to disable the faux participant without installing GitHub Copilot for Azure.
    2. If GitHub Copilot for Azure is installed then the faux chat participant implementation should:
       - Set the configuration to disable the faux chat participant.
       - Tell the user they need to restart VS Code.

Within the GitHub Copilot for Azure extension, it may want/need to try and set the configuration value to disable the faux participant.

bwateratmsft:
Fixes #1033


package.json
```diff
@@ -36,6 +36,70 @@
     ],
     "main": "./main.js",
     "contributes": {
+        "chatParticipants": [
+            {
+                "id": "ms-azuretools.azure-agent",
```
bwateratmsft, line 41: Can the ID be different?

package.json
```diff
@@ -36,6 +36,71 @@
...
+                "name": "azure",
+                "fullName": "%chatParticipants.azure.fullName%",
+                "description": "GitHub Copilot for Azure",
+                "isSticky": false,
```
bwateratmsft, line 45: I intentionally did not make it sticky, in case that messes up the switch from standin `@azure` to real `@azure`.

src/chat/chatStandIn.ts
```diff
@@ -0,0 +1,50 @@
...
+        'ms-azuretools.azure-agent',
+        (cr, cc, rs, t) => callWithTelemetryAndErrorHandling('azureResources.chatStandIn', ac => chatStandIn(ac, cr, cc, rs, t))
+    );
+    chatStandInParticipant.iconPath = vscode.Uri.joinPath(context.extensionUri, 'resources', 'GitHubCopilotforAzure.svg');
+
+    context.subscriptions.push(chatStandInParticipant);
+}
+
+async function chatStandIn(
+    _actionContext: IActionContext,
```
bwateratmsft, line 30: Any suggestions for telemetry to capture? *We can't capture the button clicks unless we go through the trouble of registering wrapper commands.

bwateratmsft: We should probably go through the trouble.

package.json
MRayermannMSFT: > Can the ID be different?

I think that would be a good thing to do from a telemetry standpoint. Should make sure there's nothing weird experience wise from that though.

src/chat/chatStandIn.ts
MRayermannMSFT: > We should probably go through the trouble.

Agreed.

package.json
bwateratmsft: Testing suggests it doesn't really work well if that's done.

MRayermannMSFT:
Need to make sure this doesn't ship until the GHCP4A side has shipped.

package.json
```diff
@@ -632,6 +697,11 @@
                         "items": {
                             "type": "string"
                         }
+                    },
+                    "azureResourceGroups.enableChatStandIn": {
```
MRayermannMSFT, line 701: So just to check, the `configurationDefaults` will override this?

bwateratmsft: Yes, I've experimented and it seems to be working as intended.

nturinski (approved):

PR #1035, opened 2025-02-04
Create fake ask azure command

JasonYeMSFT:
This PR creates a fake `Ask @azure` command to the resource extension to expose it when the GitHub Copilot for Azure extension is not installed.

## Visibility

As of today, the `Ask @azure` command is only displayed when the GitHub Copilot for Azure extension is installed. After https://github.com/microsoft/vscode-azureresourcegroups/pull/1034 adds a stand-in chat provider, we can make this command always available to the user and let the stand-in chat provider handle the scenarios where the that extension is not installed.


src/chat/resourceConnection.ts
```diff
@@ -0,0 +1,152 @@
...
+            type: type,
+            kind: kind,
+            name: name,
+            subscriptionId: subscriptionId,
+            resourceGroup: resourceGroup,
+        };
+    }
+}
+
+export async function getTokenCredentialForAzureResourceConnection(subscriptionProvider: AzureSubscriptionProvider, connection: AzureResourceConnection): Promise<{ credential: TokenCredential, tenantId: string } | undefined> {
```
MRayermannMSFT, line 138: This doesn't need to be in this extension. I'll delete next week while Jason is gone.

src/chat/resourceConnection.ts
```diff
@@ -0,0 +1,152 @@
...
+
+export type AzureResourceConnection<TypeT = string, KindT = string | undefined> = {
+    type: TypeT;
+    kind?: KindT;
+    name: string;
+    subscriptionId: string;
+    resourceGroup: string;
+};
+
+export function partialConnectionHasAtLeastOneField(partialConnection: object): boolean {
```
MRayermannMSFT, line 17: A bunch of these helpers probably aren't needed? Will need to double check next week while Jason is gone.

src/chat/chatStandIn.ts
```diff
@@ -40,7 +40,11 @@ async function chatStandIn(
         command: 'azureResourcesGroups.installGitHubCopilotForAzureFromChat',
     });
 
-    responseStream.markdown(vscode.l10n.t('After that, please repeat your question.'));
+    if (request.prompt.startsWith("/setResourceContext ")) {
```
MRayermannMSFT, line 43: Talked to Jason offline. I'll refactor to just do an exact prompt match.

bwateratmsft (approved):

PR #1037, opened 2025-02-06
Add an Ask @azure command to command palette

JasonYeMSFT:
Add an Ask azure command in command palette that can take a user prompt using input box and submit the prompt to the GitHub Copilot for Azure agent in chat window.


spboyer (approved):

jinglouMSFT:
@JasonYeMSFT if a user asks a question from the command palette, would you reset the chat history or continue with the existing chat history? Or it's conditional?  I don't see that in the code here.  Is that handled on the @azure side?

spboyer:
> @JasonYeMSFT if a user asks a question from the command palette, would you reset the chat history or continue with the existing chat history? Or it's conditional? I don't see that in the code here. Is that handled on the @Azure side?

It should start a new conversation each time the question is asked. 

The code to do that is in the PR
```
await vscode.commands.executeCommand("workbench.action.chat.newChat");
```

MicroFish91 (dismissed):

src/chat/askAzure.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from "node_modules/@microsoft/vscode-azext-utils";
```
bwateratmsft, line 6: It's a type-only import so it doesn't end up mattering after transpilation, but this import should be:
```suggestion
import { IActionContext } from "@microsoft/vscode-azext-utils";
```

spboyer:
Can we do an "if" here? If there is an existing conversation then continue otherwise start a new

JasonYeMSFT:
We reached an offline decision that we are going to always start a new conversation when the command is invoked.

MRayermannMSFT (approved):

alexweininger (approved):

bwateratmsft (approved):

PR #1052, opened 2025-02-18
Add v2 resources API for getting selected resource and recently used resources

MRayermannMSFT:


src/utils/usedAndSelectedResources.ts
```diff
@@ -0,0 +1,83 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ext } from "../extensionVariables";
+
+function nodeHasIdAndMaybeResource(node: unknown): node is { id: string, resource?: { kind?: string } } {
```
MRayermannMSFT, line 8: If there's a better way to be doing this, please let me know.

I couldn't find any types that exposed the `resource` property on nodes. I also found that `AzExtTreeItem` does expose the `id` property at least, but none of the nodes I tested against were able to pass the `isAzExtTreeItem` type guard so 🤷‍♀️.

api/src/resources/resourcesApi.ts
```diff
@@ -60,4 +60,14 @@ export interface ResourcesApi {
      * @param options - Options for revealing the resource. See {@link vscode.TreeView.reveal}
      */
     revealAzureResource(id: string, options?: VSCodeRevealOptions): Promise<void>;
+
+    /**
+     * Gets a list of Azure resources (ARM ID and maybe kind) that have been recently used/interacted with in the Azure Resources extension.
+     */
+    getRecentlyUsedAzureResources(): Promise<{ id: string, kind?: string }[]>;
```
MRayermannMSFT, line 67: I don't have a hard feeling on the names of these functions. Please tell me if you want them changed.

src/utils/usedAndSelectedResources.ts
bwateratmsft: I think the type you're looking for is `AzureResourceItem`

api/src/resources/resourcesApi.ts
alexweininger: Should we make a note that it doesn't track subscriptions or resource groups?

src/utils/usedAndSelectedResources.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+function getArmIdAndMaybeKindFromTreeNode(node: unknown | undefined): { id: string, kind?: string } | undefined {
+    if (node && nodeHasIdAndMaybeResource(node)) {
+        const resourceIdParts = node.id.split("/");
+        const subscriptionsIdx = resourceIdParts.indexOf("subscriptions");
+        const resourceGroupsIdx = resourceIdParts.indexOf("resourceGroups");
+        const providersIdx = resourceIdParts.indexOf("providers");
+        if (subscriptionsIdx === resourceIdParts.length - 8 && resourceGroupsIdx === resourceIdParts.length - 6 && providersIdx === resourceIdParts.length - 4) {
+            const resourceId = resourceIdParts.slice(-8).join("/");
```
alexweininger, line 34: We typically keep the beginning slash

```suggestion
            const resourceId = "/" + resourceIdParts.slice(-8).join("/");
```

src/utils/usedAndSelectedResources.ts
alexweininger: Yeah, if we're only looking for resources not including subs, RGs, and nodes below resources then `AzureResourceItem` should work.

src/utils/usedAndSelectedResources.ts
```diff
@@ -0,0 +1,54 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzExtTreeItem } from "@microsoft/vscode-azext-utils";
+import { TreeView } from "vscode";
+import { ext } from "../extensionVariables";
+
+
+export async function getSelectedAzureResource(): Promise<string | undefined> {
+    const resourceTreeView = (ext.appResourceTreeView as TreeView<AzExtTreeItem>);
```
bwateratmsft, line 12: I don't think these trees actually use `AzExtTreeItem`, right @alexweininger?

MRayermannMSFT: I went off the fact that this is the assignment to the extension variable.

`ext.appResourceTreeView = treeView as unknown as vscode.TreeView<AzExtTreeItem>;`

alexweininger: I would do `{ id?: string }`.

alexweininger (approved):
LGTM. I'll follow up with a PR that preps the api package for release.

alexweininger:
Since you bumped the version, could you do steps 1-4 from this doc too? https://github.com/microsoft/vscode-azureresourcegroups/blob/main/api/CONTRIBUTING.md

alexweininger (approved):

PR #1068, opened 2025-02-25
Fix ask azure gif resource path

JasonYeMSFT:


bwateratmsft (approved):
Aha! That's what it is!

PR #1071, opened 2025-02-28
Changes for adding sub/resource id telemetry

nturinski:
Relies on https://github.com/microsoft/vscode-azuretools/pull/1906/files

Same issue here-- we can't shallow copy the context as it won't preserve the properties that the wizards alter.


src/api/compatibility/pickAppResource.ts
```diff
@@ -18,6 +18,7 @@ export function createCompatibilityPickAppResource() {
             childItemFilter: convertExpectedChildContextValueToContextValueFilter(options?.expectedChildContextValue)
         });
 
+        context.resourceId = result.id;
```
nturinski, line 21: This is needed for client extensions that use this when retrieving a resource via tree item picker (usually the command pallette)

alexweininger (approved):

PR #1109, opened 2025-03-31
Update activity log collapsible state and log total duration

MicroFish91:
WIP... 
Relies on upcoming Azure Tools PRs, may be out of sync with changes proposed there, but illustrates the basic ideas.

This PR:
- Makes the initial collapsible state expanded so that progress items are now shown by default
- Shows the duration of an activity in the description of the ActivityTreeItem


MicroFish91:
Working off of Nathan's v2 changes now, closing as no longer relevant

PR #1113, opened 2025-04-04
Add files to support V2 API ActivityLog view

nturinski:
Honestly, there's some stuff in here that almost just seems like boilerplate so I'm not 100% sure that I needed to add everything at this point. But I'm sure that we will want to program some more specific behavior in the future, so I just included everything like the ActivityLogResource, etc.


src/tree/activitiyLog/ActivityItem.ts
bwateratmsft, line 1: The folder name containing this file is misspelled

src/tree/activitiyLog/ActivityLogDefaultBranchDataProvider.ts
```diff
@@ -0,0 +1,34 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+import * as vscode from 'vscode';
+import { BranchDataProvider, ResourceBase, ResourceModelBase } from '../../../api/src/index';
+import { ActivityLogResource } from './activityLog';
+
+interface ActivityLogResourceModel extends ResourceModelBase {
```
bwateratmsft, line 9: There is a duplicate identifier of the same name in `activityLog.ts`

src/tree/activitiyLog/ActivityItem.ts
```diff
@@ -3,18 +3,20 @@
...
+import { ResourceGroupsItem } from "../ResourceGroupsItem";
 
 export enum ActivityStatus {
     Running = 'running',
     Done = 'done'
 }
 
-export class ActivityTreeItem extends AzExtParentTreeItem implements Disposable {
-
+export class ActivityItem implements ResourceGroupsItem, Disposable {
```
MicroFish91, line 18: I think you can remove `hasMoreChildrenImpl` now since this is no longer an `AzExtParentTreeItem`

MicroFish91:
Closes https://github.com/microsoft/vscode-azureresourcegroups/issues/1111

src/tree/ResourceGroupsItem.ts
```diff
@@ -3,8 +3,10 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { TreeElementBase } from 'node_modules/@microsoft/vscode-azext-utils';
```
MicroFish91, line 6: node_modules in import path

MicroFish91:
Unrelated, but may as well have the initial collapsible state set to Expanded so people can see the progress items now 🥳

nturinski:
> Unrelated to v2, but may as well have the initial collapsible state set to Expanded now so people can see the activity children right away 🥳

I'm trying to think, but there's no case where the activity won't have children most likely, right? I guess we can have an optional param to specify that behavior if we need to.

MicroFish91:
Actually that's a good point, I think if there are no `context.activityResult` and `context.activityChildren` then technically there won't be any children.  I think we may need to make some changes to tools to get that info, so I guess you can omit for now.  

My draft PR that I'm recommending people use in the mean time already has the state defaulted to Expanded, so it shouldn't block anyone if we omit from this PR.

MicroFish91 (approved):

PR #1119, opened 2025-04-15
Add live timer to activities in the activity log

MicroFish91:
Relies on https://github.com/microsoft/vscode-azuretools/pull/1967: 

Example images:
image

image

We had discussed adding a live activity timer and hiding progress reports from the activity log description since we were migrating to showing activity children anyway.

(Note: Progress reports would still be used in commands leveraging the Azure Wizard without registering an activity, see below.)
https://github.com/microsoft/vscode-azuretools/blob/713123e274ceb0ae434395bbe80048e6780cddfa/utils/src/wizard/AzureWizard.ts#L293


alexweininger:
Where will progress messages go now?

MicroFish91:
> Where will progress messages go now?

Oops, I put that info in the other PR but didn't add it here.  Updated, let me know if you have any other questions!

MicroFish91:
Update: Waiting to merge this and the associated utils work until after next RG release

src/tree/activityLog/ActivityItem.ts
```diff
@@ -30,10 +30,21 @@ export class ActivityItem implements TreeElementBase, Disposable {
...
+            if (this.activity.startTime && this.activity.endTime) {
+                const startTimeMs: number = this.activity.startTime.getTime();
+                const endTimeMs: number = this.activity.endTime.getTime();
+                duration = dateTimeUtils.getFormattedDurationInMinutesAndSeconds(endTimeMs - startTimeMs);
+            }
+
             if (this.error) {
-                return localize('failed', 'Failed');
+                return duration ?
+                    localize('failedWithDuration', 'Failed ({0})', duration) :
```
nturinski, line 42: I think I prefer if it said `Failed in 20s` over the ( ). Nonblocking of course.

src/tree/activityLog/ActivityItem.ts
```diff
@@ -71,10 +82,10 @@ export class ActivityItem implements TreeElementBase, Disposable {
     public error?: unknown;
     private latestProgress?: { message?: string };
 
-    public constructor(activity: Activity) {
+    public constructor(readonly activity: Activity) {
         this.id = activity.id;
         this.setupListeners(activity);
-        this.startedAtMs = Date.now();
+        this.startedAtMs = this.activity.startTime?.getTime() ?? Date.now();
```
nturinski, line 88: I don't actually see this being used anywhere. We can probably just remove this property and use the `activity.startTime`.

It also doesn't make sense that it used to set `this.startedAtMs` on construction _and_ `onStart`, overwriting the time.

nturinski (dismissed):

src/tree/activityLog/ActivityItem.ts
MicroFish91: Do you prefer that for success as well?

src/tree/activityLog/ActivityItem.ts
MicroFish91: Yeah I think this was a legacy implementation from when we used `compareChildrenImpl` for v1.

nturinski (approved):

PR #1123, opened 2025-04-17
Have a backwards compat shim for the activity log view

nturinski:
Also expanding the tree node automatically. We need it for the steps-- we can figure out how to actually check for it later, but it's necessary for 90% of the commands anyway,

### DIFF
--- a/src/tree/activityLog/ActivityItem.ts
+++ b/src/tree/activityLog/ActivityItem.ts
@@ -65,7 +65,7 @@ export class ActivityItem implements TreeElementBase, Disposable {
         }
     }
 
-    public initialCollapsibleState: TreeItemCollapsibleState = TreeItemCollapsibleState.Expanded;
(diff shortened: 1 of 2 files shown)


MicroFish91:
Nice, I'll test with a couple commands in a bit

src/tree/activityLog/ActivityItem.ts
```diff
@@ -65,7 +65,7 @@ export class ActivityItem implements TreeElementBase, Disposable {
         }
     }
 
-    public initialCollapsibleState: TreeItemCollapsibleState = TreeItemCollapsibleState.None;
+    public initialCollapsibleState: TreeItemCollapsibleState = TreeItemCollapsibleState.Expanded;
```
MicroFish91, line 68: Can we ship this without it being initially set to expanded?  There are a couple scenarios in container apps that look kind of weird without the progress implementations.  I plan to address this once we migrate that extension to V2, but for initial pass maybe we only show the children at the end?

MicroFish91 (approved):

PR #1124, opened 2025-04-19
Hook up Copilot to the Activity Log and expose basic context commands

MicroFish91:
WIP...

Todo:  
1. Replace Copilot icons
2. Wait on feedback about header ribbon icon verbiage and behavior


package.json
```diff
@@ -101,6 +101,37 @@
...
+                        }
+                    }
+                },
+                "modelDescription": "Gets the Azure activity log",
+                "name": "azureResources_getAzureActivityLog",
+                "canBeReferencedInPrompt": true,
+                "toolReferenceName": "azureActivityLog",
+                "tags": [
+                    "azure",
+                    "vscode_editing"
```
bwateratmsft, line 131: We should verify if this line is still needed. Previously, Agent mode would only call tools with this tag.

src/chat/askAgentAboutActivityLog.ts
```diff
@@ -0,0 +1,38 @@
...
+export async function askAgentAboutActivityLogItem(_: IActionContext, promptType: ActivityLogPromptType, node?: ActivityItem | ActivityChildItemBase): Promise<void> {
+    if (!node?.id) {
+        return await askAgentAboutActivityLog();
+    }
+    await vscode.commands.executeCommand("workbench.action.chat.newChat");
+    await vscode.commands.executeCommand("workbench.action.chat.open", { mode: 'agent', query: generateActivityLogPrompt(promptType, node.id) });
+}
+
+const explainActivityLogPrompt: string = vscode.l10n.t('Help explain the important information in my VS Code activity log.');
+const fixActivityLogPrompt: string = vscode.l10n.t('Help me diagnose and fix the error in my VS Code activity log.');
```
bwateratmsft, line 25:
```suggestion
const explainActivityLogPrompt: string = vscode.l10n.t('Help explain the important information in my Azure activity log.');
const fixActivityLogPrompt: string = vscode.l10n.t('Help me diagnose and fix the error in my Azure activity log.');
```

src/chat/askAgentAboutActivityLog.ts
```diff
@@ -0,0 +1,38 @@
...
+function generateActivityLogPrompt(promptType: ActivityLogPromptType, treeId?: string): string {
+    switch (promptType) {
+        case ActivityLogPromptType.Fix:
+            return treeId ?
+                `${fixActivityLogPrompt} ${vscode.l10n.t(`I'm interested in fixing the activity item with treeId: {0}.`, treeId)}` :
+                `${fixActivityLogPrompt}`;
+        case ActivityLogPromptType.Explain:
+            return treeId ?
+                `${explainActivityLogPrompt} ${vscode.l10n.t(`I'm interested in an explanation of activity item with treeId: {0}.`, treeId)}` :
+                `${explainActivityLogPrompt}`
```
bwateratmsft, line 36: This is awkward, referencing a non-user-visible ID in a user-visible prompt. Have we found that it's really necessary? I would be surprised if Copilot wasn't good enough at figuring out what is important.

src/chat/tools/GetAzureActivityLog/GetAzureActivityLog.ts
```diff
@@ -0,0 +1,42 @@
...
+    public async invoke(context: IActionContext, options: vscode.LanguageModelToolInvocationOptions<T>, _token: vscode.CancellationToken): Promise<vscode.LanguageModelToolResult> {
+        const convertedActivityItems: ConvertedActivityItem[] = await convertActivityTreeToSimpleObjectArray(context, options.input.treeId);
+
+        return {
+            content: [
+                convertedActivityItems.length ?
+                    constructLMToolResponse(JSON.stringify(convertedActivityItems), options.input.promptType ?? ActivityLogPromptType.Explain) :
+                    new vscode.LanguageModelTextPart('There is no activity data to analyze.')
+            ]
+        };
```
bwateratmsft, line 21: IIRC, `vscode.LanguageModelToolResult` is a class, so we should use the constructor to create the return result.

src/chat/tools/GetAzureActivityLog/GetAzureActivityLog.ts
```diff
@@ -0,0 +1,42 @@
...
+
+function constructLMToolResponse(treeItemData: string, promptType: ActivityLogPromptType): vscode.LanguageModelTextPart {
+    switch (promptType) {
+        case ActivityLogPromptType.Fix:
+            return new vscode.LanguageModelTextPart(
+                `Tree item data: ${treeItemData}.
+                Analyze the activity log for any errors or failures. Focus on diagnosing the root cause of the issue and provide actionable steps to resolve it.
+                If an item is marked as selected, prioritize your analysis on the selected item and its relationship to the top-level parent activity item.
+                The top-level parent item represents the command that was run. Include any relevant context or dependencies that may have contributed to the issue.
+                Avoid exposing the raw structure or content of the underlying object in your response.`
```
bwateratmsft, line 33: Can this be split up into two `vscode.LanguageModelTextPart`s?

src/chat/tools/GetAzureActivityLog/GetAzureActivityLog.ts
```diff
@@ -0,0 +1,42 @@
...
+            );
+        case ActivityLogPromptType.Explain:
+            return new vscode.LanguageModelTextPart(
+                `Tree item data: ${treeItemData}. If an item is marked as selected, focus your summary on the selected item.
+                When a child item is selected, explain its relationship to the top-level parent activity item representing the command that was run.
+                When summarizing the data in responses, do not directly expose the structure or content of the underlying object.`
```
bwateratmsft, line 39: Same here

src/chat/askAgentAboutActivityLog.ts
MicroFish91: > This is awkward, referencing a non-user-visible ID in a user-visible prompt.

Yeah that's fair.  What options would you prefer for this?  Should we try to target using a generic but non-unique identifier that the user can see (i.e. the `label`)?  Or are you thinking about just removing the ability to target specific tree items entirely?  

(Another option could also be to expose a `viewProperties` for these items at some point).

> Have we found that it's really necessary? I would be surprised if Copilot wasn't good enough at figuring out what's important.

In some cases it can be a nice-to-have, but certainly not necessary.  I would say if we're diagnosing an error, it's mostly not necessary; Copilot will assume it needs to focus on the error.  That said, if the user wanted to have Copilot explain an item that isn't related to the error, it would need to provide follow-up guidance.

There is definitely an argument for not allowing target of specific tree items. especially considering the extra logic overhead that might be required.  I think in most cases, allowing users to target specific tree items mostly has the benefit of removing the amount of prompts the user needs to give by ~1.

Other Considerations: In our one pager discussions, part of improving the discoverability was to add inline icons for each tree item as well as the ability to right click on specific tree items to issue ask commands.

package.json
MicroFish91: Oh okay good to know, I will test it without the tag

src/chat/tools/GetAzureActivityLog/GetAzureActivityLog.ts
MicroFish91: These response types were meant to correspond with the right click ask options on activity log tree items.  Based on the above conversation, since we are not planning to ship the first pass with this capability, I'll remove this.

src/chat/tools/GetAzureActivityLog/GetAzureActivityLog.ts
MicroFish91: Ditto https://github.com/microsoft/vscode-azureresourcegroups/pull/1124#discussion_r2052973652 :)

PR #1135, opened 2025-05-07
feat: support new type of AI Foundry projects

hund030:
There is a new resource type for Azure AI Foundry projects. We need to list both type of projects in the tree view.

### DIFF
--- a/src/azureExtensions.ts
+++ b/src/azureExtensions.ts
@@ -141,7 +141,7 @@ export const azureExtensions: IAzExtMetadata[] = [
 ];
 
 export const legacyTypeMap: Partial<Record<AzExtResourceType, string>> = {
-    AiFoundry: 'microsoft.machinelearningservices/workspaces',
+    AiFoundry: 'microsoft.cognitiveServices/accounts',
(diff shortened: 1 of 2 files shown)


src/azureExtensions.ts
```diff
@@ -141,7 +141,7 @@ export const azureExtensions: IAzExtMetadata[] = [
 ];
 
 export const legacyTypeMap: Partial<Record<AzExtResourceType, string>> = {
-    AiFoundry: 'microsoft.machinelearningservices/workspaces',
+    AiFoundry: 'microsoft.cognitiveServices/accounts',
```
alexweininger, line 144: Why has this been replaced and not just added?

hund030: 1. This doesn't support adding two values with same key.
2. The `legacyTypeMap` is referenced in `createGroupConfigFromResource` and `createAzureExtensionsGroupConfig`, neither of them is used. So, this change doesn't seem to take any effect actually. 

If one day the `legacyTypeMap` take effect and it can only map to one type, I think it should be the new type.

alexweininger (approved):

PR #1175, opened 2025-07-09
Add logic to write README to auth Record location.

g2vinay:


src/exportAuthRecord.ts
```diff
@@ -122,6 +122,36 @@ async function persistAuthRecord(authRecord: Record<string, unknown>): Promise<v
...
+        '| `authority`     | The Microsoft Entra authority used for authentication               |',
+        '| `clientId`      | The client ID of the app that performed the original authentication |',
+        '| `tenantId`      | The associated Microsoft Entra tenant ID                            |',
+        '| `username`      | The username of the logged in account                               |',
+        '| `homeAccountId` | A unique identifier for the account                                 |',
+        '',
+        '### Security considerations',
+        '',
+        "- The user profile's `.azure` directory is already used by other products, such as MSAL and Azure CLI to store metadata in `msal_token_cache.bin` and `azureProfile.json`, respectively.",
+        "- While `authRecord.json` itself isn't inherently dangerous, it should still be excluded from source control. A preconfigued `.gitignore` file is written alongside the file for that purpose.",
```
joshfree, line 152: I think this is good. Only non-blocking suggestion is to add an aka.ms link that we own here to another page; but we can do that in a follow up PR if/as needed.

alexweininger (dismissed):

src/exportAuthRecord.ts
```diff
@@ -122,6 +122,36 @@ async function persistAuthRecord(authRecord: Record<string, unknown>): Promise<v
     const authDir = path.join(baseAzureDir, 'ms-azuretools.vscode-azureresourcegroups');
     const authRecordPath = path.join(authDir, 'authRecord.json');
     const gitignorePath = path.join(authDir, '.gitignore');
+    // Write README.md explaining the purpose and contents of authRecord.json
+    const readmePath = path.join(authDir, 'README.md');
+    const readmeContent = [
```
bwateratmsft, line 127: Instead of inline, can we put the readme content as a constant at the root of the file (to make indentation easier on the eyes), and use this string notation?

```typescript
const foo = `
bar
`;
```

src/exportAuthRecord.ts
```diff
@@ -122,6 +122,36 @@ async function persistAuthRecord(authRecord: Record<string, unknown>): Promise<v
...
+        'The `authRecord.json` file is created after authenticating to an Azure subscription from VS Code. For example, via the **Azure: Sign In** command in Command Palette. The directory in which the file resides matches the unique identifier of the [Azure Resources extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureresourcegroups) responsible for writing the file.',
```
bwateratmsft, line 128: The first time we use "VS Code" we should expand the name
```suggestion
        'The `authRecord.json` file is created after authenticating to an Azure subscription from Visual Studio Code. For example, via the **Azure: Sign In** command in Command Palette. The directory in which the file resides matches the unique identifier of the [Azure Resources extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureresourcegroups) responsible for writing the file.',
```

localden (approved):

PR #1177, opened 2025-07-15
Add Copilot telemetry and log VS Code command IDs for `GetAzureActivityLog`

MicroFish91:
I thought maybe this might help us triangulate better which commands were run

Relies on: https://github.com/microsoft/vscode-azuretools/pull/2037

Also includes the latest utils which adds the telemetry point `isCopilotEvent` to the LM tool calls


MicroFish91:
I preferred calling these `commandIds`, but since the `IActionContext` will call these `callbackIds`, I'll keep it the same.

nturinski (approved):

Issue #1200, opened 2025-08-12
Adopt proposed authentication APIs by September 15


alexweininger:
To avoid breakage, we must adopt the new authentication APIs by Sept 15. New API details are here https://github.com/microsoft/vscode/issues/260156

We'll have to update our auth package and let consumers of that know it'll be changing and they'll need to upgrade by Sept 15th too.

alexweininger:
Example of the new UX working:

1. Try to create a resource group
2. VS Code prompts to sign in (it requires MFA)
3. Sign in with MFA
4. Resource group creates successfully

https://github.com/user-attachments/assets/aff7f2ea-cdcb-4c3f-aaf2-077768d44c7c

After this, subsequent create resource group attempts don't require the prompt since the session is now MFA'd

bwateratmsft:
This is done, just need to adopt the finalized APIs

PR #1205, opened 2025-08-19
api: Support challenges when getting sessions from VS Code

alexweininger:
This is the first change that must be released in order to work towards closing https://github.com/microsoft/vscode-azureresourcegroups/issues/1200. This release will be consumed in utils.

Challenge support is currently a proposed api available in Insiders. It will be released in the next Stable release 1.104.

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## Unreleased
 
+## [2.6.0] - 2025-08-19
+
+- Support challenges in `AzureAuthentication.getSessionWithScopes`. This relies on the proposed authenticationChallenges VS Code API.
+
 ## [2.5.1] - 2025-08-06
 
 * Add `DurableTaskHub` resource type
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## [2.6.0] - 2025-08-19
 
-- Support challenges in `AzureAuthentication.getSessionWithScopes`. This relies on the proposed authenticationChallenges VS Code API.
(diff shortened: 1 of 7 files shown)


motm32 (dismissed):

api/docs/vscode-azureresources-api.d.ts
```diff
@@ -491,5 +492,4 @@ export declare interface Wrapper {
     unwrap<T>(): T;
 }
 
-export { };
-
+export { }
```
bwateratmsft, line 495: Nit: is this line necessary at all, since this file isn't generated?

api/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## Unreleased
 
+## [2.6.0] - 2025-08-19
+
+- Support challenges in `AzureAuthentication.getSessionWithScopes`. This relies on the proposed authenticationChallenges VS Code API.
```
bwateratmsft, line 11: Can you add a link to the VSCode issue with the proposal?

api/package.json
```diff
@@ -18,7 +18,7 @@
     "license": "MIT",
     "devDependencies": {
         "@types/node": "^16.0.0",
-        "@types/vscode": "^1.64.0"
+        "@types/vscode": "^1.90.0"
```
bwateratmsft, line 21: We should pin to a specific version so that `vsce` doesn't get mad; i.e. remove the `^`. You will also need to `npm i` again as it actually picked up `@types/vscode@1.103.0` in the lockfile.

api/docs/vscode-azureresources-api.d.ts
alexweininger: This file is generated and it just changes all the time super annoying

api/package.json
alexweininger: Ohh it'll see that we're using proposed api that isn't in that types version?

bwateratmsft: `vsce` will complain if you use `@types/vscode` higher than the minimum engine version specified in the extension's package.json.

Since this is a package itself (not an extension), the `@types/vscode` dev dependency won't bubble up to the extension, and `vsce` won't actually notice--but the same problems can result at runtime.

bwateratmsft (approved):

PR #1214, opened 2025-08-26
Enable activity item selection when providing context to Copilot

MicroFish91:
Here's a demo [link](https://microsoft-my.sharepoint.com/:v:/p/matfisher/EXWe6WxJm3lBiHZVH2uRwdcBEvx-6mgPablGHoM-1j_0-A?nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJPbmVEcml2ZUZvckJ1c2luZXNzIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXciLCJyZWZlcnJhbFZpZXciOiJNeUZpbGVzTGlua0NvcHkifX0&e=vLuikw)

This recording is a bit dated - since then I have changed the icon as well as have all activities start out selected by default.

### DIFF
--- a/src/chat/tools/GetAzureActivityLog/logTelemetry.ts
+++ b/src/chat/tools/GetAzureActivityLog/logTelemetry.ts
@@ -31,7 +31,6 @@ export function logActivityTelemetry(context: GetAzureActivityLogContext, activi
 export function logSelectedActivityTelemetry(context: GetAzureActivityLogContext, selectedActivityItems: ConvertedActivityItem[]): void {
     const telemetry: TelemetryProperties = getCommonTelemetryProperties(selectedActivityItems);
 
-    context.telemetry.properties.hasSelectedActivities = String(!!selectedActivityItems.length);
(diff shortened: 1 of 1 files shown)


src/commands/registerCommands.ts
```diff
@@ -130,7 +130,8 @@ export function registerCommands(): void {
         node.setAllSubscriptionsLoaded();
         ext.azureTreeState.notifyChildrenChanged(node.id);
     });
-    registerCommand("azureResourceGroups.askAgentAboutActivityLog", askAgentAboutActivityLog);
+    registerCommandWithTreeNodeUnwrapping("azureResourceGroups.askAgentAboutActivityLog", async (context: IActionContext) => await askAgentAboutActivityLog(context));
```
MicroFish91, line 133: Calling the function here this way is intentional.  This way we can explicitly control what nodes get passed in.  It prevents accidentally selected nodes from being inadvertently passed in

PR #1223, opened 2025-09-02
Migration of `Azure Cosmos DB for MongoDB (RU and vCore)` support to the new `vscode-documentdb` extension

tnaum-ms:
This pull request improves the handling of resource types for `Azure Cosmos DB` and `Azure Cosmos DB for MongoDB (vCore)`, with a focus on clearly distinguishing between the following variants:

- existing: `Azure Cosmos DB` (support remains in the `vscode-cosmosdb` extension)
- new: `Azure Cosmos DB for MongoDB (RU)` (support moved to the `vscode-documentdb` extension)
- existing: `Azure Cosmos DB for MongoDB (vCore)` (support moved to the `vscode-documentdb` extension)
  - Please note: the dedicated node for this resource is intentional.

It also initiates the deprecation of the `MongoClusters` resource type, transitioning it to `Azure Cosmos DB for MongoDB (vCore)`.

ℹ️ **Please note**

- In order to ensure a synchronized activation of these changes across three extensions, a temporary API has been added and is exposed.

---
_Content generated by GitHub Copilot:_

**Resource Type Enhancements**

* Added new resource types to the `AzExtResourceType` enum: `AzureCosmosDbForMongoDbRu` and `AzureDocumentDb`, and marked `MongoClusters` as deprecated.
* Updated display names for the new resource types in `azureUtils.ts`, ensuring clear differentiation between Cosmos DB variants.

**Resource Type Mapping and Identification**

* Enhanced `getAzExtResourceType` logic to distinguish between standard Cosmos DB and Cosmos DB for MongoDB (RU) accounts based on the resource kind, and updated the mapping for `mongoclusters` to use `AzureDocumentDb` instead of the deprecated `MongoClusters`. [[1]](diffhunk://#diff-46035f93451955827db967e7c21633acd4c8ceb52ee6a7b2031e660fbdfd091bR11) [[2]](diffhunk://#diff-46035f93451955827db967e7c21633acd4c8ceb52ee6a7b2031e660fbdfd091bR42-R49) [[3]](diffhunk://#diff-46035f93451955827db967e7c21633acd4c8ceb52ee6a7b2031e660fbdfd091bL58-R67)

**Extension Metadata Updates**

* Updated `azureExtensions.ts` to associate the new Cosmos DB resource types with the `vscode-documentdb` extension, supporting the migration from the old Azure Databases extension.

**Temporary API**

[...]

PR #1229, opened 2025-09-09
Cherry-pick #1223: Migration of `Azure Cosmos DB for MongoDB (RU and vCore)` support to the new `vscode-documentdb` extension

bwateratmsft:
Cherry-picks #1223.


bwateratmsft:
@tnaum-ms do you need a new release created for the `@microsoft/vscode-azureresources-api` package?

tnaum-ms:
@bwateratmsft Thank you for taking care of this work item. I really appreciate it.

We don't need these types to be out. I get around this with some casting. So, strictly speaking, we don't need these types to be published in the package npm package.

However, `vscode-azureresources` needs to use the new `getAzExtResourceType` function in order to correctly support the new branch in the Azure Resources view. If I understand the code base correctly, this will work just fine without an updated API package.

nturinski (approved):

PR #1241, opened 2025-09-18
Ensure correct session is being used when listing resources 

### LINKED ISSUE #1040: Resource list does not show resources granted access via B2B user when user is granted direct access via lighthouse to external subscription
As a partner, my account is often granted access to customer Azure subscriptions in another tenant via Lighthouse. If I am separately invited to customer tenant via B2B and granted access to the same subscription using my B2B account in their tenant, the Resources list in VS Code will only show the resources via Lighthouse. When Lighthouse permissions are removed for that subscription, I am able to see resources granted to B2B user. I am never able to see resources granted to B2B user when Lighthouse has granted permission to the same subscription.

motm32:
Fixes #1040 

In the case of a user having both lighthouse and B2B access being granted, the same subscription can be under the same account but different tenants. In this case the session being used to list resources is incorrect since it doesn't include the tenant scope. By directly calling `getSessionFromVSCode` we are able to solve this issue.

To Do:

- [x] Release and update auth package changes
- [x] Make changes to how the same sub shows under the same account currently they have the same name: 
image

_Edit_

When the same sub is from the same account the tenant id is shown instead: 
image

### DIFF
--- a/src/tree/azure/AzureResourceTreeDataProvider.ts
+++ b/src/tree/azure/AzureResourceTreeDataProvider.ts
@@ -94,7 +94,7 @@ export class AzureResourceTreeDataProvider extends AzureResourceTreeDataProvider
                     const duplicates = getDuplicateSubscriptions(subscriptions);
                     let duplicatesWithSameAccount: AzureSubscription[] = [];
                     if (duplicates.length > 0) {
-                        duplicatesWithSameAccount = getDuplicateSubsInSameAccount(duplicates)
+                        duplicatesWithSameAccount = getDuplicateSubsInSameAccount(duplicates);
                         if (duplicatesWithSameAccount.length > 0 && !this.hasShownDuplicateWarning) {
                             this.hasShownDuplicateWarning = true;
                             void callWithTelemetryAndErrorHandling('azureResourceGroups.duplicate', async (context: IActionContext) => {


src/services/AzureResourcesService.ts
```diff
@@ -1,27 +1,30 @@
...
+    listResourceGroups(subscription: AzureSubscription): Promise<ResourceGroup[]>;
 }
 
 export const defaultAzureResourcesServiceFactory = (): AzureResourcesService => {
-    async function createClient(context: IActionContext, subscription: AzureSubscription): Promise<ResourceManagementClient> {
-        const subContext = createSubscriptionContext(subscription);
-        return await createResourceClient([context, subContext]);
+    async function createClient(subscription: AzureSubscription): Promise<ResourceManagementClient> {
+        const session = await getSessionFromVSCode("https://management.azure.com//.default", subscription.tenantId, { createIfNone: false, silent: true, account: subscription.account })
+        const credential = createCredential(() => session);
```
bwateratmsft, line 16: I have a sneaking suspicion that this fixed scope is gonna come back to bite us

src/services/AzureResourcesService.ts
```diff
@@ -1,27 +1,30 @@
...
+        const client = new ResourceManagementClient(credential, subscription.subscriptionId);
```
bwateratmsft, line 17: We need to still be using `createResourceClient`. That adds a ton of extra stuff including support for the MFA challenges that we need.

motm32:
Talked offline but decided we may want to put these changes behind a setting so lighthouse users can opt in to this experience. Basically if we notice that there are multiple of the same subs under the same account a warning will pop up to turn the setting on. That way we can _hopefully_ mitigate any bugs 🤪

alexweininger (dismissed):

alexweininger (dismissed):

alexweininger:
Could you describe in a little bit more detail how to create this scenario so we can remember how to test it in the future?

nturinski (approved):

Issue #1253, opened 2025-09-30
Copilot chat asks to install GitHub Copilot for Azure extension


jonasanderssonswg:
Type: <b>Bug</b>

1. Ask copilot anything in Chat window (Ask mode)

Expected results: Copilot doesn't ask me to install an unecessary extension ("GitHub Copilot for Azure") and runs the query.
Actual results: Copilot asks me to install an unecessary extension ("GitHub Copilot for Azure") before running the query.

Extension version: 0.11.4
VS Code version: Code 1.104.2 (e3a5acfb517a443235981655413d566533107e92, 2025-09-24T11:21:37.073Z)
OS version: Windows_NT x64 10.0.26100
Modes:

MicroFish91:
My understanding is the current behavior is to toast this when Copilot intent detection sees something Azure related.

We should modify the toast UI to make it easier for user's to turn off this notification.  I believe the current user setting is `azureResources.chatStandIn`.

PR #1256, opened 2025-10-02
Add button to chat stand in to disable `enableChatStandIn` setting

motm32:
fixes: #1253 

Here is how this looks in the chat: 
image

I am open to changing the wording a bit I wasn't sure what would be best 😅


bwateratmsft:
I think it's good, though the "Or you can disable..." message might be overkill. @fiveisprime what do you think?

src/chat/chatStandIn.ts
```diff
@@ -41,8 +43,13 @@ async function chatStandIn(
...
     });
 
+    responseStream.button({
+        title: vscode.l10n.t(`Don't ask me again`),
+        command: 'azureResourceGroups.updateChatStandInSetting'
+    });
+
     const postButtonMessage = request.prompt === askAgentAboutResourcePrompt ? vscode.l10n.t(`After that, please use \`Ask @azure\` again.`) :
-        vscode.l10n.t('After that, please repeat your question.');
+        vscode.l10n.t(`After that, please repeat your question.\n\n`);
```
bwateratmsft, line 52: Is this line change still needed?

PR #1257, opened 2025-10-06
Convert `@microsoft/vscode-azureresources-api` from CJS to CJS+ESM, and update for finalization of auth challenges API

bwateratmsft:


MicroFish91 (approved):
Very cool 😎

PR #1279, opened 2025-10-29
Pick up better fix for event looping

bwateratmsft:


src/tree/ResourceTreeDataProviderBase.ts
bwateratmsft: This part was a bad issue in the past made much worse by #1277. In short, we would re-register the event listener every time `getAzureSubscriptionProvider()` was called, which is quite a bit. In the past they were debounced together so only one would get through, but with #1277 that stopped being the case, and it was not good.

Much better to just register the event listener one time, like we intended.

PR #1285, opened 2025-11-06
Add receiver method to the type definition for `AzureExtensionApi`

MicroFish91:
I need this type updated so I can include it in utils.  
I need all the package typings aligned so that typescript stops yelling at me :P

Plan to eventually add a README link in the comment as well, but that would be in the next iteration.

Needed as a prerequisite for this:
https://github.com/microsoft/vscode-azureresourcegroups/pull/1284

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## Unreleased
 
-## [3.1.0] - 2025-11-10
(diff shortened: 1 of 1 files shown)


api/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@microsoft/vscode-azureresources-api",
-    "version": "3.0.0",
+    "version": "3.0.1",
```
bwateratmsft: Obviously it's very subjective but this feels more like a 3.1.0 than a 3.0.1 :smile:

PR #1288, opened 2025-11-12
Azure Resources API (v4) internal tests

MicroFish91:


test/global.test.ts
```diff
@@ -19,7 +19,8 @@ suiteSetup(async function (this: Mocha.Context): Promise<void> {
     this.timeout(1 * 60 * 1000);
 
     // Initialize test API - this caches it for use throughout tests
-    await getTestApi();
+    const testApi = await getTestApi();
+    registerUIExtensionVariables(testApi.extVars);
```
MicroFish91: This is just a placeholder for right now to get the tests to pass again.  

I think Alex is already addressing how to handle extension variables, so I'll wait to defer to however it gets set up in his [PR](https://github.com/microsoft/vscode-azureresourcegroups/pull/1315).

PR #1316, opened 2025-12-02
Add Rerun with copilot command

motm32:
This will allow users to re run commands in the activity log. Currently I have a entry point on the top level activity item which looks like this: 
image

But I am open to changing this if people would prefer a different or more obvious entry point. 

To do: 
- [x] Finish [utils changes](https://github.com/microsoft/vscode-azuretools/pull/2136) and release them

Here is a demo of the flow: 

https://github.com/user-attachments/assets/86e9fc61-f322-470f-a072-8e15b22a9c3d


src/copilot/reRunWithCopilot.ts
```diff
@@ -0,0 +1,41 @@
...
+*--------------------------------------------------------------------------------------------*/
+
+import { CopilotUserInput, createSubscriptionContext, executeCommandWithAddedContext, type IActionContext } from "@microsoft/vscode-azext-utils";
+import * as vscode from "vscode";
+import { ActivitySelectedCache } from "../chat/askAgentAboutActivityLog/ActivitySelectedCache";
+import { convertActivityTreeToSimpleObjectArray, ConvertedActivityItem } from "../chat/tools/GetAzureActivityLog/convertActivityTree";
+import { GetAzureActivityLogContext } from "../chat/tools/GetAzureActivityLog/GetAzureActivityLogContext";
+import { ActivityItem } from "../tree/activityLog/ActivityItem";
+
+export async function reRunWithCopilot(context: IActionContext, item: ActivityItem): Promise<void> {
```
nturinski: nit: We discussed it offline, but prefer it to be rerunWithCopilot just for stylistic reasons.

nturinski (dismissed):

motm32:
The tests are failing due to a bug in the @vscode/extension-telemetry package there is already a PR up for the fix so I am hoping that gets merged soon and then I will update this PR with the new version.

PR #1355, opened 2026-02-03
Fix: ensure splitting is disabled in esbuild configuration

### LINKED ISSUE #1352: The Azure panel fails to load after installing Azure Resources extension with an Azure Functions project open
**OS:** Win11 x64
**Build Version:** 20260130.1

**Repro Steps:**
1. Install the Azure Functions extension in VS Code.
2. Open an Azure function project in VS Code.
3. Install the latest version of Azure Resources extension.
4. Navigate to Azure view.
5. Check whether the Azure panel loads successfully.

**Expect:**
The Azure panel loads successfully without any errors.

**Actual:**
The Azure panel fails to load with an error.
Image

alexweininger:
Seeing if this fixes #1352

### DIFF
--- a/api/src/utils/wrapper.ts
+++ b/api/src/utils/wrapper.ts
@@ -1,3 +1,8 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
 /**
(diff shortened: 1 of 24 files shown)


bwateratmsft (approved):
Approved, but are we sure it's the splitting that's causing the problem?

In any case, there will be splitting done with the telemetry bundle, so it can't be avoided altogether.

bwateratmsft:
Sounds good. We should have CTI check the folder to see if that file exists after the error has appeared, or if it's still gone...

PR #1383, opened 2026-02-25
Performance improvements made by copilot

nturinski:
I did an experiment by implementing a telemetry profiler to see which commands were taking a long time to run and used it to create prompts to get Copilot to improve perf.

Most of it involve parallelizing await async code, but the double telemetry wrapping was an interesting one. I'm not 100% if it's correct in that we don't need to wrap it again, but it made sense to me.

It also made a rudimentary cache since Managed Identity nodes have to leverage the getResource call as well.

Also, wasn't sure about batching all of the tenant subscriptions calls into one telemetry event. Could possibly break someone's dashboard.

Here's some time differences between fixes:
image

image

image

image

image

image

image


nturinski:
Sorry, this is a little bit of AI slop in here. I think the parallel promises I'll probably keep, but I mostly just wanted some visual aids for when we have to present our topics on Monday. I'm changing this to a draft PR.

PR #1436, opened 2026-05-07
Add copilot on rails webviews 

motm32:
Made some changes in the webview package also to support this will put that code up and link it here soon. I tested this out and it seems to work fine with opening all of these new views while debugging. Will test with a vsix also before merging.


alexweininger (dismissed):
We should check the new vsix size just for fun

motm32:
> We should check the new vsix size just for fun

Just checked and from released version the size is around 6.39 MB and now it is 8.51 MB.

alexweininger (approved):

PR #1445, opened 2026-05-14
Change Create project with copilot entrypoint

motm32:
After talking yesterday the button I added seemed hard to discover so adding this new pane: 
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -432,6 +432,11 @@
                     "visibility": "visible"
                 },
                 {
+                    "id": "azureProject",
+                    "name": "%azureProject.viewName%",
+                    "visibility": "visible"
+                },
+                {
                     "id": "azureTenantsView",
                     "name": "Accounts & Tenants",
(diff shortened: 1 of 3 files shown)


MicroFish91 (approved):
Nice I like the way it looks.

For some reason I'm still seeing the old entry-point.

Image

But I see in your screenshot that it's not there; maybe it's a me problem but just wanted to double check!

MicroFish91:
@naturins Should we install (on the devbox) the custom VSIX that gets generated by this PR after it gets merged?

motm32:
> For some reason I'm still seeing the old entry-point.


Yeah just checked again and I don't see it.

nturinski:
> @naturins Should we install (on the devbox) the custom VSIX that gets generated by this PR after it gets merged?

Let me finish making some changes, we can install it all at once tomorrow?

nturinski (approved):

PR #1449, opened 2026-05-20
Update Copilot on Rails views 

motm32:
The azure-prepare command made some changes to the md file so I updated the deploy view. It now looks something like this: 

image

I also made some updates to the local dev view the biggest being now users can modify the names if the launch configs: 
image


MicroFish91:
@nturinski The end goal is to remove `.agents/` right?

nturinski:
Yeah, we will move it to how you did.

nturinski (approved):

src/tree/project/AzureProjectProgressTreeDataProvider.ts
```diff
@@ -0,0 +1,274 @@
...
+        for (const disposable of this.disposables) {
+            disposable.dispose();
+        }
+    }
+
+    refresh(): void {
+        this.onDidChangeTreeDataEmitter.fire();
+    }
+
+    getTreeItem(element: ProgressNode): vscode.TreeItem {
```
MicroFish91, line 56: I think it gets messy if we include all tree item resolution logic in one method for the tree data provider to resolve.  

I suggest consolidating this logic like we normally do:
- A separate StageNode class
- A separate ActionNode class
- Just a simple `getChildren` call on the generic element type

Example:
Image

Same suggestion applies to `getTreeItem`.

src/tree/project/projectPlanFiles.ts
```diff
@@ -0,0 +1,112 @@
...
+    hasDeploymentPlan: boolean;
+    /** True when any of the plan files exist. */
+    hasAny: boolean;
+    /** The furthest stage reached: 0 = project creation, 1 = local dev, 2 = deployment. */
+    currentStage: ProjectStage;
+}
+
+const PLAN_FILE_GLOBS = [
+    '**/.azure/project-plan.md',
+    '**/vscode-debug-plan.md',
```
MicroFish91: can also use include the .azure directory for the vscode-debug-plan

src/tree/project/projectPlanFiles.ts
```diff
@@ -0,0 +1,112 @@
...
+    '**/.azure/deployment-plan.md',
+] as const;
+
+export async function getProjectPlanFiles(): Promise<ProjectPlanFiles> {
+    const [projectPlanFiles, localDevelopmentPlanFiles, deploymentPlanFiles] = await Promise.all(
+        PLAN_FILE_GLOBS.map((glob) => vscode.workspace.findFiles(glob, '**/node_modules/**', 1)),
```
MicroFish91: In addition to `node_modules`, there's probably a lot of other file patterns that we might want to exclude.  A couple ideas:

1. We can make a list of patterns to exclude and then combine them into an exclude pattern or
2. Instead of looking for the plans `**/..` we just look in `.azure/` and expect it to be at the root of the project where we won't need to include the same level of exclusions

src/tree/project/AzureProjectProgressTreeDataProvider.ts
MicroFish91: Actually instead of separate Stage nodes and Action nodes, what if we further split into their own tree items (e.g. CreationStageItem, DebugStageItem, DeploymentStageItem, etc.)?

src/webviews/copilotOnRails/extension/utils/singletonViewHost.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Open the plan/requirements source file in an editor, or warn the user when
+ * the location isn't known. Shared by every webview controller's "open file"
+ * affordance.
+ */
+export function openSourceFileOrWarn(sourceFileUri: vscode.Uri | undefined): void {
+    if (!sourceFileUri) {
+        void vscode.window.showWarningMessage(
+            vscode.l10n.t('The plan file location is unknown. Locate it manually in the workspace.'),
```
MicroFish91, line 40: What action does the user need to take for "Locate it manually in the workspace." ?

src/webviews/copilotOnRails/extension/copilotOnRailsCommands.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+export const copilotOnRailsCommandIds = {
+    createProjectWithCopilot: 'azureResourceGroups.createProjectWithCopilot',
+    startLocalDevelopment: 'azureResourceGroups.startLocalDevelopment',
+    startDeployment: 'azureResourceGroups.startDeployment',
+    openScaffoldPlanView: 'azureResourceGroups.openPlanView',
+    openLocalPlanView: 'azureResourceGroups.openLocalPlanView',
+    openDeploymentPlanView: 'azureResourceGroups.openDeployPlanView',
+    openRequirementsView: 'azureResourceGroups.openRequirementsView',
+    completeStep: 'azureProjectCreation.completeStep',
```
MicroFish91: `completeStep` prefix is different, consider changing for consistency

src/webviews/copilotOnRails/extension/createProjectWithCopilot.ts
```diff
@@ -6,55 +6,53 @@
...
-    switch (true) {
-        // Local Development => Deploy
-        case await hasCompletedPhase('**/local-development-plan.md', 'implemented'): {
-            const choice = await vscode.window.showInformationMessage(
-                vscode.l10n.t('We detected a previous Copilot session with a completed local debug configuration. Would you like to deploy this project?'),
-                { modal: true },
-                deploy,
-            );
+    // Local Development complete => offer Deploy.
+    if (await hasCompletedPhase('**/vscode-debug-plan.md', 'implemented')) {
```
MicroFish91: Should add `.azure/` as part of the pattern.  Consider removing `**/` depending on how opinionated we want to be.  See similar comment about this above.

src/webviews/copilotOnRails/extension/createProjectWithCopilot.ts
```diff
@@ -6,55 +6,53 @@
...
 export async function createProjectWithCopilot(_context: IActionContext): Promise<void> {
-    switch (true) {
-        // Local Development => Deploy
-        case await hasCompletedPhase('**/local-development-plan.md', 'implemented'): {
-            const choice = await vscode.window.showInformationMessage(
-                vscode.l10n.t('We detected a previous Copilot session with a completed local debug configuration. Would you like to deploy this project?'),
-                { modal: true },
-                deploy,
-            );
+    // Local Development complete => offer Deploy.
```
MicroFish91: Consider changing these section comments back as they are wordier

src/webviews/copilotOnRails/extension/utils/singletonViewHost.ts
```diff
@@ -0,0 +1,125 @@
...
+ * reveal it, and tear down the watcher when the panel is disposed.
+ */
+export class SingletonViewHost<TData, TController extends RevealableWebview> {
+    private controller: TController | undefined;
+    private watcher: vscode.Disposable | undefined;
+
+    constructor(
+        private readonly options: {
+            readonly createController: (data: TData, sourceFileUri: vscode.Uri | undefined) => TController;
+            readonly updateController: (controller: TController, data: TData, sourceFileUri: vscode.Uri | undefined) => void;
```
MicroFish91, line 86: Should we rename `updateController` here? It looks more like we're using the passed in controller to update the view data, not updating the actual controller.  Maybe `updateView`, `updateViewData`, or similar?

src/webviews/copilotOnRails/views/utils/parseRequirements.ts
```diff
@@ -0,0 +1,182 @@
...
+    questions: RequirementsQuestion[];
+    parseError?: {
+        message: string;
+        fileLabel?: string;
+    };
+}
+
+export type RequirementsInputType = 'text' | 'number' | 'boolean' | 'tags' | 'select' | 'multiselect';
+
+export function inferInputType(
```
MicroFish91, line 55: Future TODO, we should add unit tests for all these parsing functions

src/webviews/copilotOnRails/extension/utils/singletonViewHost.ts
motm32: As in they need to open the raw markdown themselves to look at the plan? This is basically a fallback for if the view is not able to rendered and causes and error and then when users try to open the markdown with the button provided it is also unable to open it. It is a very rare case but this is the warning message that gets shown.

MicroFish91 (approved):

PR #1454, opened 2026-05-27
`npm audit fix` and version bump for rel/0.12.5

bwateratmsft:


src/utils/logging/stringifyHttp.ts
```diff
@@ -18,7 +18,7 @@ export interface HttpStringifier {
  */
 function stringifyRequest(request: NormalizedHttpRequest, source: string, verbose?: boolean): string {
     let message = `[${source} Request]`;
-    message = `\n┌────── ${source} Request ${request.method} ${request.url}`;
+    message += `\n┌────── ${source} Request ${request.method} ${request.url}`;
```
nturinski, line 21: This seems more correct, but just wanted to make sure that this is intended behavior since it's changing what it's doing fundamentally.

nturinski (approved):

src/utils/logging/stringifyHttp.ts
bwateratmsft: So in fairness this is what we did on `main`, and it _seems_ to be what was actually intended. In any case, it's only for logging so how bad could it be? 😅

PR #1481, opened 2026-06-08
deps: bump azext-azureauth to 6.1.0-alpha.2 for #5073 audience-consent fix

alexweininger:
Ships the createIfNone support in getSessionWithScopes (microsoft/vscode-azuretools#2314). The Azure Resources extension provides the runtime authentication object spread onto every subscription context, so this bump is what lets appservice's eager App Service audience consent actually prompt the user fixing 'You are not signed in to an Azure account' on deploy.

Addresses microsoft/vscode-azurefunctions#5073


package-lock.json
```diff
@@ -3219,6 +3266,21 @@
                 "@azure/ms-rest-azure-env": "^2.0.0"
             }
         },
+        "node_modules/@microsoft/vscode-azext-azureutils/node_modules/@microsoft/vscode-azext-azureauth": {
```
bwateratmsft: Can you look into why this is being bundled again? My guess is that NPM is not properly resolving that 6.1.0-alpha.1 is compatible

package-lock.json
```diff
@@ -3176,20 +3205,38 @@
...
+                "@azure/core-auth": "^1.10.1",
+                "@azure/core-rest-pipeline": "^1.23.0",
+                "@azure/identity": "^4.13.1",
                 "@azure/ms-rest-azure-env": "^2.0.0"
             },
             "engines": {
                 "vscode": "^1.106.0"
             }
         },
+        "node_modules/@microsoft/vscode-azext-azureauth/node_modules/@azure/arm-resources-subscriptions": {
```
bwateratmsft: Can you look into why this is being bundled again?

alexweininger: It was always bundled, but as v2.1.0 (azureutils still imports the classic SubscriptionClient at runtime). What's new here isn't that copy reappearing; it's a second major version, 3.0.0-beta.1, getting added. Bumping azext-azureauth to 6.1.0-alpha.2 brings in auth's move to the modular v3 arm SDK (https://github.com/microsoft/vscode-azuretools/pull/2308). Since azureutils is still on the v2 classic client, npm can't dedupe the two majors, so both end up in the tree.

Should we leave this as-is, or dive into migrating the rest of the shared packages and extensions to v3 now?

package-lock.json
alexweininger: Yeah this can be fixed by expanding the range.

package-lock.json
bwateratmsft: Let's migrate them to v3 of `@azure/arm-resources-subscriptions`. It should not be difficult if we use the client in the non-modular way.

alexweininger: done

bwateratmsft (approved):

PR #1482, opened 2026-06-08
Move skills to resources/agents instruction files

nturinski:


MicroFish91 (approved):
Not sure what we want to do about the shared-references folder, but let's merge and test this

PR #1488, opened 2026-06-11
Add command to download agent instruction files to workspace

nturinski:
## Summary

The custom agents are contributed via `package.json`, but they rely on instruction folders under `resources/agents/` that must be present in the user's workspace to function. This PR downloads those instruction folders into the workspace and gates agent invocation on their presence.

## Changes

- **New command** `azureResourceGroups.downloadAgentInstructions` ("Azure: Download Azure Agent Instructions") copies all bundled instruction folders into `.github/agents/`.
  - Downloads all 6 instruction folders: `azure-debug-generate`, `azure-debug-plan`, `azure-project-plan`, `azure-project-scaffold`, `azure-project-test`, and `shared-references`.
  - When invoked directly from the Command Palette, it downloads **without prompting** (running the command is treated as consent).
- **Verification gate** (`ensureAgentInstructions`) runs before every custom-agent invocation:
  - **Missing** → modal prompt explaining why the files are needed; **Download** copies only the missing folders, declining returns `false` to abort the hand-off.
  - **Present** → returns `true` immediately with no prompts or version checks.
  - No workspace open → no-op (returns `true`).
- Wired the gate into `openChatWithAgent` (hand-off commands) and the `ScaffoldPlanView`, `CreateProject`, `LocalPlan`, and `Requirements` webview controllers.

### DIFF
--- a/src/commands/copilotOnRails/agentInstructions.ts
+++ b/src/commands/copilotOnRails/agentInstructions.ts
@@ -1,0 +1,178 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext, UserCancelledError } from '@microsoft/vscode-azext-utils';
+import * as vscode from 'vscode';
+import { ext } from '../../extensionVariables';
+
+/** Name of the metadata file bundled alongside each instruction folder. */
+const METADATA_FILE = '.metadata.json';
+
+/** Folder (relative to the workspace root) where instruction folders are downloaded. */
+const WORKSPACE_AGENTS_RELATIVE_PATH = ['.github', 'agents'];
+
+interface AgentInstructionMetadata {
+    name: string;
+    version: string;
+}
+
+/**
+ * Maps each contributed chat agent to the instruction folders (under
+ * `resources/agents`) it relies on at runtime. Agents that don't ship bundled
+ * instruction folders (e.g. `azure-deploy`, which reads an external skill) are
+ * intentionally omitted.
+ */
+const agentInstructionFolders: Record<string, string[]> = {
+    'azure-project-plan': ['azure-project-plan', 'shared-references'],
+    'azure-project-scaffold': ['azure-project-scaffold', 'shared-references'],
+    'azure-debug-plan': ['azure-debug-plan'],
+    'azure-debug-generate': ['azure-debug-generate'],
+};
+
+/** Every distinct instruction folder across all agents. */
+const allInstructionFolders: string[] = [...new Set(Object.values(agentInstructionFolders).flat())];
+
+/** Root of the instruction folders bundled with the extension. */
+function getBundledAgentsRoot(): vscode.Uri {
+    return vscode.Uri.joinPath(ext.context.extensionUri, 'resources', 'agents');
+}
+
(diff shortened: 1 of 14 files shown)


nturinski:
@copilot

This should download all of the agent instruction files. Here is a list of them all:
azure-debug-generate
azure-debug-plan
azure-project-plan
azure-project-scaffold
azure-project-test
shared-references

Keep an array of strings that have all of the agents.

nturinski:
@copilot If the user has the instructions but doens't upgrade, that shouldn't block the plan from showing. Updating or not, as long as they have, the helper should return true.

motm32 (approved):

src/commands/copilotOnRails/agentInstructions.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/** Copies the given instruction folders from the extension bundle into the workspace, replacing any existing copies. */
+async function copyInstructionFolders(folders: string[], agentsRoot: vscode.Uri): Promise<void> {
+    const bundledRoot = getBundledAgentsRoot();
+    for (const folder of folders) {
+        const source = vscode.Uri.joinPath(bundledRoot, folder);
+        const dest = vscode.Uri.joinPath(agentsRoot, folder);
+        try {
+            await vscode.workspace.fs.delete(dest, { recursive: true, useTrash: false });
```
MicroFish91: Would this delete all content in this folder, even if it wasn't related to our agent instructions? If so, consider more targeted deletion limited to the agent instructions folders you have stored in the variable above

src/commands/copilotOnRails/agentInstructions.ts
```diff
@@ -0,0 +1,163 @@
...
+ * running the command is treated as explicit consent to write into `.github/agents`.
+ */
+export async function downloadAgentInstructions(_context: IActionContext): Promise<void> {
+    const agentsRoot = getWorkspaceAgentsRoot();
+    if (!agentsRoot) {
+        throw new Error(vscode.l10n.t('Open a folder or workspace before downloading Azure agent instructions.'));
+    }
+
+    await copyInstructionFolders(agentInstructionFolders, agentsRoot);
+    void vscode.window.showInformationMessage(vscode.l10n.t('Azure agent instructions downloaded to ".github/agents".'));
```
MicroFish91, line 103: `context.ui`

src/commands/copilotOnRails/agentInstructions.ts
```diff
@@ -0,0 +1,163 @@
...
+                detail: vscode.l10n.t('Your workspace has an older version of these instructions in ".github/agents". Upgrade to the latest version?'),
+            },
+            upgrade,
+        );
+        if (choice !== upgrade) {
+            return true;
+        }
+    }
+
+    await copyInstructionFolders(agentInstructionFolders, agentsRoot);
```
MicroFish91: If the user wants to download or upgrade, do we always copy over the latest from the extension over into their workspace?  Would we care if the major version changed?

MicroFish91:
Nit: Can simplify some calls using `AzExtFsExtra`

nturinski:
@copilot Instead of updating based on version number, simply download if it doesn't exist and ignore if it does. Do not use the versions in the metadata.json in any code.

nturinski:
@copliot Use AzExtFsExtra anywhere you are able instead of vscode.fs

nturinski:
@copilot Use AzExtFsExtra anywhere you are able instead of vscode.fs

PR #1494, opened 2026-06-15
Make changes for accessibility and add loading view 

motm32:
I made some accesibility changes so when zoomed in at 200% the view is able to be scrolled and the words don't go over or get hidden. 

I also added a loading view so there is always a view open. Here is an example of how the loading view looks (the wording changes depending on which stage you are in): 

image

Finally changed the input box so it is closed it now looks like this: 
image


src/webviews/copilotOnRails/extension/controllers/RequirementsViewController.ts
```diff
@@ -71,6 +72,12 @@ export class RequirementsViewController extends WebviewController<Record<string,
         void this.panel.webview.postMessage({ command: 'submitComplete' });
 
         const relativePath = vscode.workspace.asRelativePath(this.sourceFileUri);
+        this.panel.dispose();
```
nturinski: Any idea why this moved up?

nturinski (approved):

src/webviews/copilotOnRails/extension/controllers/RequirementsViewController.ts
motm32: Because we need to dispose of the requirements view panel before opening the loading view.

motm32:
Still need to remove the debugging commands

MicroFish91 (approved):
Let's go ahead and merge this if we can, just remember to remove your commands that you marked with TODO :)

MicroFish91:
Oh small nit, but could you move the comment description saying what each number means above `currentStage`, and move it up to associate with `ProjectStage` instead.  That way when you hover over the type, it's immediately obvious via hover popup what each number means

image

PR #1498, opened 2026-06-16
Change instructions to rename src folder to services for monorepos

nturinski:


MicroFish91 (approved):
I assume that the agent understands that it has the freedom to provide better names instead of using examples like "functions" and "web" all the time?

PR #1511, opened 2026-06-23
Add manual tenant sign-in fallback for Conditional Access tenants

CoffeeKanzler:


CoffeeKanzler:
@microsoft-github-policy-service agree

bwateratmsft (commented):
The Azure Resources extension supports multi-tenant authentication on purpose. You can filter out tenants you aren't interested in looking at with the Accounts and Tenants view and the subscription filter. I don't see why this is necessary.

CoffeeKanzler:
You can not login into some tenants that enforce that via conditional login. You NEED to specifiy the tenant manually to start the Auth workflow via the Microsoft Authenticator

bwateratmsft:
Does the "Sign in to Tenant (Directory)..." command work?

CoffeeKanzler:
No sadly it does not.  The flow from my understanding is:

Azure: Sign In 
-> provider.signIn(undefined) 
-> authentication.getSession("microsoft", ARM scopes) 
-> no VSCODE_TENANT scope 
-> Microsoft Authentication defaults to the "organizations" authority

Sign in to Tenant 
-> getAccounts() 
-> for each already-authenticated account: getTenantsForAccount() 
-> call Azure listTenants() 
-> populate Quick Pick with returned tenants 
-> user selects one of those discovered tenants 
-> provider.signIn(selectedTenant) 
-> VSCODE_TENANT:<selectedTenantId> is finally added

getAccounts() requires an account already registered with the VS Code Microsoft authentication provider. If the initial generic authentication failed, there is no account to use for tenant discovery.

This creates a circular dependency:

Tenant-specific authentication is required to satisfy Conditional Access and sign into tenant requires generic authentication and tenant discovery before it can request tenant-specific authentication

bwateratmsft:
That makes sense. Perhaps an input box in `Sign in to Tenant (Directory)...` to ask for tenant ID, or something like that would work? If the default sign in fails?

(side note, I wish Azure auth wasn't so nightmarishly overcomplicated...)

CoffeeKanzler:
Yeah that sounds like a better Idea than my current solution. I can look into that and rewrite this.

CoffeeKanzler:
I'll rework the description and work on the findings later this week probably. For now my tests are successful but I'll have to work in the additions of Copilot

CoffeeKanzler:
@bwateratmsft One open design question I'd like your call on before this is final:

When a user manually enters a tenant and sign-in succeeds, we persist it (`globalState`) and use it as the **account-level** authentication fallback — i.e. it's injected as the `tenantId` only for contexts that don't already specify one (the "list tenants for account" path). Contexts that already have a tenant ID ignore it, and ARM's tenant-list endpoint is account-wide, so it doesn't hide other tenants from the UI.

The thing I'm unsure about is its **lifecycle**: it's currently never cleared. The one case that can actually wedge things is a *stale/revoked* manual tenant — if the user later loses access to it, account-level enumeration keeps trying to get a token for it. Clearing it on a successful *discovered* sign-in doesn't help there (you can't reach discovery if enumeration is already blocked).

Options I see: (a) leave it as-is, (b) add an explicit "Clear manual tenant" affordance, or (c) self-heal by dropping it if its silent token acquisition fails. Do you have a preference, or would you rather keep this PR to the bootstrap fix and handle lifecycle separately?

src/commands/accounts/signInToTenant.ts
bwateratmsft, line 1: We should make the changes in the auth package in `microsoft/vscode-azuretools` rather than duplicating the code there. Potentially we add a tenant or tenant ID arg to that function? @alexweininger thoughts?

bwateratmsft (commented):
Once an account is signed in (to a specific tenant), does it still hit the endpoint that CA is blocking? The MSAL broker might be "sticky" enough to remember the tenant for us once the account has been signed in, removing the need to persist the tenant ID ourselves. Not certain, but worth trying.

src/commands/accounts/signInToTenant.ts
alexweininger: @coffeekanzler thanks for opening the PR!

`getPicks` here duplicates the auth package's existing picker, can we instead just delegate to the auth package's `signInToTenant(provider)` and only fall back to the manual tenant input box when it throws `NotSignedInError`? That avoids the duplication and any auth-package changes. Also, do we actually need the `manualSignInTenant` persistence + `getSubscriptionContext` override?

For example: https://github.com/microsoft/vscode-azureresourcegroups/compare/main...alexweininger/sign-in-to-tenant-manual-entry

CoffeeKanzler: Happy to refactor the picker to delegate to the auth package's signInToTenant(provider) with an input-box fallback on NotSignedInError, @alexweininger.

  On whether we still need the manualSignInTenant persistence + getSubscriptionContext override: I tested exactly this. The manual sign-in itself succeeds, but resources never populate afterward.

  The reason: subscription enumeration goes through getSubscriptionContext({ account, tenantId: undefined }) → listTenants(), and with tenantId undefined the silent token request hits the common/organizations
  endpoint. So no subscriptions come back and the tree stays empty. The override re-injects the signed-in tenant into that account-level context, which scopes the silent token correctly and makes resources load. 

It has to be persisted (not just in-memory) because the same enumeration re-runs on every reload.

  @bwateratmsft — to your MSAL-stickiness question, that's what I was hoping too, but in practice the broker is only sticky for the interactive sign-in, not for the silent enumeration token. Without the persisted
  tenant, a reload gives an empty tree.

  So my proposal: Refactor the delegation (drop the duplicated picker) and keep the tenant persistence + context override (necessary for resources to actually load). 
  
  If you'd rather that re-scoping live in the auth package instead like e.g. a "default/sticky tenant" the provider applies to account-level requests then I'd need to further study the auth package after I am back.

  Heads up on timing: I'm traveling in about an hour and won't be back until Friday at the earliest, so apologies in advance for the slow follow-up. I'll pick this up as soon as I'm back.

alexweininger: Sorry to go back and forth on it, but now that we know we need to persist that value, then I think it's best kept as a setting like your original design.

1. Drop the `manualSignInTenant` globalState, use an advanced setting instead.
2. Move the discovery fix to `getTenantsForAccount` (we already override it, and it gets the `account`): when ARM `/tenants` returns empty/throws for a conditional-access user, inject `AzureTenant { tenantId: <setting>, account }`. The normal per-tenant flow then lists subscriptions via that tenant's session, so we can drop the `getSubscriptionContext` override entirely.
3. Keep using the setting for the initial `signIn({ tenantId })`.

summary: one setting + the existing public override, no globalState, no internal coupling. Clients consume our discovered subscriptions via the Resources API, so this covers the whole extension family too.

PR #1531, opened 2026-06-29
Increase max requests when using copilot on rails flow

nturinski:
image

When running the Copilot on Rails scaffold flow, a low chat.agent.maxRequests limit causes Copilot to pause partway through with a "continue to iterate?" prompt. This PR raises that limit to a high ceiling at the Workspace scope so scaffolding can run uninterrupted — for both autopilot and guided runs.

**Changes**
Autopilot (autopilot.ts): replaced the temporary global maxRequests override (saved/restored) with a persistent Workspace bump (WORKSPACE_MAX_REQUESTS = 9999) that is intentionally not restored. Added getEffectiveMaxRequests() and raiseWorkspaceMaxRequests() helpers (no-op when no workspace folder is open or the limit is already high enough). The global chat.tools.global.autoApprove toggle is still saved and restored as before.
Guided scaffold (ScaffoldPlanViewController.ts): when the effective limit is below 1000, shows a Yes/No modal asking whether to raise the request limit for the workspace before launching the scaffold chat. "Yes" raises it; "No" (the modal's close affordance) proceeds unchanged. Skipped for autopilot (which already bumps it) and for the debug plan.
**Notes**
The workspace setting is written to settings.json and left in place by design.
chat.tools.global.autoApprove remains global and is always restored after an autopilot run.

### DIFF
--- a/src/utils/settingUtils.ts
+++ b/src/utils/settingUtils.ts
@@ -17,11 +17,13 @@ export namespace settingUtils {
     }
 
     /**
-     * Uses ext.prefix 'azureResourceGroups' unless otherwise specified
+     * Uses ext.prefix 'azureResourceGroups' unless otherwise specified.
+     * When `target` is omitted, VS Code resolves the scope from the provided resource;
+     * pass an explicit `target` (e.g. `ConfigurationTarget.Workspace`) to force it.
      */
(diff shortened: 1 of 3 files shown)


MicroFish91 (approved):
LGTM, although some of the explaining comments are long and repeated in multiple sections, so might be worth trimming/consolidating those

nturinski:
@copilot Remove some of the comments that are extra verbose and don't explain critical code paths

PR #1535, opened 2026-06-30
Add resume support for the Create with Copilot flow

nturinski:
## Resume support for the "Create with Copilot" flow

Interrupted create-with-copilot runs (window reload, crash, or an abandoned chat session) previously left the user without an obvious way back in — they had to guess which chat session to reopen, and re-launching a phase restarted it from scratch. This PR detects the interrupted phase and lets the user resume from the correct spot.

### How it works
- **`flowState`** resolves the current phase + status from the `.azure/*` plan files (the source of truth, written by the agent doing the work) reconciled with a small `workspaceState` cache that records the last-launched phase. `computeFlowState` is pure and unit-tested; `resolveFlowState` does the disk reads.
- **Phase-launch recording** is added where phases actually start: `openChatWithAgent` (`recordAgentLaunch`) and the plan-approval controllers (`recordPhaseLaunch`).
- **Resume affordances** (`resumeAffordances.ts`): a status-bar item plus a one-time, non-modal activation prompt. Both funnel into a single `resumeProjectWithCopilot` command and are suppressed while a flow view (plan / requirements / loading) is already open — if you're already looking at the driving view, there is nothing to "resume".
- **Tree "Resume" action** routes through `resumeProjectWithCopilot` rather than invoking a raw phase command, so phase-specific resume arguments apply (notably the scaffold "continue, don't re-approve" prompt).
- **Scaffold agent** now treats `In Progress` as a valid, resumable status (`Approved → In Progress`) instead of hard-stopping when the plan is no longer exactly `Approved`.
- **Flow-view tracking** added to `singletonViewHost` (`trackFlowView` / `onDidChangeFlowViewState` / `isAnyFlowViewOpen`) so the affordances react to views opening and closing.

### Tests
- `test/flowState.test.ts` — covers `computeFlowState` across phases/statuses and `phaseToStageIndex`.

### Notes
- Targets `feat/CoR`.
- Scoped to the resume feature only; unrelated in-progress UI-preview changes were intentionally left out of this branch.


nturinski:
Too many changes all at once; gonna try breaking this up into separate PRs.

PR #1538, opened 2026-07-01
Add Select Subscriptions button to Resources view toolbar

alexweininger:
I've always wondering if this would provide a better UX. I've added the Select Subscriptions button to the title bar for the Azure Resources tree view.

Screenshot 2026-07-01 at 3 40 26 PM

What do we think? In my opinion this doesn't hurt anything, and can only help users who would potentially not see the button on the subscription items.

### DIFF
--- a/package.json
+++ b/package.json
@@ -460,14 +460,19 @@
                     "group": "navigation@2"
                 },
                 {
+                    "command": "azureResourceGroups.selectSubscriptions",
+                    "when": "view == azureResourceGroups",
+                    "group": "navigation@3"
+                },
+                {
                     "command": "azureFocusView.refreshTree",
                     "when": "view == azureFocusView",
-                    "group": "navigation@3"
+                    "group": "navigation@4"
                 },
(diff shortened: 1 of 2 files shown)


alexweininger:
Matt H. approved

motm32 (dismissed):

MicroFish91 (dismissed):
I like it too

alexweininger:
Realized we probably need to hide this button in states where there's no subscriptions in the tree yet, @copilot can you look into that? What if the user isn't signed in yet, does the command handle it fine? Should we hide the button in some cases?

motm32 (dismissed):

alexweininger:
@copilot what do we do for the other buttons on the ribbon, we should just handle this button the same as those. I think it's fine to keep it visible if the user isn't signed in yet if that's what we do for the other buttons and since the command handles the not signed in error gracefully.

bwateratmsft (approved):

PR #1539, opened 2026-07-01
Make it so autopilot carries over to new instances 

motm32:
When using autopilot when a new chat session opens it goes back to default approval. This should hopefully make it so autopilot is still set when a new session is open. 

I tested it and it seems to work. I also tested it with autopilot not set and it preformed as normal.


nturinski:
Autopilot seems to be working great but we should turn it off after the debug configurations are created (when it shows next steps for the test debug suite)

nturinski (approved):

nturinski (approved):

MicroFish91 (approved):

src/webviews/copilotOnRails/extension/openLocalDevNextStepsView.ts
```diff
@@ -20,6 +21,8 @@ let controller: LocalDevNextStepsViewController | undefined;
  * whether `api-test-collections/` contains any files in the workspace.
  */
 export async function openLocalDevNextStepsView(context: CopilotOnRailsContext, hasApiTests?: boolean): Promise<void> {
+    await disableAutopilot();
```
nturinski, line 24: Actually does this cause any problems if we call it but they don't have autopilot enabled? Or if they had it enabled, but we weren't the one that set it? You should probably check their plan file to see if we were the ones that set autopilot up.

motm32: In the disableAutopilot code STATE_ACTIVE is set only in the case when we activate autopilot ourselves so it shouldn't disable autopilot in the case the user set it up themselves or in the case it isn't enabled since it checks to see if STATE_ACTIVE is set to true or not.

PR #1550, opened 2026-07-15
Add a "Need help" button to loading views so users can easily resume commands 

motm32:
This is how it looks: 
image

Basically the button in the loading view is hooked up to the resume logic so if the user presses the button it will resume whatever step it was on without trying to redo everything. I also added some interupption instructions to the agents so if something goes wrong the agent may try to resolve the issue itself I haven't been able to test this out much since in the runs I have done nothing has gone wrong. 

I also added a delay to the "Need help" button (currently at 30 seconds) but I can remove that if we would rather always see the button


MicroFish91 (approved):

nturinski (approved):

nturinski:

nturinski:
Oh we should definitely add telemetry for this button being clicked as well.

MicroFish91 (approved):

MicroFish91 (approved):

PR #1554, opened 2026-07-16
Add warning for Azure resources emulators with limited support 

motm32:
This will add a warning symbol to the emulators sections for data store emulators with limited support. 
Here is the warning: 
image

Here is how the tooltip looks: 
image


MicroFish91:
For the tooltip, how would you feel about something like:

`Limited Support - Emulator support for this service has not yet been fully implemented` or similar

nturinski:
Is there any warning message in the requirements view as well?

Also I agree, the shorter message would be better.

motm32:
No message in the requirements view right now but I will add that.

src/webviews/copilotOnRails/views/utils/emulatorSupport.ts
```diff
@@ -0,0 +1,29 @@
...
+ *  Licensed under the MIT License. See License.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Data store services whose local emulators only partially replicate the real
+ * Azure service. Azure Storage (Azurite) and PostgreSQL have full-fidelity
+ * emulators, so they're excluded. Non-data-store services (e.g. messaging) are
+ * intentionally not flagged.
+ */
+export function isLimitedSupportDataStore(service: string): boolean {
```
MicroFish91, line 37: - I'd remove this part of the comment: `Azure Storage (Azurite) and PostgreSQL have full-fidelity emulators, so they're excluded. Non-data-store services (e.g. messaging) are intentionally not flagged.`
- Instead of two regexp tests, I wonder if we can run with just the first one.

Also how would you feel about maintaining the supported constant separately like:

```typescript
const supportedDataStoreAliases = {
  azureStorage: ["storage", "azurite"],
  postgres: ["postgres"],
} as const;

export function isLimitedSupportDataStore(service: string): boolean {
  const serviceName = service.trim().toLowerCase();
  
  const isSupported = Object.values(supportedDataStoreAliases).some((aliases) =>
    aliases.includes(serviceName)
  );
  
  return !isSupported;
}
```

MicroFish91 (approved):
Otherwise, LGTM

src/webviews/copilotOnRails/views/utils/emulatorSupport.ts
```diff
@@ -0,0 +1,29 @@
...
+        s,
+    );
+}
+
+/**
+ * The user-facing message shown when a data store lacks a full-fidelity local
+ * emulator. Kept in one place so every view surfaces identical wording.
+ */
+export function limitedSupportWarningMessage(): string {
+    return 'Limited Support - Emulator support for this service has not yet been fully implemented';
```
MicroFish91: Nit: localize

src/webviews/copilotOnRails/views/utils/emulatorSupport.ts
motm32: We can't just use the first regex since the second includes a list of data stores otherwise this warning would show up on resources that arent data stores. I can combine the two but it makes it a bit harder to read in my opinion.

MicroFish91 (approved):

PR #1558, opened 2026-07-17
CoR: Improve prerequisites detection to be more shell agnostic

### LINKED ISSUE #1556: Copilot on Rails reports Node.js/npm missing when installed through a shell version manager
## Problem

The Copilot on Rails project-plan prerequisite inventory can report Node.js and npm as not installed when they are provided by a shell version manager such as fnm or nvm.

## Reproduction

On macOS with fish as the default shell:

- `fnm` is installed at `/opt/homebrew/bin/fnm`.
- fish initializes it from `~/.config/fish/conf.d/fnm.fish` with `fnm env --use-on-cd | source`.
- `node` and `npm` resolve from an fnm multishell directory under `~/Library/Caches/fnm_multishells/.../bin`.
- An initialized fish shell reports Node.js `v22.18.0` and npm `11.12.0`.
- The Copilot on Rails project plan nevertheless marks Node.js/npm as missing.

## Root cause

`resources/agents/shared-references/prerequisites.md` assumes prerequisite checks can recover PATH by sourcing bash/zsh startup files. It explicitly does nothing for fish and then falls back to global directories such as `/opt/homebrew/bin`, `/usr/local/bin`, and `~/.npm-global/bin`.

Version managers commonly expose runtimes only after shell startup and place their shims/binaries elsewhere. The current fallback therefore misses:

- fnm multishell paths initialized by fish/zsh/bash
- nvm shell functions and selected Node versions
- other shell-initialized managers such as asdf, mise, and Volta

Directly sourcing a user's zsh startup file from a bash command is also not shell-safe when the file contains shell-specific syntax.

## Expected behavior

Before marking a CLI prerequisite missing, detection should retry the version command through the user's configured default shell in an initialized login/interactive context. If that probe succeeds, the plan should record the tool as installed with the detected version.

## Proposed fix

- Run the direct version check first.
- On failure, invoke `$SHELL` itself with login/interactive initialization rather than sourcing another shell's startup file.
- Use a marker-based result format so shell greetings/startup output do not corrupt the detected executable/version.
- Apply the fallback to every CLI prerequisite, including both Node.js and npm.
- Document version-manager coverage and avoid declaring a tool missing until both direct and initialized-shell probes fail.

MicroFish91:
Fixes #1556 

Also no longer showing an X since there are a lot of subtleties that can cause an agent in a sandboxed environment to fail to find an installed tool.  Options now will be `Installed` vs. `Unknown` to account for this.


MicroFish91:
Example output:

image

MicroFish91:
@alexweininger If you have time could you check to see if the prerequisites were detected successfully with this branch's changes?

alexweininger:
@microfish91 yes this seems to fix it

PR #1566, opened 2026-07-21
Add telemetry for next steps buttons 

motm32:
This telemetry event will show which button was clicked in both the local dev and scaffold next step views.


MicroFish91 (approved):
That change looks good, just add the same call and set pattern to the other two files and they should be good to go

MicroFish91 (approved):

MicroFish91 (approved):

PR #1572, opened 2026-07-22
Add no-datastore project requirement

### LINKED ISSUE #1568: [CoR] Missing "No datastore required" option in Project Requirements
**OS:** Win11 x64
**Build Version:** [0.12.7-alpha.0](https://github.com/microsoft/vscode-azureresourcegroups/actions/runs/29781520624)

**Repro Steps:**
1. Create a new Azure Project.
2. Enter the following prompt: "Create a simple React web application with no authentication and no database."
3. Wait for the Project Requirements view to load.
4. Navigate to the DATA section.
5. Review the available datastore options.

**Expect:**
The Requirements view should provide a "None" / "No datastore required" option. Projects that explicitly specify no database should be able to proceed without selecting a datastore.

**Actual:**
The DATA section does not provide a way to indicate that no datastore is required. The Submit button remains disabled until a datastore is selected.

alexweininger:
Fixes #1568

## Summary

Adds an explicit **"No datastore required"** option to the Copilot-on-Rails project requirements flow, so users (and the inference logic) can positively express that an app needs no persistence, storage, queue, or cache — rather than leaving the `dataStores` answer as an ambiguous empty array.

## Mutually exclusive behavior

`No datastore required` is modeled as a generic **exclusive** option (`{ "exclusive": true }`) rather than a special-cased string:

- **Parsing** (`parseRequirements.ts`): `RequirementsOption` gains an optional `exclusive` boolean, parsed from the requirements JSON. A new `toggleRequirementsOption` helper centralizes multi-select toggle logic.
- **Sentinel semantics**: selecting an exclusive option clears every other selection (concrete options *and* custom free-form values); selecting any concrete option, or typing a custom value, clears the exclusive selection. The two can never coexist.
- **UI** (`RequirementsView.tsx`): the options list and custom free-form input both route through `toggleRequirementsOption` / filter out exclusive labels, so the mutual exclusivity holds no matter how the user interacts.
- **Non-empty answer**: `["No datastore required"]` is treated as a real answer (`isAnswerEmpty` returns `false`), so the question counts as resolved.

## Downstream agent instructions

The `azure-project-plan` agent docs (`requirements.md`, `plan.md`, `azure-project-plan.agent.md`) are updated to:
- offer `No datastore required` (exclusive) in the `dataStores` question,
- infer `["No datastore required"]` when the app needs no persistence/storage/queue/cache and no service requires an associated storage account,
- never combine the sentinel with a concrete store, and
- skip adding any datastore resource to the plan (and never render the sentinel as an Azure service) when it's selected.

## Tests

Adds focused unit tests (`test/parseRequirements.test.ts`) covering exclusive-option parsing, exclusive-to-concrete replacement in both directions, and the non-empty-answer semantics.

## Validation

- `npm run lint` — passed
- `npm run build` (esbuild + webviews + `tsc --noEmit`) — passed
- `parseRequirements` test suite — 4 passing


MicroFish91 (approved):
Can you explain a bit more what "exclusive" means here? I'm not 100% sure I understand the information it's conveying

alexweininger:
> Can you explain a bit more what "exclusive" means here? I'm not 100% sure I understand the information it's conveying

If "No datastore" is selected then no other datastore options can be selected.

PR #1614, opened 2026-07-27
Add a warning for when the GHC4A extension/skills are not installed when trying to deploy

motm32:
Once a user goes from the local deb to deploy phase we check to see if the GHC4A extension is installed and if the azure prepare skill is installed. If not it will warn the user to install to continue. If the user clicks install the extension and the skills will be installed into the folder that is being used utilizing the `@azure: Install Azure Skills Locally` command

### DIFF
--- a/src/commands/copilotOnRails/openChatWithAgent.ts
+++ b/src/commands/copilotOnRails/openChatWithAgent.ts
@@ -145,14 +145,16 @@ export async function launchAgentChat(agentName: string, query: string, model?: 
 }
 
 export async function openChatWithAgent(agentName: string, prompt: string, loading?: LoadingViewConfiguration): Promise<void> {
-    if (agentName === azureDeployAgent && !(await ensureAzureDeploymentPrerequisites())) {
-        return;
+    if (agentName === azureDeployAgent) {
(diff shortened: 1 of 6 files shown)


src/commands/copilotOnRails/openChatWithAgent.ts
```diff
@@ -148,6 +145,9 @@ export async function launchAgentChat(agentName: string, query: string, model?:
 }
 
 export async function openChatWithAgent(agentName: string, prompt: string, loading?: LoadingViewConfiguration): Promise<void> {
+    if (agentName === azureDeployAgent && !(await ensureAzureDeploymentPrerequisites())) {
```
MicroFish91: I think we can also skip the `ensureAgentInstructions` check on deploy, so something like:

```typescript
    if (agentName === azureDeployAgent) {
        if (!await ensureAzureDeploymentPrerequisites(context)) {
            return;
        }
    } else {
        if (!await ensureAgentInstructions(context, agentName)) {
            return;
        }
    }
```

MicroFish91 (approved):

MicroFish91 (approved):

PR #1618, opened 2026-07-28
CoR: Add extra telemetry for `openWithChatAgent` family of functions

MicroFish91:
A lot of it is recording more of the setup telemetry (e.g. outcome of trying to install agent instructions, GitHub Copilot for Azure, Azure skills, etc.)


src/commands/copilotOnRails/agentInstructions.ts
```diff
@@ -110,14 +114,20 @@ export async function ensureAgentInstructions(agentName: string): Promise<boolea
             missingFolders.push(folder);
         }
     }
+    setCorProp(context, 'agentInstructionsMissingCount', missingFolders.length);
+    setCorProp(context, 'agentInstructionsMissingFolders', missingFolders.join(','));
 
     if (missingFolders.length === 0) {
+        // Todo: Come up with a more robust update system
```
motm32, line 121: Is this something we are trying to add in the future?

motm32 (approved):

src/commands/copilotOnRails/agentInstructions.ts
MicroFish91: Yeah more of a thing we should consider in a future release cycle

Issue #1578, opened 2026-07-23
Intermittent state synchronization issue causes approved project plan to remain in Planning status


v-ruizh:
**OS:** Win11 x64
**Build Version:** [0.12.7-alpha.0](https://github.com/microsoft/vscode-azureresourcegroups/actions/runs/29781520624)
**Regression:** Not a regression

**Repro Steps:**
1. Create a project using Copilot on Rails.
2. Allow Autopilot to review and approve the generated project plan.
3. Observe the scaffolding stage.

**Expect:**
- After approval, project-plan.md should be updated to an approved state.
- Scaffolding should recognize the approved plan and continue automatically.
- The workflow should not become blocked by a stale plan status.

**Actual:**
- Autopilot reports that the plan has been approved.
- The project-plan.md still shows Status: Planning.
- Scaffolding reports: No approved project plan found.
- The workflow remains stuck on the scaffolding screen.
    <img width="1911" height="998" alt="Image" src="https://github.com/user-attachments/assets/937c1702-6ef7-4679-a671-e478050f547a" />

**More Info:**
1. This issue is intermittent and not consistently reproducible.
2. Observed approximately 2 times during testing.
3. Manually reopening the plan and approving it again does not resolve the issue. The workflow continues to report that no approved plan exists and returns to the approval step.
4. Once the issue occurs, the workflow appears to enter an approval loop where the plan can be approved repeatedly, but the approval state is never recognized by the next stage, preventing scaffolding from proceeding.

MicroFish91:
Any idea if clicking the "Something went wrong? Click here to resume." helps to unblock Copilot?

v-ruizh:
> Any idea if clicking the "Something went wrong? Click here to resume." helps to unblock Copilot?

Clicking the "Something went wrong? Click here to resume." can't fix this issue.
<img width="1867" height="991" alt="Image" src="https://github.com/user-attachments/assets/60b89b6b-8465-42b5-b227-e860048e9a06" />

v-ruizh:
Verified on [0.12.7-alpha.0](https://github.com/microsoft/vscode-azureresourcegroups/actions/runs/30405592460), the issue has been fixed.

PR #1619, opened 2026-07-28
Write the status for approved project plan

motm32:
This is an attempt to help mitigate this issue: https://github.com/microsoft/vscode-azureresourcegroups/issues/1578.

After approval we will attempt to write the `Status: Approved` ourseleved instead of relying on the agent to do it. In the case we fail to do this it will fall back to the default behavior we have in there at the moment.

### DIFF
--- a/src/webviews/copilotOnRails/extension/controllers/ScaffoldPlanViewController.ts
+++ b/src/webviews/copilotOnRails/extension/controllers/ScaffoldPlanViewController.ts
@@ -16,10 +16,12 @@ import { ext } from "../../../../extensionVariables";
 import { CopilotOnRailsContext } from "../../../../utils/copilotOnRails/CopilotOnRailsContext";
 import { callWithDiagnosticsAndTelemetryHandling, corId, setCorProp } from "../../../../utils/copilotOnRails/telemetryUtils";
 import { type PreviewPage, type ScaffoldPlanData } from "../../views/utils/parseScaffoldPlanMarkdown";
+import { ProjectPlanStatus } from "../../views/utils/projectPlanStatus";
 import { AUTOPILOT_QUERY_MARKER, disableAutopilot, enableAutopilot, getEffectiveMaxRequests, raiseWorkspaceMaxRequests } from "../autopilot";
 import { getCopilotOnRailsBundleLocation } from "../copilotOnRailsBundleLocation";
 import { openLoadingView } from "../openLoadingView";
 import { suppressTrackedViewCloseOnce } from "../projectSession";
+import { writeProjectPlanStatusAtUri } from "../utils/planStatus";
 import { PREVIEW_FOLDER_RELATIVE_PATH, readPreviewPages, type PreviewPagesResult } from "../utils/previewPagesReader";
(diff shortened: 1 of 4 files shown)


src/webviews/copilotOnRails/extension/controllers/ScaffoldPlanViewController.ts
```diff
@@ -167,9 +169,17 @@ export class ScaffoldPlanViewController extends WebviewController<Record<string,
...
         setCorProp(context, ENSURE_AGENT_INSTRUCTIONS_KEY, true);
 
-        const planBeforeAutopilot = confirmedAutopilot
-            ? await this.recordAutopilotMode()
+        const planBeforeApproval = this.sourceFileUri
+            ? await writeProjectPlanStatusAtUri(this.sourceFileUri, ProjectPlanStatus.approved)
             : undefined;
+        setCorProp(context, 'statusWriteSucceeded', planBeforeApproval !== undefined);
+        if (planBeforeApproval !== undefined) {
+            this.onSelfWrite?.(planBeforeApproval.replace(/^(\*\*Status\*\*:[ \t]*)([^\r\n]*)/m, `$1${ProjectPlanStatus.approved}`));
```
MicroFish91: Is this the same regex as `STATUS_LINE_REGEX`? If so, might be worth bring that over here as well.  That would be good also because we know that regex has test coverage via your new tests.

MicroFish91 (approved):

src/webviews/copilotOnRails/extension/controllers/ScaffoldPlanViewController.ts
```diff
@@ -180,9 +190,10 @@ export class ScaffoldPlanViewController extends WebviewController<Record<string,
         if (!(await launchAgentChat(azureProjectScaffoldAgent, confirmedAutopilot ? `${AUTOPILOT_QUERY_MARKER} ${baseQuery}` : baseQuery))) {
             if (confirmedAutopilot) {
                 await disableAutopilot();
-                if (planBeforeAutopilot !== undefined && this.sourceFileUri) {
-                    await vscode.workspace.fs.writeFile(this.sourceFileUri, Buffer.from(planBeforeAutopilot, 'utf-8'));
-                }
+            }
+            if (planBeforeApproval !== undefined && this.sourceFileUri) {
+                await vscode.workspace.fs.writeFile(this.sourceFileUri, Buffer.from(planBeforeApproval, 'utf-8'));
```
MicroFish91, line 199: Any idea why there is a second writeFile happening here? Was it not already written in the function call above?

motm32: It's a undo operation. So basically if for some reason scaffolding fails to start there is no way for the user to go back and re approve the plan. Otherwise when approved is in the plan file the Approve Plan is greyed out so they can't re run that operation.

MicroFish91: Ahh I see, could you leave a comment above the section in charge of undoing, just so it's easier to quickly parse

MicroFish91 (approved):

Issue #1622, opened 2026-07-29
[CoR]No validation for invalid subscription input from chat during deployment planning


v-amyang:
**OS:** Win11 x64
VS Code Insider: 1.132.0-insider
**Build Version:** [0.12.7-alpha.0](https://github.com/microsoft/vscode-azureresourcegroups/actions/runs/30405592460)
**Regression:** Not a regression

**Repro Steps:**
1. Start a deployment workflow
2. When prompted for the Azure subscription in chat, enter an arbitrary invalid value (for example, "test-subscription")
3. Continue through the deployment planning process
4. Approve the deployment plan and proceed with deployment

**Expect:**
The workflow should validate the subscription input and reject invalid values

**Actual:**
The invalid subscription input is accepted and the workflow continues without any validation or error message.
<img width="997" height="674" alt="Image" src="https://github.com/user-attachments/assets/e735daa9-a230-4563-a516-89b52b1f5c35" />
<img width="486" height="588" alt="Image" src="https://github.com/user-attachments/assets/24ed116d-257e-4e83-9d62-58572f3f6914" />

**More Info:**
1. After entering "test-subscription", the value appeared as the first option and selected value in the Subscription dropdown on the Deployment Plan page.
2. During deployment, the workflow used the current Azure subscription (AzCode2605) instead of the value entered by the user.
3. No validation message was shown and the workflow did not indicate that the entered subscription value was invalid.

PR #1627, opened 2026-07-29
Deployment plan view fixes 

motm32:
Fixes: https://github.com/microsoft/vscode-azureresourcegroups/issues/1622

I also noticed the subscription and location had not been properly being shown on the view even if they have been chosen in the chat and show up in the plan.md. I tested this and it seemed to fix it.


src/webviews/copilotOnRails/extension/openDeploymentPlanView.ts
```diff
@@ -5,8 +5,8 @@
 
 import * as vscode from "vscode";
 import { ext } from "../../../extensionVariables";
-import { CopilotOnRailsContext } from "../../../utils/copilotOnRails/CopilotOnRailsContext";
 import { DEPLOYMENT_PLAN_FILE_GLOB } from "../../../tree/project/projectPlanFiles";
+import { CopilotOnRailsContext } from "../../../utils/copilotOnRails/CopilotOnRailsContext";
```
nturinski, line 11: Any idea why this got moved? I don't see any obvious difference

src/webviews/copilotOnRails/views/utils/parseDeploymentPlanMarkdown.ts
```diff
@@ -79,10 +80,27 @@ export function parseDeploymentPlanMarkdown(markdown: string): DeploymentPlanDat
         { name: 'East US 2', code: 'eastus2' },
         { name: 'West US', code: 'westus' },
         { name: 'West US 2', code: 'westus2' },
+        { name: 'West US 3', code: 'westus3' },
```
nturinski: I know that it's been like this for a while, but I feel like we should get this static list from the locations step or something similar in the tools package rather than managing it

nturinski (dismissed):

src/webviews/copilotOnRails/extension/openDeploymentPlanView.ts
motm32: Probably just formatting differences I'll move it back

MicroFish91 (approved):
Can also wrap with `corId('deploymentPlan.getLocations')`

MicroFish91 (dismissed):

motm32:
Making more changes so waiting to get those into this PR before merging

motm32:
The above changes include some updates to the view to better align with what the azure-prepare skill flow looks like. I also removed the already approved plan functionality as it was super buggy since the status could be different with each iteration which made it so users could not continue.

PR #1615, opened 2026-07-27
🤖 🚆 Copilot on Rails 

MicroFish91:
## Status

Feel free to start reviewing this if you haven't already. We do have a few open PRs to merge still linked below. There shouldn't be that many of them, but if you see some files that don't belong, it's probably because we haven't synced with main in a little while. For example I cherry-picked a couple PRs from main to be able to grab the mcp updates, which is why you may see some unrelated activity log mcp tooling showing up in the diff. I'll sync up a merge once we get more of the remaining PRs added in.

## Overview

Copilot on Rails is a guided, end-to-end project creation flow for building an Azure-ready application without leaving
VS Code. A set of custom chat agents walks the user through three phases — planning & scaffolding the
project, setting up local development and debugging, and preparing deployment to Azure.  The first two phases are handled by us, the last phase is handed off to the GitHub Copilot for Azure deployment skills. 

The extension puts rails around that flow with embedded webviews for reviewing and approving each plan, MCP tools the agents call to hand control back to us, and a new **Azure Project** Explorer pane that tracks progress across the stages.

## Table of Contents

This PR is large, so here's a guided map of the main areas of change. Reviewing the groups below should give the clearest picture of how Copilot on Rails is wired together.

| # | Files | What's in it |
| - | ----- | ------------ |
| 1 | [`resources/agents`](https://github.com/microsoft/vscode-azureresourcegroups/tree/feat/CoR/resources/agents) | Custom agents + their instructions and shared references |
| 2 | [`src/commands/copilotOnRails`](https://github.com/microsoft/vscode-azureresourcegroups/tree/feat/CoR/src/commands/copilotOnRails) | Full list of Copilot on Rails commands, grouped by phase |
| 3 | [`src/chat/tools/copilotOnRails`](https://github.com/microsoft/vscode-azureresourcegroups/tree/feat/CoR/src/chat/tools/copilotOnRails) | MCP tools chat calls back into when phase work completes |
| 4 | [`src/tree/project`](https://github.com/microsoft/vscode-azureresourcegroups/tree/feat/CoR/src/tree/project) | New `azureProject` Explorer pane that tracks stage progress |
| 5 | [`src/webviews/copilotOnRails/extension`](https://github.com/microsoft/vscode-azureresourcegroups/tree/feat/CoR/src/webviews/copilotOnRails/extension) | Extension-side controller logic that the webviews message |
[...]

PR #1628, opened 2026-07-29
CoR: Remove `onStartupActivation` by making `Azure Project` view always visible but collapsed

MicroFish91:
Fixes: https://github.com/microsoft/vscode-azureresourcegroups/pull/1615#discussion_r3667849975

### Changes
- Azure project view will no longer conditionally appear based on immediate activation and FS read, which was previously a design choice that prevented us from lazy activation
- Now that Azure project is always showing up, we need to handle how it behaves when a non-CopilotOnRails project is already open
  - If not empty, modal to prompt user to pick a new empty project to start with
  - Then automatically resume from there in the new empty project

### Flow
1. Click create with a non-empty workspace:
  image
2. Browse
  image
3. Re-opens and immediately starts the flow:
  image

### DIFF
--- a/src/tree/project/registerProjectSubmissionStateWatcher.ts
+++ b/src/tree/project/registerProjectSubmissionStateWatcher.ts
@@ -1,0 +1,27 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { getProjectPlanFiles, type ProjectPlanFilesWatcher } from './projectPlanFiles';
+import { projectSubmissionState } from './projectSubmissionState';
+
+/**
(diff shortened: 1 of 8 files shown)


src/tree/project/registerProjectSubmissionStateWatcher.ts
```diff
@@ -0,0 +1,27 @@
...
+import { getProjectPlanFiles, type ProjectPlanFilesWatcher } from './projectPlanFiles';
+import { projectSubmissionState } from './projectSubmissionState';
+
+/**
+ * Watches for project plan files to appear, and clears the pending submission
+ * state once they do. That state exists only to show progress in the gap between
+ * the user submitting and the first file being written, so it's no longer needed
+ * once the files are on disk.
+ */
+export function registerProjectSubmissionStateWatcher(context: vscode.ExtensionContext, planFilesWatcher: ProjectPlanFilesWatcher): void {
```
MicroFish91, line 16: This is the same logic that was previously in `extension.ts`.  I moved it to its own file and renamed / improved the function description

alexweininger (approved):

PR #1645, opened 2026-08-03
scaffold: document monorepo self-contained deploy; bundling optional

nturinski:
Keep the idiomatic modular per-file + glob-main Functions layout as the default. Fix the monorepo deployment failures at the deployment layer instead of switching to bundling:

- Preferred shared fix: import shared via relative paths so tsc compiles it into the service dist (no workspace-package runtime dependency to resolve in a per-service deploy).

- Resolve hoisted third-party deps with a service-level production install.

- Document esbuild --packages=external single-file bundling as an OPTIONAL cold-start/single-artifact optimization, not the default.


MicroFish91 (approved):

motm32 (approved):

MicroFish91:
Is this no longer needed now that you have the updated skills entrypoint?

nturinski:
I think it's probably still worth having just so that the scaffolder is aware of the monorepo formatting.

PR #1676, opened 2026-08-14
Fix docker-compose credential corruption and blank local database credentials

alexweininger:
## Problem

Reference docs under `resources/` inlined concrete credentials — for example a literal `DATABASE_URL: postgresql://user:password@host:5432/db`. The `azure-debug-generate` skill then wrote the same shape into generated local dev files such as `docker-compose.yml`. This produced two distinct failure modes.

### 1. Fatal YAML corruption

A secret-redaction filter rewrites the credential literal to a masked value. In YAML a leading `*` is an **alias indicator**, so the moment a redacted value lands in `docker-compose.yml` the file stops parsing entirely.

This is why redaction is fatal here rather than merely cosmetic: the masked text isn't just an unusable value, it's a syntax error. Local dev is dead on arrival, and the error the user sees points at YAML syntax — not at the credential that caused it.

### 2. Silent blank credentials

Where compose *did* use `${POSTGRES_USER}` / `${POSTGRES_PASSWORD}`, the skill often generated no workspace-root `.env`. Docker Compose interpolates `${...}` from `.env` or the shell environment — **never** from a service's own `environment:` block. With no `.env`, the variables resolve to empty strings and Postgres starts with blank credentials, producing auth failures that look like application bugs.

## Fix

The guidance now:

- Declares local credentials **once** in a workspace-root `.env`, and references them via `${...}` everywhere else.
- Forbids inlining a concrete `user:password@host` URL into any generated file, and explains the YAML-alias rationale so the rule isn't mistaken for style advice.
- Requires `.env` to exist and to declare **every** variable compose references — including a note that `.env.example` is documentation only and nothing loads it.
- Documents which host to use in a connection string: `localhost` for clients running on the host machine (VS Code debug target, `npm` task) vs. the `postgres` service name for clients running inside compose.
- Notes that `local.settings.json` is plain JSON with no `${...}` interpolation, and that a runtime settings file does not cover host-run tasks such as `npm run db:migrate`.

## `.env` must be gitignored, and secrets belong in no commit

Per review feedback: the guidance told the agent to create a workspace-root `.env` holding real credential values, but never said it must be ignored.

`.gitignore` must now list `.env` before the file is created, stated at each point where it actually gets written (`emulators/postgres.md`, `references/generate.md`), with an explicit instruction to fix a missing or incomplete `.gitignore` first. `shared-references/architecture.md` § .gitignore Additions carries the rule in full, since that section defines the generated `.gitignore` contents: a secret that reaches a commit is compromised the moment it is pushed, deleting it in a later commit does **not** undo that because the value stays in history and in every clone and fork, recovery means rotating the credential rather than editing a file, and private repositories are no exception.

`azure-project-scaffold/instructions.md` makes the ordering explicit — `.gitignore` is written in Step 2 (Foundation), before Step 3 (Configuration & Environment) creates env files.

## Audit: no reference doc can seed a masked value

The remaining reference docs carrying a concrete credential literal are fixed too. This matters because a reference doc is read by the agent *before* it writes anything — the literal is masked in the agent's context by secret redaction and then transcribed verbatim into the generated project. **All concrete credential literals under `resources/` are now eliminated, so no reference doc can seed a masked value into a generated project.**

- `azure-project-plan/plan.md` — the two connection-string table rows are templates the planning agent fills in, so a masked value propagates into `.azure/project-plan.md` and downstream into generated code. Highest-impact of the set.
- `shared-references/resilience.md` and `shared-references/service-abstraction.md` — the sample test setup now builds the URL from `process.env` via a template literal (quoting changed from single quotes to backticks accordingly).

Verified with `git grep -lE '://[A-Za-z0-9_.-]+:[A-Za-z0-9_.!$-]+@' HEAD -- resources/` → zero files.

## Files changed

- `resources/agents/azure-debug-generate/references/emulators/postgres.md`
- `resources/agents/azure-debug-generate/references/generate.md`
- `resources/agents/azure-project-plan/plan.md`
- `resources/agents/azure-project-scaffold/instructions.md`
- `resources/agents/shared-references/architecture.md`
- `resources/agents/shared-references/resilience.md`
- `resources/agents/shared-references/runtimes/python.md`
- `resources/agents/shared-references/runtimes/typescript.md`
- `resources/agents/shared-references/service-abstraction.md`

Markdown instruction files only — no extension source changes, nothing to compile.

## Related

The accompanying evaluation-system detection for this defect ships separately in #1669.

### DIFF
--- a/resources/agents/azure-debug-generate/references/emulators/postgres.md
+++ b/resources/agents/azure-debug-generate/references/emulators/postgres.md
@@ -10,12 +10,12 @@ postgres:16
 
 ## docker-compose Service Block
 
-Declare the local credentials **once** in a workspace-root `.env` file, then reference them
-everywhere else. Docker Compose interpolates `${...}` from `.env` (or the shell environment),
-so the same values feed the database service and every application service.
+Declare the local credentials **once** in a workspace-root `.env`, then reference them everywhere
+else. Compose interpolates `${...}` from `.env` (or the shell environment), so the same values feed
+the database service and every application service.
 
-`.env` at the workspace root — this file is required; without it Compose interpolates
-`${POSTGRES_USER}` to an empty string and the database silently fails to authenticate:
+`.env` is required — without it Compose interpolates `${POSTGRES_USER}` to an empty string and the
+database silently fails to authenticate:
 
 ```
 POSTGRES_USER=postgres
@@ -23,11 +23,15 @@ POSTGRES_PASSWORD=postgres
 POSTGRES_DB=localdev
 ```
 
-> **Never inline a concrete `user:password@host` URL** in a generated file. Build the URL from
-> the discrete variables above. Besides hard-coding credentials, concrete credential literals are
-> rewritten by secret-redaction filters, which silently corrupts the generated file — a masked
-> value starting with `*` is a fatal YAML parse error, because YAML reads a leading `*` as an
-> alias reference.
+> ⛔ **`.gitignore` must list `.env` before you create it.** A credential that reaches a commit is
+> compromised and has to be rotated — deleting it in a later commit leaves the value in history and
+> in every existing clone and fork. Private repositories are no exception. If the project has no
+> `.gitignore`, or has one that omits `.env`, fix that before writing the file.
+
+> **Never inline a concrete `user:password@host` URL** in a generated file — build it from the
+> variables above. Beyond hard-coding a credential, such literals are rewritten by secret-redaction
+> filters, and a masked value starting with `*` is a fatal YAML parse error: YAML reads a leading
+> `*` as an alias reference.
 
 ```yaml
 services:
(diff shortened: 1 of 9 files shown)


alexweininger:
I ran into this while testing. Verified that my change does fix it. However, I am worried that what Copilot wrote here is too verbose

resources/agents/azure-debug-generate/references/emulators/postgres.md
```diff
@@ -10,20 +10,39 @@ postgres:16
 
 ## docker-compose Service Block
 
+Declare the local credentials **once** in a workspace-root `.env` file, then reference them
```
bwateratmsft: Add that `.env` must be `.gitignore`'d. Maybe extra emphasis here and/or elsewhere that secrets anywhere in committed source are completely unacceptable--doesn't matter if open or closed source. Secrets cannot exist in any commits at all, and removing them and then recommitting is not good enough.

alexweininger: Added — good catch, the guidance was creating a file full of real credential values without ever saying to ignore it.

`.gitignore` must now list `.env` before it is created, stated at each point where the file actually gets written (`emulators/postgres.md`, `references/generate.md`), with an explicit instruction to fix a missing or incomplete `.gitignore` first.

On the broader emphasis: `shared-references/architecture.md` § .gitignore Additions now carries the rule in full — a secret that reaches a commit is compromised the moment it is pushed, deleting it in a later commit does not undo that because the value stays in history and in every clone and fork, recovery means rotating the credential rather than editing a file, and private repos are no exception. I put it there because that section is what actually defines the generated `.gitignore` contents, so the scaffold agent reads it at the point of use.

I also made the ordering explicit in `azure-project-scaffold/instructions.md`: `.gitignore` is written in Step 2 (Foundation) and env files arrive in Step 3 (Configuration & Environment), so "ignored from the first commit" is already achievable as the steps are sequenced — it just was not spelled out.

Separately, on Alex's verbosity note: I trimmed the prose I had added earlier in this PR, which ran long without adding information. Net length is roughly unchanged despite the new rule.

bwateratmsft (approved):

MicroFish91 (approved):
LGTM

> I ran into this while testing. Verified that my change does fix it. However, I am worried that what Copilot wrote here is too verbose

Yeah I noticed this too.  I think it's because it's reiterating the same reminders across multiple agents.  Ideally I would just like the agent responsible for creating the `.env` the first time to add it to `.gitignore` and not have to keep reiterating / reminding the later agents about it.  In practice though I know an agent could accidentally skip and I could see the benefit of having those reminders as fallbacks in later phases, but that does come at the cost of instruction bloat.

PR #1680, opened 2026-08-17
CoR: Ensure we are using the local harness during runs

MicroFish91:
Closes https://github.com/microsoft/azcode-internal/issues/270


nturinski:
Do we know if they're getting rid of the local harness eventually?

MicroFish91:
No idea, we'll need to be on the lookout for that

PR #1683, opened 2026-08-17
Add vally framework and graders for azure-project-plan

motm32:
## Add a Vally eval harness for the `azure-project-plan` agent

Adds an automated evaluation suite that runs the **shipped** agent instructions against a set of realistic prompts and grades the artifacts it produces, so regressions in `resources/agents/**` are caught in CI instead of by hand.

### Main changes

- **Eval harness** (`evals/`) — a Vally suite (`evals/project-plan/eval.yaml`) with 7 stimuli covering requirements gathering (`photo-app-requirements`, `api-only-inventory`, `multi-service-order-processing`, `no-datastore-converter`) and plan generation/approval/feedback (`plan-generation-task-app`, `plan-approval-scrapbook`, `plan-feedback-recipe-app`).
- **Fidelity by construction** — `evals/executor/agent-assets.mjs` generates the eval skill directly from `azure-project-plan.agent.md` (carrying `name`/`description` verbatim so activation is tested for real) and recursively copies the whole agent folder plus `shared-references/` into the workspace. Nothing is hand-restated, so the evals can't drift from what users receive or "teach to the test."
- **Stub MCP gate tools** (`evals/mcp/`) — a stdio server exposing the 12 `workflow-tools-*` webview/gate tools so tool-call contracts are observable outside VS Code. The tool list lives in `workflow-tools.mjs` so it can be imported without starting a server.
- **Drift guard** (`evals/check-agent-drift.mjs`) — asserts 12 agent contracts plus cross-file consistency, and pins a SHA-256 hash of `resources/agents/**` in `agent-assets.lock.json`. Editing agent instructions fails CI until the evals are re-run and the lock is updated with `--update`.
- **CI** (`.github/workflows/vally-evals.yml`) — runs the suite and the drift check; the Vally CLI is pinned via `npm ci` in `evals/` instead of an unpinned global install.
- **Dependency hygiene** — Vally stays out of the extension manifest (`evals/` owns its own lockfile); root only gains `typescript-eslint` and an `eval:drift` script. `evals/` is excluded from the VSIX.

### Graders

Built-in graders cover artifact presence/absence and tool calls (`file-exists`, `file-not-exists`, `file-matches`, `tool-calls`, `transcript-not-contains`), plus three custom program graders:

| Grader | Checks |
|---|---|
| `validate-requirements.mjs` | `.azure/requirements.json` parses and matches the schema; datastore rules (e.g. no datastore invented for a stateless converter, Blob Storage handled correctly) |
| `validate-project-plan.mjs` | `.azure/project-plan.md` has the required skeleton — `Status`/`Created` metadata rows, mandated sections and ordering |
| `validate-webview-parseable.mjs` | the generated plan is actually parseable by the extension's `parseScaffoldPlanMarkdown` logic, so the plan webview can render it |

These also enforce negative contracts: no premature `project-plan.md` during requirements, no dotfile `.requirements.json`, no chat questions where the UI should be used, no inline scaffolding during planning.

### Report changes

`evals/generate-report.cjs` emits a `run-diagnostics.md` + `run-diagnostics.json` per run containing:

- a run diagnosis header — outcome, summary, **primary failure**, classification, error and observed issue
- suite and per-stimulus result tables, a full grader matrix, and per-stimulus detail (skills invoked, tool calls, files touched)
- failing command detail with exit code, stderr and stdout
- derived **recommended actions** and a full evidence file list
- failures are attributed to the layer that owns the fix: `agent_failure`, `product_failure`, `harness_failure`, or `infrastructure_failure` (upstream 5xx/429/auth/network errors are marked **inconclusive**, excluded from blame, and never allowed to headline a run that also contains a real defect)


.github/workflows/vally-evals.yml
```diff
@@ -0,0 +1,156 @@
+name: Vally Agent Evals
+
+on:
+  # Run per-agent evals on PRs that touch agent instructions
+  pull_request:
+    branches: [main]
```
alexweininger: we should make this run on the feat/CoR branch too

.github/workflows/vally-evals.yml
```diff
@@ -0,0 +1,156 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          # Most recent artifact from main branch — you'd configure this
+          # with a separate baseline-publish step or use a fixed artifact name
+          name: vally-baseline-main
+          path: results/main
+        continue-on-error: true
+
+      - name: Compare (fail on regression)
+        if: steps.download-baseline.outcome == 'success'
```
alexweininger, line 139: `steps.download-baseline` doesn't resolve to anything — the download step above has `continue-on-error: true` but no `id:`, and there's no `id:` anywhere in this file. The expression is always empty, so `vally compare --fail-on-regression` never executes and the regression gate is off.

Same theme elsewhere in the PR: `eval:cor:graders:certify` and `eval:cor:thresholds:validate` are defined as root scripts (`package.json:1027-1028`) but no job calls them, so grader certification and threshold validation don't run either.

`id: download-baseline` on the step above fixes this one; the other two want a step in the contracts job.

evals/run-eval.cjs
```diff
@@ -0,0 +1,108 @@
...
+        console.log(`\n[run-eval] Results saved to: ${outputDir}`);
+        // Generate readable diagnostics from the latest timestamped subfolder
+        try {
+            const subs = require("fs").readdirSync(outputDir).filter(d => d.match(/^\d{4}-/)).sort();
+            if (subs.length) {
+                const latest = path.join(outputDir, subs[subs.length - 1]);
+                require("child_process").execFileSync("node", [path.join(evalsDir, "generate-report.cjs"), latest], { stdio: "inherit" });
+            }
+        } catch { /* best-effort */ }
+        process.exit(code || 0);
```
alexweininger: This is where the run's verdict comes from, and it's Vally's raw exit code — `classify()` in `generate-report.cjs` never gets a say. The classification is computed afterwards, written into markdown, and discarded.

Concretely: a 429 or a GitHub outage gets correctly labelled `infrastructure_failure` in the report and still fails CI as though the agent regressed. `generate-report.cjs:229` has the same split — it prints `passed.length/results.length` with infra-failed trials in the denominator, directly next to prose calling them inconclusive.

This one bit us four separate times on the other branch; an outage once scored the product at 0.0%. Deriving the exit code from `run-diagnostics.json` and excluding `infrastructure_failure` from both numerator and denominator would close it. `evals/src/report.ts:245-264` on `copilot-on-rails-evaluation-system` has the classification logic if it's useful.

Separately: the report generation just above is inside `catch { /* best-effort */ }`, so if the classifier itself throws you get no diagnosis at all and just the bare failure.

evals/generate-report.cjs
```diff
@@ -0,0 +1,543 @@
...
+function classify(result, grader) {
+    if (isInfrastructureError(result)) {
+        return { classification: 'infrastructure_failure', reason: 'An upstream service (auth, model endpoint, or network) rejected the request before the agent ran. Re-run when the service recovers.' };
+    }
+    if (result.status && result.status !== 'success') {
+        return { classification: 'harness_failure', reason: `Trial status was \`${result.status}\` — the executor did not complete the run.` };
+    }
+    const exitCode = grader?.metadata?.exit_code;
+    const stderr = grader?.metadata?.stderr ?? '';
+    if (exitCode !== undefined && exitCode !== 0 && exitCode !== 1) {
```
alexweininger, line 215: Exit code 1 is excluded here, but that's what Node returns on an uncaught exception — and it's also what the graders deliberately return when the product fails (`validate-requirements.mjs:166`). So a `TypeError` inside a grader is scored as a product defect.

The stderr check below catches `SyntaxError`, which is a parse-time failure, but not `TypeError` or `ReferenceError` — the ones you'd actually hit at runtime.

A top-level try/catch in each grader exiting a reserved code (3, say) and classifying that as harness here is probably the cheapest fix.

evals/generate-report.cjs
```diff
@@ -0,0 +1,543 @@
...
+report.push('## Full evidence\n');
+for (const file of evidenceFiles()) {report.push(`- \`${file}\``);}
+if (runSummary?.sessionLogsDir) {report.push(`- \`${runSummary.sessionLogsDir}\` (session logs)`);}
+report.push('');
+
+const reportPath = path.join(resultsDir, 'run-diagnostics.md');
+fs.writeFileSync(reportPath, report.join('\n'));
+
+// Machine-readable counterpart so CI can assert on the same diagnosis.
+const diagnostics = {
```
alexweininger, line 585: `run-diagnostics.json` has no `schemaVersion`, which will bite once we're aggregating across runs — `graderCertification.ts` and `release-thresholds.v1.json` in this same PR both version their artifacts.

The bigger risk is that the reader degrades silently rather than loudly. Line 35 is `results.filter(r => r.gradeResult?.passed)` and the scores are `?? 0` throughout. If Vally renames or nests `gradeResult`, every stimulus becomes a silent 0% and the report will state with confidence that the product regressed to nothing.

We hit exactly this on the other branch when gate explanations moved from `evidence` to `reason` — 30 results were misclassified before anyone caught it; there's a note about it at `evals/src/gateHealth.ts:108-117`. Stamping a version here and throwing on a missing `gradeResult` instead of coercing to false would make the failure loud. Worth noting line 527 reads `grader.evidence`, which is the older of the two shapes.

evals/project-plan/eval.yaml
```diff
@@ -0,0 +1,224 @@
...
+  weights:
+    program: 0.3
+    tool-calls: 0.3
+    file-exists: 0.15
+    file-not-exists: 0.1
+    file-matches: 0.05
+    file-not-matches: 0.04
+    transcript-not-contains: 0.03
+    output-not-contains: 0.03
+  threshold: 0.9
```
alexweininger: Weighting by grader type lets a low-weight grader fail without failing the stimulus. `plan-approval-scrapbook` is the clearest case — its three graders are `file-matches` (0.05), `tool-calls` (0.3) and `output-not-contains` (0.03), totalling 0.38. If `no-inline-scaffolding` fails, the score is 0.35/0.38 = **0.921**, which clears this threshold. The agent inlines scaffolding, a contract the PR description says we enforce is violated, and the stimulus is green.

`photo-app-requirements` has the same shape: `no-chat-questions` failing gives 0.95/0.98 = **0.969** and still passes.

It also gets worse as gates land — at 17 gates a single failure is roughly 6% of total weight and can never cross 0.9. Treating contract graders as required (any failure fails the stimulus) and keeping the weighted score for genuinely graded dimensions would hold up as this grows.

evals/src/graderCertification.ts
```diff
@@ -0,0 +1,225 @@
...
+): Promise<Map<string, string[]>> {
+    const azure = path.join(workspace, '.azure');
+    const [requirements, projectPlan] = await Promise.all([
+        fs.readFile(path.join(azure, 'requirements.json'), 'utf8'),
+        fs.readFile(path.join(azure, 'project-plan.md'), 'utf8'),
+    ]);
+    const validations: Array<readonly [string, ArtifactValidationResult]> = await Promise.all([
+        Promise.resolve(['requirements', validateRequirementsArtifact(requirements, { requireConfirmed: true })] as const),
+        Promise.resolve(['project-plan', validateProjectPlanArtifact(projectPlan, { expectedStatus: 'Integrated' })] as const),
+        Promise.resolve(['plan-gate', validatePlanEvaluationContract(true, true, {
```
alexweininger: `plan-gate` can't fail. Every argument here is a literal, and walking the three checks in `planEvaluation.ts` against them:

- `expectedFrontend !== generatedFrontend` → `true !== true` → never fires
- `expectedFrontend && !previewManifestPresentAtCall` → `true && !true` → never fires
- `expectedFrontend && previewHtmlFilesAtCall?.length` → `true && 0` → never fires

So it returns `valid: true` with zero issues regardless of the fixture. It reads nothing from the workspace, which means no mutation to any file could falsify it. It's listed in `manifest.json:7-12` with the other three validators, and it's the one with no mutation — consistent with there being no mutation that could work.

The bigger issue is what certification covers at all. These validators are `evals/src/artifacts/*.ts`, but the suite runs `evals/graders/*.mjs`, which import only `node:fs` and `node:path` — they're independent reimplementations, not wrappers. So the certified code isn't the executed code, and the two are already out of sync:

- `requirements-schema-version` asserts code `schemaVersion` (`requirements.ts:33-34`); `validate-requirements.mjs` never mentions `schemaVersion`.
- `project-plan-numbering` asserts `nonSequentialHeading` (`projectPlan.ts`); `validate-project-plan.mjs` never mentions it.

Both things certification proves are things the running grader can't do. Downgrade `schemaVersion` to `"1"` in a real run and the suite passes it.

We prefer TS anyway, so I'd collapse this rather than wrap it: drop the `.mjs` graders and make them `.ts`. Node 22.18+ runs `.ts` directly, so `eval.yaml` keeps `program: node` and only the filenames change — no tsx, no build step. Worth adding `"node": ">=22.18"` to `engines` since there's no floor declared today. Each grader then becomes a thin adapter: parse argv, call the validator in `src/artifacts`, map returned issues to an exit code. One implementation, and certification finally covers the path that runs.

While you're in there, have the adapter exit 3 when the grader itself throws, so it's distinguishable from 1 (product failed) — that's the split the report can't currently make.

evals/check-agent-drift.mjs
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * Each contract is a rule the graders rely on. `pattern` must still match the
+ * shipped file; `grader` names what breaks when it doesn't.
+ */
+const contracts = [
+    {
+        file: `${PLAN}.agent.md`,
+        name: "skill-frontmatter",
+        pattern: /^---\r?\n[\s\S]*?^name:\s*azure-project-plan\s*$[\s\S]*?^description:\s*\S[\s\S]*?^---/m,
+        grader: "evals/executor/agent-assets.mjs generates SKILL.md from this frontmatter",
```
alexweininger: This is the part of the PR I'd most like to keep. Tying each contract to the grader that breaks when the instruction text changes means a deleted sentence fails fast and names its own consequence, instead of turning up as a confusing eval failure 20 minutes into a run.

The eval branch I've been building (`copilot-on-rails-evaluation-system`) has nothing equivalent — it checks agent assets for secret literals and reference reachability, but nothing maps a shipped instruction to the grader that depends on it. I'm planning to port this in the other direction.

One thing to think about as it grows: the lock hashes all of `resources/agents/**`, so once the other agents have suites, edits to unrelated agents will trip it and `--update` becomes reflexive. Per-agent hashes would keep the signal.

evals/project-plan/eval.yaml
alexweininger: Copilot is super long winded here, but I think that using a weighted grader for a stimulus is a bad idea, at least for now. 

I think for now it should just be pass/fail, and note which step in the process it failed at. E.g. if stimulus A. failed to build (early step) vs. stim B. failed to deploy (late step), then stim B was more successful, but still not a pass.

We can figure out the exact details of how to measure it later on once we have more graders

alexweininger (commented):
we can try to tackle this later, but it'd be really nice if we could use TS files intead of .mjs

alexweininger (approved):

PR #1687, opened 2026-08-19
Add a deployment results view 

motm32:
Now with the new deployment changes there is a deploy-result.json that gets created. This adds a view for users to look at those results more easily. Here is how the view looks: 

https://github.com/user-attachments/assets/e82b2e57-263c-437a-abf3-683e3b0689df

I haven't fully tested the flow yet but feel free to give feedback on the design of the view.


nturinski:
Looks great! Just a few notes discussed offline:

- Weird bug where deploy view opened twice. Can possibly fix that in the agent and instruction files
- Make the approve from deploy view continue in the same session rather than starting a new session
- Expand cost in deploy view by default

nturinski (approved):

PR #1693, opened 2026-08-25
Add initial set of scaffold evals 

motm32:
## What this does

Adds evals for the `azure-project-scaffold` agent that run in MSBench.

The evals check the two things that matter: that the agent **stops at the right
gates**, and that the project it produces **builds and the frontend works**.

## How it works

- **5 scaffold suites** cover the gate decisions — missing plan, unapproved plan,
  fullstack (opens preview gate), API-only, and autopilot (skips the gate).
  `fullstack` and `autopilot` share one plan and differ only in the prompt prefix,
  so an agent that always opens the gate — or never does — fails exactly one.
- **No checked-in fixtures.** The scaffold agent needs a `.azure/project-plan.md`
  to start from, so 2 seed suites run the real planner in the same harness and
  `harvest-seed.mjs` extracts the result.
- **Graders are real checks**, not string matching — they validate the plan
  structure, run the build, and verify the frontend renders.

## Safety rails

- A seed is only promoted if its run passed every check, so a broken or
  rate-limited run can't quietly become a fixture.
- Staging refuses a seed harvested against different agent assets.
- Every suite carries a liveness sentinel — without one, a run that dies early
  scores partial credit because `COUNT(*) = 0` over an empty table is trivially true.
- `validate-suites.mjs` enforces this in CI.

## Notes

- Nothing runs automatically; these are submitted manually via
  `evals/msbench/run.sh`.
- Validated end to end: a real seed run produced a clean plan, and the harvester
  correctly refused a rate-limited one.


alexweininger:
Closing — this work has landed on `feat/CoR`.

The graders, artifact validators and certification fixtures from this PR were adopted in #1707 (scaffold gates) and #1708 (local-debug gates), ported onto the TypeScript `config/` layout introduced by #1696. Everything of substance survives with `Co-authored-by` attribution; what was dropped was only the plumbing that collided with the merged layout (`stage.sh`, `stage-graders.sh`, `preflight.sh`, `harvest-seed.mjs`, `stage.mjs`, `validate-suites.mjs`, the per-suite `user-overrides.yaml` layout, and the pre-conversion `.mjs` executor edits).

Grader certification went from 9 cases to **44** as a result, and the fixture split introduced here (`reference-node-fullstack` = the app you seed a workspace with, `sample-agent-output` = what the agent leaves behind, which is the thing you grade) was kept and extended — certification is now multi-fixture because of it.

Closing so the open-PR list reflects work actually outstanding.

PR #1694, opened 2026-08-25
Add initial set of local dev evals 

motm32:
## Add local-development phase evals

Extends the eval system to the phase after scaffold (`azure-debug-plan` → `azure-debug-generate`)

**Graders** — four validators (plan contract, `launch.json`/`tasks.json` structure, plan/artifact conformance, completed validation checklist) plus a gate check that planning doesn't generate early. Certification goes 9 → 26 cases.

**Real graders on MSBench** — `stage-graders.sh` bundles the actual grader sources into the container, so MSBench runs the same code Vally does. This replaces the SQL approximation of `requirements-schema-valid` that #1689 flagged as partial.

**`preflight.sh`** — verifies a suite offline before spending container time, checking each `exec:` assertion fails on a bare workspace *and* passes on a valid one so vacuous passes can't slip through.

⚠️ **`run.sh` now requires `./stage.sh <suite>` first** — `assets/user-overrides.yaml` is generated rather than tracked now that each suite owns its config. CI updated.

**Verification:** certify 26/26 · typecheck · eslint · lint · preflight clean on both suites.

**Known gaps:** checks are structural only — nothing boots an emulator yet (a Docker capability probe is included to scope that work). We would need to add a bit more within MSBench to test if the local development set up is working properly.

### DIFF
--- a/.github/workflows/msbench-evals.yml
+++ b/.github/workflows/msbench-evals.yml
@@ -99,6 +99,14 @@ jobs:
           tenant-id: ${{ secrets.MSBENCH_AZURE_TENANT_ID }}
           subscription-id: ${{ secrets.MSBENCH_AZURE_SUBSCRIPTION_ID }}
 
+      # assets/ is generated rather than tracked, so the suite config, the grader
+      # bundle and the workspace seed have to be assembled before run.sh reads them.
+      # preflight catches a broken config offline, before ~15 minutes of container time.
+      - name: Stage project-plan assets
+        run: |
+          ./evals/msbench/stage.sh project-plan
+          ./evals/msbench/preflight.sh project-plan
+
       - name: Run project-plan eval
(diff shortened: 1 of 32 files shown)


alexweininger:
Closing — this work has landed on `feat/CoR`.

The graders, artifact validators and certification fixtures from this PR were adopted in #1707 (scaffold gates) and #1708 (local-debug gates), ported onto the TypeScript `config/` layout introduced by #1696. Everything of substance survives with `Co-authored-by` attribution; what was dropped was only the plumbing that collided with the merged layout (`stage.sh`, `stage-graders.sh`, `preflight.sh`, `harvest-seed.mjs`, `stage.mjs`, `validate-suites.mjs`, the per-suite `user-overrides.yaml` layout, and the pre-conversion `.mjs` executor edits).

Grader certification went from 9 cases to **44** as a result, and the fixture split introduced here (`reference-node-fullstack` = the app you seed a workspace with, `sample-agent-output` = what the agent leaves behind, which is the thing you grade) was kept and extended — certification is now multi-fixture because of it.

Closing so the open-PR list reflects work actually outstanding.

PR #1697, opened 2026-08-25
CoR: Add safe install links for prerequisites and standardize format across all phases

MicroFish91:
- Use a hardcoded map of links so we can offer install links without risking prompt injection
- Add prerequisites and support for deploy, surface in the new deploy webview
- Fixed some tests that were already failing


nturinski:
@Copilot Why is the agent contracts failing exactly?

PR #1705, opened 2026-08-26
Fix the evals lint break left by the TypeScript conversion

alexweininger:
## What's broken

`npm run lint` is **currently red on #1615** — the `feat/CoR` → `main` integration PR — and has been at every recent `feat/CoR` commit (`2b1e6465`, `e4e719f1`, current HEAD). It is not caused by any of the in-flight eval PRs; they just inherit it. That's the most useful sentence in this PR.

Worth knowing where it does and doesn't bite: `main.yml` only triggers on PRs targeting `main`/`rel/*`, so PRs *into* `feat/CoR` never run it and look green. The breakage is therefore invisible on the feature PRs and only shows up on the integration PR — which is exactly why it has survived this long.

The cause is a one-line mismatch. `eslint.config.mjs` has an override that lints the eval harness loosely, because it was plain Node scripts rather than extension code:

```js
files: ['evals/**/*.{js,cjs,mjs}'],
```

#1696 then converted `evals/` from `.cjs`/`.mjs` to **TypeScript**. The glob didn't follow, so every converted file quietly fell through to the extension's strict ruleset. Under `eslint --max-warnings 0` that's an instant failure:

```
evals/check-agent-drift.ts               1 warning   __dirname naming-convention
evals/check-gate-tools.ts                1 warning   __dirname naming-convention
evals/ci-local.ts                        1 warning   __dirname naming-convention
evals/executor/azure-agent-executor.ts   2 warnings  __filename / __dirname
                                         1 error     Unexpected any
✖ 6 problems (1 error, 5 warnings)
```

## The fix, and why not just widen the glob

Adding `ts` to that glob would work, but the override also sets `disableTypeChecked` and `projectService: false` — so it would silently drop **type-aware** linting across all of `evals/`. That trades a visible failure for an invisible loss of coverage. Instead, both underlying causes are removed:

**1. Drop the ESM `__dirname` shim (4 files).**

```diff
-import { fileURLToPath } from "node:url";
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const scriptDir = import.meta.dirname;
```

`import.meta.dirname` has been native since Node 20.11 and we run 22. Node defines it as precisely `path.dirname(fileURLToPath(import.meta.url))`, so this is equivalent — it deletes the thing the linter objected to rather than suppressing the objection.

**2. Narrow the SDK event `any` (1 file).** `rawEvents` is already declared `unknown[]`, so `convertSdkEvent` was taking `any` purely to avoid narrowing. It now takes `unknown` and is narrowed **once**, at that boundary, to a structural type covering only the fields it actually reads.

## Is narrowing the `any` a behaviour change?

**Effectively no — it is type-level, with two deltas that are unreachable in practice.** Stating them explicitly rather than claiming "purely type-level":

1. **A new `if (!raw || typeof raw !== "object") return null;` guard.** For a primitive input, behaviour is identical (`.type` was `undefined`, hit `default`, returned `null`). For `null`/`undefined` the old code would have **thrown** a `TypeError` and now returns `null`. Events come straight from `session.on(event => rawEvents.push(event))`, so the SDK never delivers `null` here — and turning a crash into a skip is the safe direction regardless.

2. **`toolCallId: event.data?.toolCallId` gained `?? ""`**, required because the optional chain yields `string | undefined` against a required `string` field. This only differs when the SDK omits `toolCallId`, and downstream it cannot matter: the id is read behind a truthiness check (`if (... && e.data.toolCallId && ...)`), and the lookup map is only ever populated with truthy keys — so `get("")` and `get(undefined)` both miss and fall back to `"unknown"` identically.

No grading logic, event mapping, or control flow is otherwise touched.

## Verification

- `npm run lint` — clean (was 6 problems)
- `npm run build` — clean
- `npm run typecheck` in `evals/` — clean
- `node check-agent-drift.ts` — still green, *16 agent contracts intact*, which confirms `import.meta.dirname` resolves correctly at runtime under Node's type stripping

## Scope

Four files, all in `evals/`. No config change, no functional change, nothing to do with timeouts or stimuli. Split out of #1700 so each PR is a single review.

> Note: this does **not** block #1700 (the timeouts PR) — that one is green, because PRs into `feat/CoR` don't run `main.yml` at all. This PR is what turns #1615 green.

### DIFF
--- a/evals/check-agent-drift.ts
+++ b/evals/check-agent-drift.ts
@@ -23,13 +23,12 @@
 import { createHash } from "node:crypto";
 import * as fs from "node:fs";
 import * as path from "node:path";
-import { fileURLToPath } from "node:url";
 import { listEvalAssetFiles, readSupportedModels } from "./src/agent-definition.ts";
 
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const repoRoot = path.resolve(__dirname, "..");
+const scriptDir = import.meta.dirname;
+const repoRoot = path.resolve(scriptDir, "..");
 const agentsRoot = path.join(repoRoot, "resources", "agents");
-const lockPath = path.join(__dirname, "agent-assets.lock.json");
+const lockPath = path.join(scriptDir, "agent-assets.lock.json");
 const update = process.argv.includes("--update");
 
 const PLAN = "azure-project-plan";
@@ -277,7 +276,7 @@ const currentFiles = agentAssetFiles();
  * The eval spec must run the agent on a model the product actually ships it on.
  * Catching a bad pin here costs a second; catching it at trial time costs a run.
  */
-const evalSpecPath = path.join(__dirname, "project-plan", "eval.yaml");
+const evalSpecPath = path.join(scriptDir, "project-plan", "eval.yaml");
 try {
     const supported = readSupportedModels(repoRoot, PLAN);
     const spec = fs.readFileSync(evalSpecPath, "utf8");
--- a/evals/check-gate-tools.ts
+++ b/evals/check-gate-tools.ts
@@ -25,13 +25,12 @@ import { approveAll, CopilotClient, RuntimeConnection } from "@github/copilot-sd
 import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
-import { fileURLToPath } from "node:url";
 import { MCP_SERVER_NAME, workflowToolDefinitions } from "./mcp/workflow-tools.ts";
 
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const scriptDir = import.meta.dirname;
 
 function bundledCliPath() {
-    const pkgDir = path.resolve(__dirname, "node_modules/@github/copilot");
+    const pkgDir = path.resolve(scriptDir, "node_modules/@github/copilot");
     for (const candidate of ["npm-loader.js", "index.js"]) {
         const full = path.resolve(pkgDir, candidate);
         if (fs.existsSync(full)) {return full;}
@@ -48,7 +47,7 @@ function bundledCliPath() {
  * themselves would not.
  */
 function requiredToolNamesFromSpecs(): Set<string> {
-    const specDir = path.join(__dirname, "project-plan");
+    const specDir = path.join(scriptDir, "project-plan");
     const names = new Set<string>();
     let files: string[];
     try {
--- a/evals/ci-local.ts
+++ b/evals/ci-local.ts
@@ -30,11 +30,10 @@ import { spawnSync, type SpawnSyncOptions, type SpawnSyncReturns } from "node:ch
 import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
-import { fileURLToPath } from "node:url";
 import { parse as parseYaml } from "yaml";
 
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const repoRoot = path.resolve(__dirname, "..");
+const scriptDir = import.meta.dirname;
+const repoRoot = path.resolve(scriptDir, "..");
 const WORKFLOW = ".github/workflows/agent-contracts.yml";
 const IMAGE_NODE = "22";
 
--- a/evals/executor/azure-agent-executor.ts
+++ b/evals/executor/azure-agent-executor.ts
@@ -15,17 +15,15 @@ import { computeMetrics } from "@microsoft/vally";
 import type { Executor, ExecutorOptions, ExecutorRegistry, Stimulus, Trajectory, TrajectoryEvent } from "@microsoft/vally";
 import * as fs from "node:fs";
 import * as path from "node:path";
-import { fileURLToPath } from "node:url";
 import { buildEvalSkill, prepareAgentWorkspace, resolveEvalModel } from "./agent-assets.ts";
 import { workflowToolDefinitions } from "../mcp/workflow-tools.ts";
 
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = path.dirname(__filename);
-const repoRoot = path.resolve(__dirname, "../..");
+const scriptDir = import.meta.dirname;
+const repoRoot = path.resolve(scriptDir, "../..");
 const AGENT_NAME = "azure-project-plan";
 
 function getBundledCliPath() {
-    const pkgDir = path.resolve(__dirname, "../node_modules/@github/copilot");
+    const pkgDir = path.resolve(scriptDir, "../node_modules/@github/copilot");
     const candidates = [];
     try {
         const pkg = JSON.parse(fs.readFileSync(path.join(pkgDir, "package.json"), "utf8"));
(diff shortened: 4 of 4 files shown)


alexweininger:
## Handoff: inventory for deleting the retired headless eval path

Parking this here rather than opening a PR, because the deletion belongs **inside** the #1693/#1694 adoption work, not ahead of it — deleting first would just manufacture a conflict that adoption then has to resolve. Everything below is verified; where I'm inferring, I say so.

**Background in one line:** running the agent headlessly moved to MSBench. `agent-contracts.yml` says so itself — *"Running the agent itself now happens on MSBench... Deliberately absent: check-copilot-auth.ts, check-gate-tools.ts and run-eval.cjs."* Those files are still in the tree, still linted, still typechecked, but nothing invokes them. The credential-free gates (drift / typecheck / certify / lint) stay — they're cheap, need no token, and guard the assets the MSBench run depends on.

### 1. What to delete (7 files)

| File | Why it goes |
| --- | --- |
| `evals/run-eval.cjs` | The headless runner entry point. Only reachable via `npm run eval`. |
| `evals/generate-report.cjs` | Only ever invoked by `run-eval.cjs`. Orphaned with it. |
| `evals/executor/azure-agent-executor.ts` | The Vally executor. Only referenced by `run-eval.cjs` and `project-plan/eval.yaml`. |
| `evals/executor/agent-assets.ts` | Only imported by the executor. |
| `evals/check-copilot-auth.ts` | Fails fast on a missing Copilot token — only meaningful for the headless run. Already excluded from CI. |
| `evals/check-gate-tools.ts` | Drives a live agent. Already excluded from CI. |
| `evals/mcp/workflow-tools.ts` | Imported **only** by the three files above; orphaned once they go. |

### 2. What must be edited as a consequence (7 files)

These are the ones that bite if you delete without grepping — a dangling `npm run` script is worse than no deletion.

| File | Why it needs touching |
| --- | --- |
| `evals/package.json` | Remove the `eval`, `eval:plan` and `gate-tools` scripts. Also `@github/copilot-sdk` becomes unused — **only** those three deleted files import it. Keep `@microsoft/vally`: the `vally` CLI is still what `npm run lint` runs. |
| `evals/tsconfig.json` | `include` explicitly lists `executor/*.ts` and `mcp/*.ts`. |
| **`evals/project-plan/eval.yaml`** | Declares `executor: azure-agent-executor` in `defaults`, plus a comment block pointing at `agent-assets.ts`. **See the warning below.** |
| `evals/graders/graderHarness.ts` | Its header documents the exit-code contract as being *"a contract with `generate-report.cjs`"*. Comment only, but it'd point at nothing. |
| `.github/workflows/agent-contracts.yml` | Header comment names `check-copilot-auth.ts`, `check-gate-tools.ts` and `run-eval.cjs` as deliberately absent. |
| `.vally.yaml` | See §4 — already stale today. |
| `evals/msbench/README.md` | Comparison table cites `evals/mcp/workflow-tools.ts` as the tool stand-in. |

### 3. CI evidence: the deletion is safe — and CI won't catch you if it isn't

I moved `evals/executor/` aside and ran the PR gates:

| Gate | Result |
| --- | --- |
| `npm run lint` (`vally lint --eval-spec project-plan/eval.yaml`) | **pass** |
| `npm run drift` | **pass** — *16 agent contracts intact* |
| `npm run typecheck` | **pass** |

Two things worth internalising:

- **`vally lint` does not validate the `executor:` key.** A dangling executor reference in `eval.yaml` lints clean. That's what makes the deletion safe — and equally means **CI will not catch a mistake in it**. Verify by hand, don't lean on the gates.
- `check-agent-drift.ts` mentions `agent-assets.ts` only inside a message string; it does not import it. Safe.

### 4. Pre-existing stale reference in `.vally.yaml`

Independent of any deletion, this is already broken today:

```yaml
environments:
  default:
    mcpServers:
      workflow-tools:
        command: node
        args: [evals/mcp/workflow-tools-server.mjs]   # <- this file does not exist
```

The TypeScript conversion left it behind. Worth fixing whenever `.vally.yaml` is next opened.

### 5. `stage-graders.ts` — the MSBench path is genuinely independent

The staged grader closure is only:

```
evals/graders/{graderHarness,validate-project-plan,validate-requirements,validate-webview-parseable}.ts
evals/src/artifacts/{projectPlan,requirements,validationTypes}.ts
src/webviews/copilotOnRails/views/utils/{parseRequirements,parseScaffoldPlanMarkdown}.ts
```

It does **not** reach the executor, `agent-assets.ts`, or `workflow-tools.ts`. So none of the deletions above can affect the live `exec:` assertions.

### 6. ⚠️ The `eval.yaml` warning, and the `.mjs` trap

`evals/project-plan/eval.yaml` is both the **MSBench source of truth** (the stimuli and graders in `msbench/config/stimuli/` are a port of it) and a file the in-flight PRs extend. Editing it in a deletion PR collides with adoption work — which is the whole reason this is a comment and not a PR.

**The trap to watch during adoption:** #1693 and #1694 both modify `evals/executor/azure-agent-executor.**mjs**` — the *pre-conversion* file. #1696 converted it to `.ts`. So both PRs carry a stale copy of a file that no longer exists under that name.

> When resolving those conflicts, **delete their `.mjs` changes — do not restore the file.** A naive "keep both sides" resolution silently resurrects the `.mjs` executor and reverts #1696's TypeScript migration.

The same hazard applies more broadly than the executor: both PRs predate the current msbench layout too. #1693 carries `msbench/stage.mjs`, `msbench/validate-suites.mjs` and a `msbench/suites/*.yaml` tree; #1694 carries `msbench/stage.sh`, `stage-graders.sh`, `preflight.sh` and `msbench/project-plan/user-overrides.yaml`. Today's layout is `config/base.yaml` + `config/stimuli/` + TypeScript `build-config.ts` / `stage-graders.ts`. Treat those paths as *intent to port*, not as files to merge.

---

Not planning further work on this — flagging for whoever picks up the convergence.

alexweininger:
## Follow-up: the red `Build / Build` on #1703 / #1704 is a **stale check**, not a live failure

Worth writing down, because it's confusing and will recur.

`main.yml` ("Node PR Lint, Build and Test") really does only trigger on PRs targeting `main`/`rel/*`:

```yaml
pull_request:
  branches:
    - main
    - rel/*
```

So no `Build / Build` run is ever created for a PR into `feat/CoR`. The red badge on #1703 and #1704 is left over from the brief window when those PRs targeted `main`:

| PR | Failing run created | `base_ref_changed` to `feat/CoR` |
| --- | --- | --- |
| #1703 | `2026-08-26T01:20:32Z` ([run 32918606897](https://github.com/microsoft/vscode-azureresourcegroups/actions/runs/32918606897)) | `2026-08-26T01:20:44Z` — 12s later |
| #1704 | — | `2026-08-26T01:31:34Z` |

The run fired against `main`, the PR was retargeted seconds afterwards, and GitHub keeps the old check attached. Nothing re-runs to supersede it, so it sits there looking live.

This is the `create_pull_request`-defaults-to-`main` quirk. #1700 hit it too — its stale red vanished only because it was force-pushed after retargeting, which regenerated the check set.

**To clear it on #1703 / #1704:** any new push clears the stale check. Rebasing onto current `feat/CoR` is the better move anyway, since #1705 has now merged and the underlying lint error is genuinely gone.

So: the lint break was real and worth fixing — it was red on #1615, the integration PR — but it was never *gating* the feature PRs. They only looked gated.

## Completeness check: the fix covers every site

Verified on merged `feat/CoR` (`9a2373a5`), full repo, exactly as CI runs it:

```
$ npm run lint          # eslint --max-warnings 0
LINT EXIT: 0
```

Clean with zero warnings, so all the `__dirname` / `__filename` sites are covered — not just the four files touched here.

### One gotcha if you reproduce this locally

If you run `npm run lint` after having run `evals/msbench/stage-graders.ts`, you'll get 9 spurious errors like:

```
evals/msbench/assets/graders/evals/src/artifacts/validationTypes.ts
  0:0  error  Parsing error: ... was not found by the project service
```

That's the **generated, gitignored** staging tree (`.gitignore:77`), not a real failure — those files don't exist in CI. Remove `evals/msbench/assets/graders/` and re-run.

PR #1716, opened 2026-08-26
Wire the eight merged scaffold and debug graders to real MSBench stimuli

alexweininger:
## ELI5

We recently merged eight "graders" — small programs that look at what the AI agent produced and say pass or fail. Four check the **scaffold** phase (#1707), four check the **local debugging** phase (#1708).

All eight had a problem: **they had never looked at anything an agent actually made.** They were only ever tested against example files we wrote by hand. There was no MSBench stimulus — no prompt, no workspace, no way to run them against the real extension in real VS Code.

This PR builds those missing stimuli. Six of them.

**Nothing in this PR has been run on MSBench.** Not once. Running costs money and needs sign-off. So this is the wiring, checked as far as it can be checked without spending anything — and the PR is careful not to imply otherwise.

## What a red assertion would now mean about the product

This is the part that matters. Each gate now corresponds to a specific way the product could be broken:

| Grader | Stimulus that now exercises it | A red result means |
| --- | --- | --- |
| `validate-frontend-scaffold` | `scaffold-fullstack`, `scaffold-autopilot` | The frontend the agent built **can't render in our own preview webview** — wrong vite host settings, a dev script that doesn't serve, a frame-busting CSP. The user clicks "Approve UI" and sees nothing, and the whole flow stalls. |
| `validate-integration-plan` | `scaffold-fullstack`, `scaffold-autopilot` | The hand-off note to the *integrate* agent is incomplete. That agent starts in a fresh session with no chat history, so a thin note means it wires the frontend to the wrong place, or not at all. |
| `validate-project-builds` | `scaffold-fullstack`, `scaffold-autopilot` | **The generated project does not install and build.** Every other check can pass on a project that looks right and is broken; this is the one that actually runs the package manager. |
| `validate-no-scaffold` | `scaffold-missing-plan`, `scaffold-unapproved-plan` | The agent **built a project it was not authorised to build** — from a plan the user never approved, or from no plan at all. This is a bypassed approval gate. |
| `validate-debug-plan` | `debug-plan-approval-gate`, `debug-generate-artifacts` | The debug plan is malformed, or (in 2.2) was never updated to record what was generated — so it stops being a truthful record of what the user approved. |
| `validate-debug-gate` | `debug-plan-approval-gate` | The debug agent **wrote launch/tasks configuration before the user approved the plan.** Same class of bug as above, different phase. |
| `validate-debug-config` | `debug-generate-artifacts` | `launch.json`/`tasks.json` are broken in a way the user hits **on their very first F5** — dangling `preLaunchTask`, a dependency cycle, a compound that starts a service twice. |
| `validate-debug-artifacts` | `debug-generate-artifacts` | Generation **quietly diverged from the plan the user approved** — produced something they didn't ask for, or skipped something they did. |

## The bit I'm most pleased with: falsifiable pairs

A single test can be passed by an agent that isn't really thinking. If the only test is "did you open the approval gate?", an agent that *always* opens it scores full marks without ever deciding anything.

So two of the scaffold stimuli are built as pairs that differ in **exactly one thing**:

| Pair | The single difference | What it forces |
| --- | --- | --- |
| `scaffold-fullstack` vs `scaffold-autopilot` | the `[AUTOPILOT MODE]` prefix on the prompt | must open the UI gate / must skip it and hand off |
| `scaffold-fullstack` vs `scaffold-unapproved-plan` | the plan's `**Status**` line, `Approved` to `Planning` | must scaffold / must refuse |

Same phase, same seed source, same graders, same flags. So an agent that *always* opens the gate fails one member of each pair, and one that *never* does fails the other. **There is no lazy strategy that passes both.**

**Untested as a pair, and that matters here more than elsewhere.** `scaffold-fullstack` has run (red, on the vite defect above); **`scaffold-autopilot` has never been submitted**, so the pair has only ever been asserted in one direction. The property above is a design claim, not a measured result. The run that would verify it is `scaffold-autopilot`, and the verdict to look for is *the gate skipped* rather than *the gate never reached* — both are green against a `COUNT(*) = 0` assertion, and only the first one means the pair discriminates.

The second pair's two plans are derived from one document in code — with an assertion that the status line was actually found — rather than checked in twice, because two files would drift in a second dimension the first time either was edited, and the pair would quietly stop discriminating while still reading green.

## Deliberately incomplete, and written down

`scaffold-api-only` is **not** included. It needs a plan with no frontend, and neither checked-in plan is API-only. Writing one by hand means inventing a project plan no planner ever emitted — worse than a stale harvested one, in a suite whose entire selling point is that it grades real output.

The gap is documented in `config/stimuli/README.md`, not just here, because an undocumented gap silently becomes a claim of coverage. It records:

- **The blind spot**: no scaffold coverage for the no-frontend shape. A scaffolder that invents a React frontend for an API-only project — and opens a UI approval gate for a project with no UI — would go completely uncaught.
- **The unblock is one run**: `harvest-seed.mjs` (commit `cc75a4e1`) already had `--target api-only` as a first-class target. This is a missing seed run, not a missing design.

## The honest part about seeding

Scaffold agents read `.azure/project-plan.md` as their first action, so the workspace needs one before turn 0. This PR seeds it from a checked-in fixture — **which is exactly the option a previous design rejected.** `harvest-seed.mjs` harvested plans from finished runs because:

> Checking one in makes that document a second source of truth: edit the planner's template and the stored copy still describes the old shape, so the scaffold graders keep passing against a plan no agent would emit.

That objection isn't answered here. It's **bounded**, by that script's own companion property:

> It is an *input, not an expected answer*. No scaffold grader reads the plan -- they assert against `resources/agents/**`. A wrong plan therefore makes scaffold trials fail loudly; it cannot make them pass wrongly.

So the drift is **fail-safe, not fail-open**: a stale plan makes real runs go red for a bad reason — annoying and expensive — but it cannot make a broken scaffolder look green, which is the failure that would actually matter. What we genuinely give up is that script's `--check` mode, which reported seed freshness and exited 1 when stale. Nothing here replaces it, and the README says so.

Stimuli declare *what state they need* (`# seed: approved-fullstack`), never a path or a command, so switching to baked images or a private benchmark repo later replaces one module instead of every stimulus.

## A guard that turns a paid failure into a local one

Both new phases set `snapshotWorkspace: false`, because they run a real `npm install` and snapshotting would copy `node_modules` into `session.sqlite` after every step. The schema warns that produces "multi-gigabyte session.sqlite files (and time out the run)".

But turning it off has a trap:

> To avoid silent/confusing results, the run fails fast if snapshotting is disabled while any assertion queries the 'files' table.

The `SELECT ... FROM files WHERE path LIKE ...` one-liner every existing stimulus uses — and therefore the obvious thing to copy — would now **take the whole run down**. `build-config.ts` performs that same check locally. Not second-guessing the schema; moving the discovery from "a submitted run, after the queue wait and container pull" to "milliseconds, on your machine, quoting the schema sentence".

## Verification (all local, no runs)

- `npm run typecheck` clean; **`npm run certify` 44/44**; `npm run drift` 16 contracts intact.
- The five pre-existing stimuli generate **byte-identical** output after the changes.
- All 11 stimuli generate and parse as real YAML; per-step `chatMode` resolves; every step begins with a liveness sentinel; **zero `files` queries under `snapshotWorkspace: false`**.
- The new guard was verified to fire (exit 1) with a deliberate `files` query, and the pre-existing duplicate-key error was verified still to fire.
- **Grader staging was verified outside the repo**, which is the only way to prove module resolution isn't accidentally falling back to the repo's own `node_modules`. Copied the staged tree to `/tmp` (no `node_modules` at or above it) and ran the graders there: all pass against certification fixtures, all exit **1** (product failure) rather than 3 (harness fault) against an empty workspace. `validate-debug-config`'s `jsonc-parser` import resolved from the staged copy.
- `preConditions` commands were executed against correct, wrong and missing seeds — each passes only on its own seed, including rejecting an accidentally-*approved* plan for `scaffold-unapproved-plan`.
- The seed pair was diffed: the approved seed is the fixture **verbatim**, and the unapproved seed differs by **exactly one line**.
- `validate-no-scaffold`'s seeded-entry list was checked against reality rather than assumed — read from the stored `files` table of run `2026082579322454`, the container workspace root holds `.gitignore` and `.gitkeep` and nothing else, neither of which trips the grader.

## What has now actually been run

**Three runs. Two green, and one red that is the most valuable result here.**

| Stimulus | Run | Result |
| --- | --- | --- |
| `scaffold-unapproved-plan` | `2026082618693091` | 6/6, `resolved: true` |
| `scaffold-missing-plan` | `2026082619460117` | 7/7, `resolved: true` |
| `scaffold-fullstack` | `2026082620153444` | **6/7, `resolved: false`** |

### The red: a real product defect, found on first contact with real agent output

`2026082620153444` is the first time any of these eight gates has graded a project the scaffold agent **actually built**. It failed, and it failed correctly:

```
FAIL: scaffolded frontend is preview-embeddable and keeps the API seam
  • [devServerRejectsWebviewOrigin] services/web/vite.config.ts: vite.config server
    must set `allowedHosts: true` or the dev server 403-blocks the webview origin.
```

The agent wrote `host: true`, `port`, `strictPort: false` and an `/api` proxy — and **omitted `allowedHosts: true`**. Two of the three settings `azure-project-scaffold.agent.md` explicitly requires. Confirmed against the agent's own `vite.config.ts` in `patch.diff`, not taken from the grader's message.

**Why it matters to a user.** The scaffold agent opens the frontend preview webview and *stops*, because that webview's **Approve UI** button owns the hand-off to the integrate agent. A dev server that 403-blocks the webview origin never renders — so the button can never be clicked and **the workflow stalls with no error**, while the app looks perfectly fine in an ordinary browser. That is exactly the failure the agent instructions warn about, now observed.

**Everything else passed, which is what makes this precise rather than "scaffolding is broken":**

| Check | Result |
| --- | --- |
| `validate-project-builds --require-frontend` | **pass** — the project really installs and builds |
| `validate-integration-plan --has-frontend` | **pass** |
| opened the preview gate | **pass** — `open_frontend_preview_view` called once |
| no premature hand-off to integrate | **pass** |
| `validate-frontend-scaffold` | **fail** — the one defect |

**Attribution was established before calling it anything:** exit **1**, not 3 → product failure, not a broken grader. No `NOT_APPLICABLE` marker → not a coverage gap. No `error.json`, zero `-> 429` arrows → not throttling. 14m34s elapsed against a 20-minute grader timeout → not a timeout kill.

**Neither the stimulus nor the grader was touched in response.** A gate edited until it passes is worth nothing.

### The two green runs, checked past the tally

A refusal stimulus scoring full marks is also what a dead run looks like, so both were read rather than tallied:

- **`2026082618693091`** — `read_file`, then refused: *"I notice the **Status** is still `Planning` rather than `Approved`."*
- **`2026082619460117`** — `read_file` → genuine not-found → `file_search` → contracted refusal **verbatim**. A refusal that never looked would have satisfied every negative assertion equally.

### Three things the runs settled that argument could not

1. **The `snapshotWorkspace` guard is necessary.** `2026082618693091` returned `files` at **0 rows** — any `files` query would have killed the run in-container, after the queue wait and the payment.
2. **The seed-clear works.** `2026082619460117` ran straight after a run that left a plan in `assets/workspace/`, and its fingerprint shows `ls: cannot access '.azure'`. A `test ! -f` precondition and a documented run ordering now make that mechanical.
3. **Latent limits are bounded, not guessed.** `output.zip` 8.8 MiB against a 1 GiB ingest cap; 874s elapsed against a ~7200s runner budget.

### Cost — and a correction

From the ATIF `final_metrics` block, read from the **local** artifact:

| Run | Total | Uncached prompt | Cache hit | Steps | Tool calls |
| --- | --- | --- | --- | --- | --- |
| `2026082618693091` | 44,041 | 12,018 | 72.2% | 4 | 2 |
| `2026082619460117` | 65,412 | 11,893 | 81.6% | 5 | 3 |
| `2026082620153444` | 2,625,438 | 89,215 | 96.6% | 39 | 88 |

**An earlier revision of this description reported roughly double these figures.** That method walked the trajectory summing every token-shaped field, which counted `final_metrics` *and* the per-step entries — exactly 2x. Corrected above.

**Do not read these as confirming that delegation grows with phase complexity.** All three runs produced a *single* trajectory with **no sub-agent calls**, so the #1714 under-count is zero here by construction. The scaffold run delegated *less* than the earlier planning run did, which contradicts the expectation rather than supporting it. n=1.

## What is still not verified

- **`scaffold-autopilot`** — so the autopilot half of the falsifiable pair is unexercised, and the pair property is untested as a pair.
- **Both local-dev stimuli** — so all four debug graders (`validate-debug-plan`, `validate-debug-gate`, `validate-debug-config`, `validate-debug-artifacts`) remain wired but never run.
- **`scaffold-api-only`** — deliberately not authored; see `config/stimuli/README.md`.

For those, offline certification is still the only evidence, and it proves a grader agrees with a fixture while saying nothing about whether a stimulus elicits the behaviour it looks for.

## CI: two guards fired on first contact with a clean environment

**1. The grader allowlist.** `build` went red with:

```
jsonc-parser is staged for the container but is not installed. Run 'npm ci' in evals/.
```

**This is the mechanism working, not failing.** `stage-graders.ts` refused to stage a package it could not find rather than emitting a tree that would have died minutes into a paid container run — and been reported as a product regression, since MSBench collapses every non-zero exit into "failed". Fixed by adding `npm ci` in `evals/` to **both** jobs: `--skip-build` skips the VSIX but still stages graders, so the `eval` job would have hit the same error on its first dispatch.

**2. Gate identity.** The liveness sentinel had drifted into **ten different wordings** across eleven stimuli. A SQL assertion has no grader filename, so its `comment` is the only stable identity it carries into stored results — rewording one silently forks that gate into a second identity with no history. The fork was live within this branch's own stimuli: two of these three runs stored *different* sentinel strings.

All sixteen are now one canonical string, and `npm run gates` fails CI if any deviates. The check keys on what an assertion *does*, not on what it says, so a copied query with a reworded comment is caught.


alexweininger:
needs msbench verification before merging

PR #1731, opened 2026-08-26
Derive which gates a stack runs, instead of hand-wiring them

alexweininger:
PR 1 (#1726) landed the stack schema as data. This makes it do something: **gate wiring is now a pure function of (stack facts × gate table × phase)** — no filesystem, no probing, so it is reviewable as a diff and testable without a container.

## Explain it like I'm 5

A stack file says what a project *has*. `config/gates.yaml` says what each gate *needs to look at*. Wiring is the intersection.

```yaml
# in the stack
project:
  frontend: none          # <- this one line unwires three gates

# in gates.yaml
- id: runtime-frontend
  requires:
    project.frontend: [spa]
```

Nobody maintains a 19×N matrix of gate names by hand, and `frontend: none` can be checked against the prompt by someone who has never read the gate.

## The rule for `requires:`, which is the whole design

**Put a fact there only when its absence means the gate has nothing to look at.**

| Situation | in `requires:`? | Outcome |
| --- | --- | --- |
| Nothing to grade (`frontend: none` vs a frontend gate) | yes | never wired — this is what makes `outOfScope` unreachable |
| Real question, prerequisite missing (no `func`, no Python support) | **no** | wired, red, declared in `knownGaps` |
| The grader cannot parse this ecosystem at all | yes | unwired, and reported *in words* as a gap in the gate |

The third row is narrow and was added after your ruling. `project-builds` runs the real package manager, so wiring it against a Python project makes it exit 1 with "no package.json anywhere" — **blaming the agent for a project the grader cannot read.** A fabricated product failure is worse than a missing one. My first draft justified this backwards in a comment; it now says what it is, and the derivation prints *"this gate only supports ecosystem node … a coverage gap in the gate, not a property of the project"* rather than letting the gate vanish silently.

I checked the rest of the table against your general form — *a declaration may change what a gate asserts; it must never reduce a gate to asserting nothing* — and every other exclusion unwires rather than passes.

## `--require-health` is now live

`args:` are derived from the same facts with the same predicate language, and this is not decoration. `--require-health` had existed on `validate-runtime-health` since it was written and **nothing had ever passed it** — a capability that existed only in its own documentation. A stack declaring `healthPath` now passes it. #1728 fixed the classification at source; this is the wiring-side half.

## Both directions, observable

Same gate table, `local` phase, the two stacks — this is a literal `diff` of the checked-in snapshots:

```
< react-functions-postgres                    > node-express-postgres
  runtime-frontend --require-frontend  WIRED    runtime-frontend      not wired (frontend is none)
  runtime-frontend-api                 WIRED    runtime-frontend-api  not wired (frontend is none)
  runtime-health   not wired (no healthPath)    runtime-health --require-health  WIRED
  runtime-crud     not wired (no route)         runtime-crud                     WIRED
```

`config/__snapshots__/<stack>.wiring.md` is generated and checked in, and the check fails when it drifts (`-- --update` regenerates). A one-line change to a `requires:` can silently unwire a gate across every stack, and **a gate wired nowhere reads as a clean run** — so "this change moves no wiring" is now something a reviewer sees rather than believes. The check also fails outright if two stacks ever derive an identical gate set, because that would mean the facts are not reaching the derivation and every "this is derived" claim in the tree is decoration.

## `--stack`, and the warning you asked for

```
./run.sh --stack node-express-postgres --phase plan
```

`build-config` synthesises the third layer — prompt, liveness sentinel, derived gate assertions — instead of reading a hand-written stimulus. **Every existing stimulus keeps working untouched.** The two sources are mutually exclusive rather than resolved by precedence, because silently grading one while the operator asked for the other is the expensive failure.

It **warns without blocking** when every gate a phase would run is a declared known gap: such a run is pre-determined to produce no new information, and that is computable before the money is spent.

The generated stimulus also carries an extended fingerprint line (`assertZeroExitCode: false`) probing `pip3 go dotnet docker azd func` and `ensurepip`, so the next stack run measures `container.yaml` instead of confirming it — closing the asserted-vs-measured gap from PR 1 without touching `stimuli/`.

## The projection, for rung 0

`assets/stack.json`, written on every `--stack` build. JSON rather than the YAML source because graders run from staged source with no `node_modules` — **the staging guard rewritten in #1727 catching a real problem at design time, on the first thing built after it.**

Only fields with a named consumer are projected — `healthPath`, `collectionRoute`, `start`, and `api` (which lets `runtime-app-starts` tell a background worker that never listens from a web app that failed to). Absent fields are omitted rather than nulled, so "declared" and "declared as nothing" stay distinct.

**One deviation to flag:** you specified `assets/graders/stack.json` plus a `stage-graders` change. I wrote `assets/stack.json` instead. `stage-graders` **wipes `assets/graders/` on every run**, and it runs *before* `build-config` in `run.sh` — so the original path works today only by ordering, and silently loses the projection the moment anyone reorders those two lines. Everything under `assets/` is copied to `/agent/assets` regardless, so the container path is `/agent/assets/stack.json` and **no `stage-graders` change is needed at all.** Say the word if you want it moved back.

## Proof, by regression rather than assertion

34 rejection fixtures, each asserting its own error code (16 stack, 8 container, 8 gate table, 2 cross-check). Coverage ratchet raised 24 → 34. Then each new mechanism was deliberately broken:

```
R1  remove a stack fact
      ✖ knownGaps[0] declares a gap for 'runtime-crud', but this stack's facts
        never wire that gate in any phase                             EXIT=1

R2  delete one fixture
      Rule coverage: 33 of 87 error codes proven by a fixture.
      ✖ rule coverage fell to 33; MIN_PROVEN_CODES is 34              EXIT=1

R3  strip every `requires:` from the gate table
      ✖ node-express-postgres.wiring.md is out of date
      ✖ react-functions-postgres.wiring.md is out of date
      ✖ every stack derives an identical gate set, so the derivation is
        not discriminating between them                               EXIT=1
```

The three derivation branches are driven directly rather than left to be right by assumption, including the one no fixture reaches: **an empty wired set must not warn**, since "every gate is a known gap" over zero gates is vacuously true — the `COUNT(*) = 0` shape again.

## Verification

All six credential-free gates green on the rebased base (`fafab221`, after #1728): `typecheck`, `stacks:check`, `imports:self-test`, `drift`, `certify` (81/81), `lint`. Both `build-config` paths produce YAML that parses, and every error path exits non-zero. 20 files, +1405/−22.

**No MSBench run was submitted.**


alexweininger:
**Amended after CI caught me walking into the trap this PR stack documents.**

The first push turned `build` red:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'yaml'
  imported from evals/src/containerInventory.ts
```

`build-config.ts` had imported `node:` builtins and nothing else. I added four static imports of the stack machinery, which pulls in `yaml` — and `run.sh` invokes `build-config` on hosts with **no `node_modules` at all**. That is the identical constraint I wrote up two PRs ago as the reason `stage-graders.ts` cannot use `ts.preProcessFile`, and I broke it in the next thing I built, in the same file family, having just documented it.

Worth saying plainly: **knowing the rule and writing it down did not make me apply it.** CI caught it because the guard exists; nothing about my care did.

Fixed by deferring the import to the point of use:

- the **stimulus path stays dependency-free** — verified by hiding `evals/node_modules` and building `photo-app-requirements` successfully;
- the **stack path reports the missing install** rather than a stack trace about a package nobody mentioned:

```
--stack needs the eval dependencies, which are not installed here.
Run 'npm ci' in evals/ and try again. (Cannot find package 'yaml' ...)
EXIT=1
```

- `run.sh` installs the eval dependencies **only** on the `--stack` path, so a stimulus run on a bare host still costs nothing.

`build`, `contracts` and `api-extractor` are green on the amended commit.

PR #1758, opened 2026-08-30
Scaffold: forbid tsconfig `paths` aliases for cross-workspace imports (emitted specifier must resolve)

### LINKED ISSUE #1757: Scaffolded Functions project builds green but cannot start: tsconfig path alias emits an unresolvable runtime specifier
> **For a coding agent:** the fix is in this repo, in the scaffold agent instructions — not in any generated project. Jump to [Where to fix it](#where-to-fix-it).

## Summary

The scaffold agent emits a monorepo whose Functions workspace imports shared code through a **tsconfig `paths` alias** (`@shared/*`). TypeScript resolves that at compile time and **does not rewrite module specifiers on emit**, so the emitted JavaScript still asks Node for a package that does not exist. The project builds clean, type-checks clean, and the Functions worker dies on load.

This violates a rule the product documentation already states — it is simply not enforced anywhere.

## Evidence

Found by the MSBench eval suite in run `2026083057881445` (stack `react-functions-postgres`, phase `local`).

`validate-project-builds` passes:

```
[project-builds] services/functions: npm ci
[project-builds] services/functions: npm run build
[project-builds] built 3 package(s): services/functions, services/shared, services/web
PASS: gate=project-builds — scaffolded project installs and builds
```

The app then fails to start:

```
Azure Functions Core Tools
Core Tools Version:       4.14.0+4a17060ecc915f1672d86717a487ace30f535e74 (64-bit)
Function Runtime Version: 4.1052.200.26352

[...] Worker was unable to load entry point "dist/functions/src/functions/createTask.js":
      Cannot find module '@shared/schemas/index'
Require stack:
- /workspace/services/functions/dist/functions/src/functions/createTask.js
- /usr/lib/azure-functions-core-tools-4/workers/node/dist/src/worker-bundle.js
```

All five `runtime-*` gates fail as a cascade of this one root cause.

## Root cause

The generated project contains **two competing mechanisms for the same shared code**.

`services/shared/package.json` is correct and well-formed, with real subpath exports:

```json
{
  "name": "@task-tracker/shared",
  "type": "module",
  "main": "./dist/index.js",
  "exports": {
    ".":         { "types": "./dist/index.d.ts",         "default": "./dist/index.js" },
    "./types":   { "types": "./dist/types/index.d.ts",   "default": "./dist/types/index.js" },
    "./schemas": { "types": "./dist/schemas/index.d.ts", "default": "./dist/schemas/index.js" }
  }
}
```

But the Functions code never imports that package. It goes through an alias pointing at the shared package's **source**:

```jsonc
// services/functions/tsconfig.json
"paths": { "@shared/*": ["shared/src/*"] }
```

```ts
import { createTaskSchema } from '@shared/schemas/index';
import type { Task, CreateTaskDto } from '@shared/types/index';
```

`tsc` resolves `@shared/*` and emits `require('@shared/schemas/index')` verbatim. At runtime there is no `@shared` package in `node_modules` and no runtime path resolver — no `tsconfig-paths`, no bundler, no `imports` map — so Node cannot resolve it.

The correct specifier for the package that was actually generated is `@task-tracker/shared/schemas`; the documented convention is the relative form `../shared/...`.

## The rule this breaks already exists

`resources/agents/shared-references/architecture.md` states it twice, in the naming callout:

> Whatever names you pick, apply them **consistently everywhere** — npm `workspaces`, `cd` commands, tsconfig `rootDir`, and the computed `main` field […] **Imports of the shared package stay `../shared/...`.**

and prescribes the supported mechanism under *TypeScript Cross-Workspace Import Configuration*:

> When Functions imports from `../shared/`, `tsconfig.json` must set `rootDir` to reach outside workspace

```jsonc
// services/functions/tsconfig.json
{
  "compilerOptions": { "rootDir": "..", "outDir": "dist" },
  "include": ["src/**/*.ts", "../shared/**/*.ts"]
}
```

The generated project **did** follow the `rootDir` half — the emitted path is `dist/functions/src/functions/createTask.js`, exactly the documented `rootDir: ".."` shape — and then imported by alias anyway. Every code example in the instructions uses relative imports (`'../../services/shared/types/entities'`). No file in `resources/agents/` mentions `paths` or `@shared` at all, so the alias is invented by the model, and nothing tells it not to.

## Why the existing checkpoints miss it

`resources/agents/azure-project-scaffold/instructions.md`, Step 2 checkpoint block:

| # | Checkpoint | Why it passes anyway |
|---|---|---|
| 3 | Shared package has `exports`/`main`, builds to `dist/` | True here. The package is fine; nothing imports it. |
| 4 | **Cross-workspace imports (CRITICAL)**: run `tsc --noEmit`, fix `TS2307` | The alias is *precisely* what makes `tsc` succeed. This check can never fire on this defect. |
| 5 | `rootDir`/`main` match actual `dist/` output | True here. |
| 6 | Production install must satisfy every `import` in `dist/` | Closest in spirit, but framed as a `dependencies`/`devDependencies` split, so it reads as "is the package in the right section" rather than "is this specifier resolvable at all". |

Checkpoint 4 is the one that looks like it covers this and does not. Every check is a **compile-time** check, and this is a defect that only exists **after** emit.

<a id="where-to-fix-it"></a>
## Where to fix it

**1. `resources/agents/azure-project-scaffold/instructions.md`** — Step 2 checkpoint block (the `✅ Checkpoint` list containing items 1–6 and the `⚠️ Pitfalls` line).

Add a checkpoint that inspects **emitted output**, not source. Suggested wording:

> **Emitted specifiers resolve (CRITICAL)**: `tsc` does not rewrite module specifiers. Any non-relative import that only resolved through a tsconfig `paths` alias will still be in `dist/` and will fail at runtime with `Cannot find module`. After building, verify every non-relative specifier in `dist/` resolves from that workspace. Do **not** use `paths` aliases for cross-workspace imports — import the shared package by its `name`, or use the relative `../shared/...` form.

Add to the `⚠️ Pitfalls` line: *`paths` aliases resolve at compile time only → `Cannot find module` at runtime.*

**2. `resources/agents/shared-references/architecture.md`** — *TypeScript Cross-Workspace Import Configuration* section.

The "imports stay `../shared/...`" rule currently lives inside a callout about **folder naming**, which is not where anyone looks for import mechanics. State it in this section as a rule with its reason, and show the alias as an explicit anti-pattern:

```jsonc
// ❌ compiles, fails at runtime — tsc emits require('@shared/schemas/index') unchanged
{ "compilerOptions": { "paths": { "@shared/*": ["shared/src/*"] } } }
```

**Non-goal:** do not add `tsconfig-paths` or a bundler to make the alias work. The supported mechanisms are the relative import and the workspace package name; adding a third would leave the same two competing mechanisms in place.

## Verification command

Cheap, dependency-free, and would have caught this. Run from the workspace that emitted `dist/`:

```bash
node -e "
const fs=require('fs'),path=require('path'),Module=require('module');
const bad=[];
(function walk(d){ for(const e of fs.readdirSync(d,{withFileTypes:true})){
  const p=path.join(d,e.name);
  if(e.isDirectory()) walk(p);
  else if(e.name.endsWith('.js')) for(const m of fs.readFileSync(p,'utf8').matchAll(/require\(['\\\"]([^'\\\"]+)['\\\"]\)/g)){
    const s=m[1];
    if(s.startsWith('.')||s.startsWith('node:')||Module.builtinModules.includes(s)) continue;
    try { Module.createRequire(p).resolve(s); } catch { bad.push(p+' -> '+s); }
  }}})('dist');
if(bad.length){ console.error('unresolvable specifiers in emitted output:'); bad.forEach(b=>console.error('  '+b)); process.exit(1); }
console.log('all emitted specifiers resolve');
"
```

On the run above this prints `dist/functions/src/functions/createTask.js -> @shared/schemas/index` and exits 1.

## How to confirm the fix

The eval suite reproduces this end to end. From `evals/`:

```bash
BENCHMARK=corbench.cor_functions_host bash msbench/run.sh --skip-build \
  --stack react-functions-postgres --phase local \
  --dataset evals/msbench/container/dataset.jsonl

npm run analyze-run -- <run-id>      # per-gate verdicts
```

Fixed looks like `runtime-app-starts` passing and `runtime-crud` completing a round-trip. The datastore emulators are already running in this image, so a failure there will be about the app, not the environment — `gates-selftest-emulators` proves the emulator path independently (run `2026083068733550`, `runtime-crud` green against a preamble-started PostgreSQL).

## Not the cause

The environment was healthy for the run that found this:

```
func already present: 4.14.0
azurite: listening on 127.0.0.1:10000
postgres: ready on 127.0.0.1:5432 (role taskuser, db tasktracker)
```

Run link: https://msbenchapp.azurewebsites.net/run-analysis/2026083057881445 — verified as a genuine result (156 `/v1/messages` calls, `claude-sonnet-4.5` requested and active, not throttled).

nturinski:
﻿## The defect

The scaffold agent generates a Functions monorepo that **builds green and cannot start**. Found by MSBench run [`2026083057881445`](https://msbenchapp.azurewebsites.net/run-analysis/2026083057881445) (stack `react-functions-postgres`, phase `local`).

It emits `services/functions/tsconfig.json` with a path alias:

```jsonc
"paths": { "@shared/*": ["shared/src/*"] }
```

and imports through it (`import { createTaskSchema } from '@shared/schemas/index'`). `tsc` resolves `paths` at compile time and **does not rewrite module specifiers on emit**, so `dist/functions/src/functions/createTask.js` still contains `require('@shared/schemas/index')`. There is no runtime resolver — no `tsconfig-paths`, no bundler, no `imports` map — so the worker dies:

```
Worker was unable to load entry point "dist/functions/src/functions/createTask.js":
Cannot find module '@shared/schemas/index'
```

`npm run build` and `tsc --noEmit` both pass. All five `runtime-*` gates fail as a cascade of this one root cause.

The generated project also contains a correct `services/shared/package.json` named `@task-tracker/shared` with proper subpath `exports`. Nothing imports it. Two competing mechanisms, and the compile-time-only one wins silently.

## Why the existing checkpoints miss it

Every Step 2 checkpoint is a **compile-time** check; this defect only exists **after** emit. Checkpoint 4 ("Cross-workspace imports (CRITICAL): run `tsc --noEmit`, fix TS2307") is the one that looks like it covers this and can never fire — the alias is precisely what makes `tsc` succeed.

### 1. `resources/agents/azure-project-scaffold/instructions.md` — Step 2 checkpoint block

New checkpoint 7, which inspects **emitted output** rather than source:

> **Emitted specifiers resolve (CRITICAL — after build, not `--noEmit`)**: `tsc` does **not** rewrite module specifiers on emit. Any non-relative import that only resolved through a tsconfig `paths` alias is still in `dist/` verbatim and dies at runtime with `Cannot find module`, even though checkpoints 1–6 all pass. **Do NOT use `paths` aliases for cross-workspace imports** — import the shared package by its `package.json` `name`, or use the relative `../shared/...` form. After building, verify every non-relative specifier in `dist/` actually resolves from that workspace.

followed by the issue's dependency-free Node script (walk `dist/`, skip relatives and builtins, `Module.createRequire(file).resolve(spec)` everything else, exit 1 on failure). Plus a new Pitfalls entry:

> (4) tsconfig `paths` aliases (e.g. `@shared/*`) resolve at compile time only — `tsc --noEmit` passes and the emitted `dist/` still says `Cannot find module` at runtime.

### 2. `resources/agents/shared-references/architecture.md` — *TypeScript Cross-Workspace Import Configuration*

The "imports stay `../shared/...`" rule already existed, but only inside a callout about folder **naming** — not where anyone looks for import mechanics. Restated in the import section as a rule with its reason, with the alias as an explicit anti-pattern (`❌ compiles, fails at runtime — tsc emits require('@shared/schemas/index') unchanged`), a ✅ relative form, a ✅ package-name form, and the closing line: *these are the only two supported mechanisms; picking both leaves two competing paths to the same code, and the compile-time-only one wins silently.* The relative example counts levels from the importing file (`services/functions/src/functions/createTask.ts` → `../../../shared/schemas/index`), since handlers sit one level deeper than `src/`.

## What is and is not verified

**Verified end to end against a real MSBench run.** Run [`2026083152684407`](https://msbenchapp.azurewebsites.net/run-analysis/2026083152684407) (`react-functions-postgres`, phase `local`) vs baseline [`2026083057881445`](https://msbenchapp.azurewebsites.net/run-analysis/2026083057881445). Run on a VSIX built from these `resources/` changes plus #1755's harness, with the built VSIX confirmed to actually carry them (`extension/resources/agents/azure-project-scaffold/instructions.md` contains "Emitted specifiers resolve") before the run was trusted.

| Gate | Baseline | With this change |
|---|---|---|
| `project-builds` | pass | pass |
| `runtime-app-starts` | **FAIL** — `Cannot find module '@shared/schemas/index'` | **pass** |
| `runtime-health` | FAIL (cascade) | **pass** |
| `runtime-frontend` / `runtime-api` | FAIL (cascade) | n/a — `frontendDevServerUnsupported` |
| `runtime-crud` | FAIL (cascade) | n/a — `noCollectionRouteDeclared` |

The `debug-*` gates are unchanged in both runs, so the runtime column is the only thing that moved.

**The mechanism was checked, not inferred from the gate.** In the newly generated project, `services/functions/tsconfig.json` has **no `paths` alias at all** and uses exactly the shape architecture.md prescribes (`"rootDir": ".."`, `"include": ["src/**/*.ts", "../shared/**/*.ts"]`). The one surviving alias is in `services/web/tsconfig.app.json` under `"moduleResolution": "bundler"` with `"noEmit": true` — Vite resolves and rewrites it, so it never reaches runtime. That is the correct outcome: both new rules are scoped to *cross-workspace* imports ("Do NOT use `paths` aliases for cross-workspace imports"; the architecture.md rule sits in the cross-workspace-import section and gives its reason as "no runtime counterpart … **in these projects**"). A blunter "never use `paths`" would have broken the bundled frontend, where aliases are legitimate.

**Also verified:**
- The Node script in checkpoint 7 was run against a fixture reproducing the emitted defect. It reports both specifiers and exits 1:
  ```
  unresolvable specifiers in emitted output:
    dist\functions\src\functions\createTask.js -> @shared/schemas/index
    dist\functions\src\functions\createTask.js -> @shared/types/index
  ```
  and on a clean fixture prints `all emitted specifiers resolve` (exit 0). The regex was broadened from the issue's `require(...)`-only form to `(?:require\(|from\s*)` so it catches ESM emit too — the generated shared package is `"type": "module"`, so an ESM-emitting variant of this defect would otherwise slip through.
- Nothing else in `resources/agents/` recommends a `paths` alias — before this change the string did not appear anywhere; every code example uses relative imports.

**Still NOT proven — please keep these in mind when reading the green:**

1. **n=1.** This is a probabilistic behaviour change in a model, not a deterministic code fix. One green run is real evidence, not proof. The agent previously followed half this convention correctly (`rootDir: ".."`) and invented the other half, so a single conforming run does not guarantee every run conforms.
2. **The CRUD path is still unproven on an agent-generated project.** `runtime-crud` did not exercise a datastore round-trip — it stood down on `noCollectionRouteDeclared` because the frontend runs as its own dev server, which the gates do not start. That is a pre-existing harness limitation unrelated to this change, but it means "app starts and is healthy" is the strongest claim supported here, not "CRUD works".

`npm run lint` (`eslint --max-warnings 0`) was not run: `node_modules` is not installed in this worktree, and `eslint.config.mjs` only matches `.ts`/`.tsx`/`.js` — this change touches only Markdown, so ESLint would not read either file. There is no markdown linter in the repo and none was added.

## Contradictions found in existing instructions (not changed — flagging for a follow-up)

Two places still describe the shared-types entry point as an "import **alias**", which is the exact mental model that produces this bug. Both are ambiguous rather than wrong — `@app/shared` is a legitimate workspace *package name* — but the word "alias" invites the tsconfig `paths` reading:

- `resources/agents/azure-project-scaffold.agent.md:68` — "the shared package/location and import alias (e.g. `@app/shared`) for the typed client"
- `resources/agents/azure-project-scaffold/instructions.md:417` (Step 11 integration artifact) — "the shared-types package + import alias"

Left alone to keep this diff tight and reviewable, as the issue asks. Changing "alias" → "specifier"/"package name" in both would be a one-word follow-up.

## Non-goals respected

- No `tsconfig-paths`, no bundler, no `imports` map. The two supported mechanisms stay two.
- Nothing under `evals/` touched (that's #1755). `npm run eval:drift` **does fail on this branch** — here is exactly what it says, so no reviewer has to investigate it:
  - **`agent-assets-changed` is the only failure.** It is the aggregate SHA-256 over the tracked scope. Expected and benign.
  - **The changed-files list it prints depends on when you run it.** Today it names only `shared-references/architecture.md`. It does *not* name `azure-project-scaffold/instructions.md` — the larger half of this fix — because the tracked scope on `feat/CoR` is currently `azure-project-plan.agent.md, azure-project-plan/**, shared-references/**`, and the scaffold agent is not in it. I verified that first-hand: on a pristine `feat/CoR` agents tree, appending a comment line to `azure-project-scaffold/instructions.md` leaves drift **green**. So checkpoint 7 is presently unguarded — it could be deleted tomorrow without drift noticing. #1755 (`934d54f`) widens tracking to every agent under `resources/agents` bar `azure-deploy`; **once that lands in `feat/CoR`, this PR will correctly fail drift naming both files.** That is the guard working, not a regression.
  - **All contract checks pass** — 16 under the current scope. Those are the per-rule assertions that each name the grader depending on them; none reference the sections I edited. `check-agent-drift.ts` accumulates every contract failure, consistency-rule failure, and the hash mismatch into a single `failures` array with no early exit, so a broken contract would be listed *alongside* the hash mismatch, not hidden by it. Confirmed independently by reverting to `origin/feat/CoR` in the working tree, where the check prints `✔ 16 agent contracts intact; assets match the verified baseline`, then restoring.
  - **The baseline is deliberately not regenerated.** The check's own remedy text is *"Re-run the evals against the new instructions, then run `node evals/check-agent-drift.ts --update`"* — that ordering is the whole point of the guard. Updating `evals/agent-assets.lock.json` here would rubber-stamp an instruction change with no evidence it doesn't move the graders, which is precisely what the guard exists to prevent, and it applies with more force once the scaffold agent is in scope. It should move as part of the confirming eval run. No CI workflow invokes it today.
- No reformatting or restructuring beyond the change.

Fixes #1757


nturinski:
## Confirming run: `2026083170541011` — the runtime column reproduces

Run [`2026083170541011`](https://msbenchapp.azurewebsites.net/run-analysis/2026083170541011) (`react-functions-postgres`, phase `local`), on a VSIX built from this branch plus #1759. **Verified as a genuine result** by `verify-run.ts`, not a `RATE_LIMIT` void.

| Gate | Baseline `2026083057881445` | This PR `2026083152684407` | This run `2026083170541011` |
|---|---|---|---|
| `project-builds` | pass | pass | pass |
| `runtime-app-starts` | **FAIL** `Cannot find module '@shared/schemas/index'` | pass | **pass** |
| `runtime-health` | FAIL (cascade) | pass | **pass** |
| `runtime-frontend` / `-api` | FAIL (cascade) | `frontendDevServerUnsupported` | `frontendDevServerUnsupported` |
| `runtime-crud` | FAIL (cascade) | `noCollectionRouteDeclared` | `noCollectionRouteDeclared` |
| `debug-config` / `-artifacts` / `-breakpoint` | FAIL / FAIL / exit 3 | same | same |

Gate-for-gate identical to this PR's own verification run. The description's "n=1" caveat is now **n=2**, on two independently generated projects. `functionsHostUnavailable` and `Cannot find module` are both absent from each.

## Two things this run also settles

**The `agent-assets-changed` baseline in `cb77ff9` is now signed for.** It was refreshed before a confirming run existed, which I flagged in that commit message. This is that run.

**A detour was needed to get here.** The first attempt on this branch ([`2026083167798590`](https://msbenchapp.azurewebsites.net/run-analysis/2026083167798590)) came back red with all five runtime gates reporting `declares dependencies but has no node_modules` — on a tree where `project-builds` passed. That was a harness bug, not this PR: the agent installed with `npm install --workspace=services/functions`, npm hoisted to `/workspace/node_modules`, and the precondition only looked for `services/functions/node_modules`. Fixed in #1759, which this run includes.

The three runs happen to form a clean experiment, because the agent picked a different npm invocation each time:

```
2026083152684407  npm --prefix services/functions install     -> local     -> pass
2026083167798590  npm install --workspace=services/functions  -> hoisted   -> GRADER ERROR
2026083170541011  npm install                                 -> hoisted   -> pass (with #1759)
```

## The finding I'd most like a decision on

**The agent has never executed checkpoint 7's script.** Not in any of the three runs.

```
run 2026083152684407 — 26 run_in_terminal calls  -> checkpoint-7 script executed? NO
run 2026083167798590 — 19 run_in_terminal calls  -> NO
run 2026083170541011 — 16 run_in_terminal calls  -> NO
```

In each case the `read_file` of the Step 2 block **did** return the script to the agent, and in each case it ran `find services/functions/dist -type f -name "*.js"` — that is checkpoint 5's "list actual dist/ contents" — and then moved on.

So every green above was earned by the **prose** half of this PR: "do not use `paths` aliases for cross-workspace imports; use the relative form or the package `name`." That is the part with runtime evidence behind it, and it is the larger half of the change.

It also explains something that looked odd earlier: the script as originally written reported *every* bare specifier as unresolvable (see `626fb38`), so `@azure/functions` and `zod` would have been flagged on a perfectly healthy project. That never caused a false-red storm — because it never ran.

Worth deciding whether checkpoint 7's script should become imperative (an explicit required step the agent must report the output of, the way checkpoints 1 and 4 name a command and a failure mode), or be dropped as decoration. As it stands it is carrying no weight.

Still open from my earlier review, unchanged and out of scope for the fixes in `626fb38`: `.mjs`/`.cjs` are skipped and dynamic `import()` is unmatched; the word `from` inside a string false-positives; and the ✅ relative example omits the `.js` extension, which is the same defect class under ESM emit and is invisible to checkpoint 7 because it skips relative specifiers.

PR #1759, opened 2026-08-31
Runtime gates: resolve dependencies the way Node does, not by one directory's existence

nturinski:
Found while re-running #1758's confirming eval. Two runs of `react-functions-postgres`/`local` that differ in nothing but which equivalent `npm` command the agent chose:

| Run | Install command | Result |
|---|---|---|
| [`2026083152684407`](https://msbenchapp.azurewebsites.net/run-analysis/2026083152684407) | `npm --prefix services/functions install` | local `node_modules`; gates ran, `runtime-app-starts` + `runtime-health` **passed** |
| [`2026083167798590`](https://msbenchapp.azurewebsites.net/run-analysis/2026083167798590) | `npm install --workspace=services/functions` | hoisted to `/workspace/node_modules`; **all five** runtime gates exit 3 |

The second run's gates all said:

```
GRADER ERROR: gate=runtime-app-starts — could not run —
/workspace/services/functions declares dependencies but has no node_modules,
so the app cannot start. Run validate-project-builds (or npm install) first.
```

…on a tree where `project-builds` **passed**. A passing build gate and five gates claiming nothing is installed cannot both be describing the product.

## Cause

`runtimeTarget.ts` tested `existsSync(<package>/node_modules)`. Node resolves a bare specifier by walking `node_modules` *up* the directory chain, and npm workspaces hoist to the monorepo root — which is exactly the layout `architecture.md` prescribes (`workspaces: [services/functions, services/web, services/shared]`). So a fully installed workspace legitimately has no `node_modules` of its own, and the one precondition whose purpose is to avoid inventing a failure invented one.

## Fix

Walk up for each declared dependency, mirroring Node. Two deliberate choices, both recorded in the code:

- **Presence on disk, not `require.resolve`** — resolution applies `exports` conditions, so an ESM-only dependency throws `ERR_PACKAGE_PATH_NOT_EXPORTED` under CJS resolution and would be reported missing. That trades this false negative for a subtler one.
- **Bounded at `workspaceRoot`** — Node walks to the filesystem root; this also runs on developer machines, where an unbounded walk out of a fixture finds the repo's own `node_modules` and reports an uninstalled project as installed.

## Verification

Six fixtures covering both real run layouts. The fix passes all six; the previous implementation fails exactly one — `hoisted`, i.e. run `2026083167798590` — and no safety case changes:

```
local     (npm --prefix ...)                 true   was true
hoisted   (npm install -w ...)               true   was FALSE
split     (hoist + nested version pin)       true   was true
partial   (one dep missing)                  false  was false
none      (never installed)                  false  was false
bounded   (node_modules ABOVE workspaceRoot) false  was false
```

`drift`, `typecheck`, `imports:self-test`, `certify`, `stacks:check`, `gates`, `lint`, `clean-machine:check` all pass.

## Not done

No permanent self-test wired into `contracts`. The six fixtures above were scratch. Happy to add one if you want this guarded the way `imports:self-test` and `msbench:self-test` are — it is the kind of rule this repo usually turns into a mechanism.

### DIFF
--- a/evals/src/runtime/runtimeTarget.ts
+++ b/evals/src/runtime/runtimeTarget.ts
@@ -177,9 +177,13 @@ export interface RuntimeTarget {
     /** True when that package declares runtime dependencies. */
     declaresDependencies: boolean;
     /**
-     * True when `node_modules` is present. Reported rather than judged: the runtime gates
-     * treat a missing install as a harness fault ("run the build gate first"), while the
-     * debug probe does not care, because VS Code runs the `preLaunchTask` install itself.
+     * True when those dependencies are installed somewhere Node will find them. Reported
+     * rather than judged: the runtime gates treat a missing install as a harness fault ("run
+     * the build gate first"), while the debug probe does not care, because VS Code runs the
+     * `preLaunchTask` install itself.
+     *
+     * Not the same question as "does `<package>/node_modules` exist" — see
+     * `dependenciesAreInstalled`.
      */
     dependenciesInstalled: boolean;
 }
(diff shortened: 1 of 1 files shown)


nturinski:
## Confirmed in situ: run `2026083170541011`

The six fixtures in the description are synthetic. This is the fix doing its job on a real run.

Run [`2026083170541011`](https://msbenchapp.azurewebsites.net/run-analysis/2026083170541011) (`react-functions-postgres`, phase `local`) on a VSIX built from this branch plus #1758. Verified as a genuine result by `verify-run.ts`.

```
project-builds --require-frontend    pass
runtime-app-starts                   pass
runtime-health                       pass
runtime-frontend / -api              SKIP  frontendDevServerUnsupported
runtime-crud                         SKIP  noCollectionRouteDeclared
```

**The fix was actually exercised** — worth checking rather than assuming, since if the agent had happened to install locally again the run would prove nothing. It did not. Its install commands:

```
npm --prefix services/web install
npm --prefix services/web install @fluentui/react-components @fluentui/react-icons react-router-dom
npm install                        <- root, npm workspaces hoists to /workspace/node_modules
npm --prefix services/shared run build
npm --prefix services/functions run build

used workspace/root install (exercises this fix)? YES
used --prefix local install (would NOT exercise it)? no
```

A bare root `npm install`. `services/functions/node_modules` never existed, so on `feat/CoR` as it stands this run would have produced the same five `declares dependencies but has no node_modules` harness faults as `2026083167798590`.

## Three runs, three different npm invocations, all agent-chosen

This is the part that makes the fix worth having rather than a one-off patch — the agent picks a different form each time, and none of them is wrong:

| Run | Install command | `services/functions/node_modules`? | `runtime-app-starts` |
|---|---|---|---|
| [`2026083152684407`](https://msbenchapp.azurewebsites.net/run-analysis/2026083152684407) | `npm --prefix services/functions install` | yes | pass |
| [`2026083167798590`](https://msbenchapp.azurewebsites.net/run-analysis/2026083167798590) | `npm install --workspace=services/functions` | no (hoisted) | **GRADER ERROR** |
| [`2026083170541011`](https://msbenchapp.azurewebsites.net/run-analysis/2026083170541011) | `npm install` | no (hoisted) | pass (this fix) |

Two of the three hoist. Before this change the gates were effectively flipping a coin on every run, and losing it produced five red gates and a contradiction — a passing `project-builds` beside five gates claiming nothing was installed — that cost a full run to diagnose.

## Note on the `certify` timestamp

`npm run certify` rewrites `generatedAt` in `evals/results/grader-certification/offline/report.json` on every invocation, so it shows up as a diff even when no certification outcome changed. I reverted it here rather than committing the churn. Might be worth making that file's timestamp stable, or gitignoring it, since it guarantees a spurious dirty tree for anyone who runs the suite locally.

evals/src/runtime/runtimeTarget.ts
```diff
@@ -374,11 +378,66 @@ async function completeTarget(
...
+ *                     node_modules" while `project-builds` passed on the same tree
+ *
+ * Both invocations are correct npm. The gates were grading which one the agent happened to
+ * pick, and reporting the difference as a harness fault — a red with nothing to say about the
+ * product, from a precondition that exists specifically to avoid inventing one.
+ *
+ * Two deliberate choices:
+ *
+ * - **Presence on disk, not `require.resolve`.** Resolution applies `exports` conditions, so an
+ *   ESM-only dependency (`"type": "module"` exposing only an `import` condition) throws
```
nturinski:
```suggestion
```

PR #1764, opened 2026-08-31
Phase1: feat(copilot-on-rails): add Podman as a local-debug container runtime…

nturinski:
… option

Phase 1 of Podman support. The local-debug agents can now select Podman (or Docker, the default) to run emulator containers, and generation emits matching `podman compose` / `docker compose` task commands. The generated compose file is unchanged for either engine.

- prerequisites: detect Docker and Podman (CLI, Compose provider, and the Podman machine on Windows/macOS); select one, default Docker
- debug plan: record container runtime + Compose command in the Orchestrator table; generation/validation/preflight/migrations read that command instead of hard-coding `docker compose`
- preflight: container-runtime readiness check that starts a stopped Podman machine only with user approval (never auto-inits one)
- emulators: certify Azurite and PostgreSQL for Podman; others emit a LIMITED SUPPORT warning under Podman
- UI: treat Podman Compose prereq rows as unknown-status like Docker Compose
- evals: parse the runtime from the plan, add a task compose-command conformance check, an --assert-runtime grader flag, and a Podman plan-gate stimulus; refresh agent-assets baseline
- docs: update the create-project guide (Stage 6 + troubleshooting)


nturinski:
Pulled the branch down and exercised it rather than reading the diff. The Podman work itself is solid — findings are mostly about what runs vs. what's written.

## Verified

All nine eval gates pass on a clean checkout: `typecheck`, `drift`, `gates`, `certify`, `stacks:check`, `imports:self-test`, `clean-machine:check`, `msbench:self-test`, `lint`. `drift` passing matters here — the `agent-assets.lock.json` regeneration matches the `resources/agents/**` edits. Product side, `npm run build:check` (tsc --noEmit) is clean.

I tested the new grader flags against purpose-built orchestrator fixtures rather than trusting them — 13 cases, all correct:

| fixture | flags | exit |
| --- | --- | --- |
| podman plan | `--assert-runtime=podman` | 0 |
| docker plan | `--assert-runtime=podman` | 1 |
| podman plan | `--assert-runtime=docker` | 1 |
| **old single-column plan** | *(none)* | 0 |
| **old single-column plan** | `--assert-runtime=podman` | 1 |
| runtime/command mismatch | *(none)* | 1 |
| `Podman (Docker-compatible)` + `docker compose` | *(none)* | 0 |
| docker-compat plan | `--assert-no-docker-compat` | 1 |
| podman plan | `--assert-runtime=containerd` | 3 |
| podman plan | both compat flags | 3 |

The backward-compat pair is the one I'd have expected to get wrong and it's right: an old plan passes with no flags, and fails when asked to prove a runtime it has no data for. Miswired flags exit 3 rather than passing silently. `validateComposeCommandConformance` closing the loop from plan runtime → generated `tasks.json` is the right call — parsing the compose command without enforcing it would have been decoration.

The named-volume ripple-through is complete. I checked all four sites the commit message claims: `postgres.md` (volume + top-level `volumes:` key), `generate.md` (exclusions now bind-mount-only), `preflight.md` (stale check covers `volume ls` / `down -v`), and all three `runtimes/*.md` clean scripts. No leftover `./.postgres` bind mount anywhere except the ⛔ note explaining why not to use one.

The install-link ordering is a genuinely good catch — `Podman (Docker-compatible)` contains "docker", so Podman rules had to precede Docker rules, and there's a test pinning exactly that.

## The Podman stimulus never runs

`debug-plan-podman-runtime` is added to `evals/local-dev/eval.yaml`, which has no `executor:` key — it's lint-validated only. Both of its siblings have MSBench ports:

| stimulus | in `local-dev/eval.yaml` | MSBench port |
| --- | --- | --- |
| `debug-plan-approval-gate` | yes | `debug-plan-approval-gate.yaml` |
| `debug-generate-artifacts` | yes | `debug-generate-artifacts.yaml` |
| `debug-plan-podman-runtime` | yes | **none** |

So the whole Podman path has validator capability and zero execution. Everything above about `--assert-runtime` working is a statement about the grader, not about what the agent actually writes when a user asks for Podman.

Worth being explicit about this in the description, because "add a Podman plan-gate stimulus" reads as coverage. It's the same trap I hit on #1763 — a stimulus that exists but never runs is indistinguishable from a passing one on a summary sheet.

## `validateComposeCommandConformance` can be evaded by a legal task shape

The check reads `task.command` and matches `\b(docker|podman)[ -]compose\b`. VS Code also accepts the split form, and that form slips through. Tested on a Podman plan with a `docker compose` task written both ways:

```
command: "docker compose up -d"                    exit=1  CAUGHT
command: "docker", args: ["compose","up","-d"]     exit=0  not caught
```

`generate.md` § 171 does direct the whole command into `command`, so the documented shape is the caught one and this is unlikely in practice. But it's a check whose coverage depends on the agent picking one of two valid encodings, and nothing enforces that choice. Reading `args` when `command` is a bare engine name would close it.

## Description covers Phase 1; the branch is Phases 1–3 plus the hardening

Title says `Phase1` and the body opens "Phase 1 of Podman support", but the branch carries five commits:

```
e8e9dcf  Phase 1  add Podman as a runtime option
b5c96d5  Phase 2  prefer Podman when it's ready
cb2a48f  Phase 3  Docker-compatibility mode
382462a           install links
4ec0a7b           harden Podman on Windows
```

The bullet list does cover most of the surface, but Phases 2 and 3 aren't named, and neither is the hardening — which is arguably the most valuable commit here, since it's two real bugs found in live testing (rootless-Podman `initdb` failing on a Windows bind mount, and the agent silently switching a user-requested Podman plan to Docker). The second one is a genuine agent-behaviour defect and the lifecycle-wide STOP-and-ask rule is the right fix. Both deserve to be in the description rather than buried in a commit body.

## Not this branch's problem

Root `npm run lint` fails with 31 errors, all in `evals/grader-certification` (`reference-node-postgres`, `refusal-bait/draft-server.ts`). That's the pre-existing `feat/CoR` breakage — #1768 fixes it by ignoring the fixture tree. This branch adds zero new lint errors; I confirmed by clearing a local staged-graders artifact that was inflating the count to 80.

## Overall

The engineering is careful and the two Windows findings are real. My only substantive ask is the missing MSBench port — without it, "Podman is supported" rests on instructions and a validator that never sees agent output. The args-form gap and the description scope are minor.

nturinski:

PR #1766, opened 2026-09-01
Add 8 plan-phase scenario stimuli and two requirements grader flags

patverb:
Adds 8 plan-phase stimuli and two `validate-requirements.ts` flags. All 8 were run against live agents before this PR was opened.

## Why

The plan-phase corpus covered four app shapes — a photo app, an API-only inventory service, a no-datastore converter, a task app. Ordinary requests a user would actually make (I need login, keep the data in the EU, we already use MySQL) had never been put in front of the graders.

## What was run

| stimulus | constraint under test | check | run | result |
|---|---|---|---|---|
| `auth-expense-tracker` | identity is required | `--assert-auth-includes=Entra` | `2026083158957999` | 5/5 |
| `ai-chatbot-documents` | document storage | `--assert-blob-storage` | `2026083160386928` | 5/5 |
| `scheduled-nightly-report` | no frontend at all | `--assert-no-frontend` | `2026083160830938` | 5/5 |
| `mysql-cms` | unlisted database | probe (see below) | `...59733325` / `...77939774` | see below |
| `cost-constrained-blog` | budget ceiling | schema + probe | `2026083161269955` | 5/5 |
| `private-internal-hr-tool` | no public ingress | schema + probe | `2026083161668243` | 5/5 |
| `realtime-whiteboard` | persistent connections | schema + probe | `2026083162132945` | 5/5 |
| `eu-data-residency-portal` | EU data residency | schema + probe | `2026083165493260` | 5/5 |

Four carry semantic assertions. Four record probes instead, because the constraint reaches `requirements.json` only as free text — asserting a field the artifact does not structure yet would be asserting a shape rather than a behaviour.

## Grader flags

`--assert-datastore-includes=<substring>` generalises the hard-coded Cosmos and Blob flags, which were on track to become one flag per store. `--assert-auth-includes=<substring>` reaches the shared `auth` question, which is mandatory in every valid artifact (`requirements.ts` raises `missingAuth` without it) but which no scenario had ever answered with anything but "No auth" — so the identity path was untested.

Both take substrings so a stimulus can say `MySQL` or `Entra` without pinning Azure marketing strings. Both throw on an empty needle, so a flag that *cannot* fail is not mistaken for a flag that passed. Each was tested positive **and** negative before use.

Covered by 14 new certification mutations (95 -> 109). Certification: **139/139** (137 mine + 2 from #1755, post-rebase).

## `mysql-cms` ships as a probe, and found a product bug

Prompt says *"please use MySQL for the database."* Two runs, same prompt, opposite outcomes:

| run | `recommendedChoice` | |
|---|---|---|
| `2026083159733325` | `["MySQL", "Blob Storage"]` | invented a MySQL option — user intent kept, spec broken |
| `2026083177939774` | `["Blob Storage"]` | obeyed the fixed list — spec kept, **user intent lost** |

The agent understood the constraint in both (28 MySQL mentions; explicit *"Tech constraint: MySQL"*). Run 2's own rationale: *"Team prefers MySQL for relational data (Azure Database for MySQL), which will be configured in the plan"* — and then it emitted no relational datastore at all, for a CMS.

Root cause is not the agent. `azure-project-plan/requirements.md` line 133 fixes the `dataStores` options to `No datastore required, Blob Storage, Queue Storage, PostgreSQL, CosmosDB, Redis, Azure SQL` with `allowFreeformInput: false`. MySQL is absent, so the choice is unsatisfiable and the agent improvises differently per run. Meanwhile `azure-deploy` supports MySQL across **17 files** — `scaffold-conformance.ps1` (9 mentions, incl. three MySQL-only BLOCK rules), `bicep-patterns-data.md` (6), `database-post-deploy.md` (5), `service-mapping.md` (4), `sku-quota-validation.md` (3). The pipeline is MySQL-capable end to end *except at the point where the user states the requirement.*

Asserting `MySQL` would be correct and red about half the time, which a shared suite cannot carry. So it is recorded, not asserted; the evidence still lands in `session.sqlite` every run. **Adding MySQL to that option table re-arms it by deleting one line.** Filed separately rather than fixed here to keep this PR to test cases. Two runs establish that both behaviours occur — they do not establish a rate.

## Second finding

The agent **preserves data-residency constraints but drops network-isolation ones.** `eu-data-residency-portal` carried 6x EU / 6x residency / 2x region into `requirements.json`; `private-internal-hr-tool` carried only 4x "internal", with zero `private` / `vnet` / `endpoint` / `firewall`, despite an equally plain "must not be reachable from the public internet". A selective loss, not a general one. Found by a probe — an assertion would have said nothing.

## Validation

All 9 checks green: `gates` (23 stimuli), `typecheck`, `stacks:check`, `imports:self-test`, `clean-machine:check`, `drift`, `msbench:self-test`, `seed:self-test`, `certify` 139/139.

## Merge note

Rebased onto `feat/CoR` after #1755 landed. The only conflicts were the two *generated* certification reports, resolved by regenerating rather than picking a side. `manifest.json` auto-merged: my 14 mutations and #1755's 2 coexist at **139/139**, and `gates` now reports 19 staged graders (up from 13) with #1755's runtime graders alongside these. Full suite re-run green after the rebase.

#1755 adds certification manifest entries around L894; these are at the tail. No conflict expected, but whoever lands second should rebase.

### DIFF
--- a/evals/grader-certification/manifest.json
+++ b/evals/grader-certification/manifest.json
@@ -1167,6 +1167,164 @@
             "operation": "relocate",
             "replacement": "infra/main.tf",
             "expectedCode": "!noIacFound"
+        },
+        {
+            "comment": "The preview manifest is absent entirely. Until now the only preview case was previewStatus:draft, which means every failure path that runs *before* a manifest is in hand — the file missing, the file unparseable — was uncertified, and `preview` was the least-exercised validator we ship at one single case. A missing manifest is the ordinary shape of a preview step that never ran, so this is the first thing the validator should be able to say.",
+            "id": "preview-manifest-absent",
+            "tier": "offline",
+            "fixture": "sample-agent-output",
+            "validator": "preview",
+            "file": ".azure/.preview-temp/manifest.json",
+            "operation": "delete",
+            "expectedCode": "missingPreviewManifest"
+        },
+        {
+            "comment": "The manifest exists but is not JSON. Distinct from the case above because the remedy is different: a missing file means the preview step did not run, a corrupt one means it ran and produced garbage. Collapsing the two would send the reader to look for a step that plainly executed.",
+            "id": "preview-manifest-unparseable",
+            "tier": "offline",
+            "fixture": "sample-agent-output",
+            "validator": "preview",
+            "file": ".azure/.preview-temp/manifest.json",
+            "operation": "replace",
+            "search": "\"previewStatus\": \"ready\",",
+            "replacement": "\"previewStatus\": \"ready\",,",
+            "expectedCode": "invalidPreviewManifest"
+        },
+        {
+            "comment": "Paired with the case above, and the reason the pair exists: previewStatus is read *out of* the manifest, so a manifest that does not parse cannot also be meaningfully 'not ready'. Reporting previewNotReady here would name a status field inside a file the validator was never able to read. Negative because includes() cannot falsify a spurious extra code.",
+            "id": "preview-unparseable-is-not-called-not-ready",
+            "tier": "offline",
+            "fixture": "sample-agent-output",
+            "validator": "preview",
+            "file": ".azure/.preview-temp/manifest.json",
+            "operation": "replace",
+            "search": "\"previewStatus\": \"ready\",",
+            "replacement": "\"previewStatus\": \"ready\",,",
+            "expectedCode": "!previewNotReady"
+        },
+        {
+            "comment": "A manifest that parses, declares itself ready, and lists no pages. This is the empty-success case — the preview step reports completion having rendered nothing — and it is worth its own case because status and content are independent claims: 'ready' is the agent's assertion, pages[] is the evidence for it. Renaming the populated array rather than emptying it in place keeps the mutation to a single line while leaving the JSON valid.",
+            "id": "preview-ready-with-no-pages",
+            "tier": "offline",
+            "fixture": "sample-agent-output",
+            "validator": "preview",
+            "file": ".azure/.preview-temp/manifest.json",
+            "operation": "replace",
+            "search": "\"pages\": [",
+            "replacement": "\"pages\": [], \"ignoredPages\": [",
+            "expectedCode": "missingPreviewPages"
+        },
+        {
+            "comment": "A slug that is neither lowercase nor hyphen-separated. Slugs become filenames — the validator resolves `${slug}.html` on disk — so a slug carrying uppercase or an underscore is a page that cannot be found on a case-sensitive filesystem even though the manifest swears it is there. Certifying the regex matters because it is the only thing standing between a manifest entry and a broken link.",
+            "id": "preview-slug-not-kebab-case",
+            "tier": "offline",
+            "fixture": "sample-agent-output",
+            "validator": "preview",
+            "file": ".azure/.preview-temp/manifest.json",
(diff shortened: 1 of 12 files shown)


nturinski:
Pulled this down and ran it rather than just reading the diff. Notes below.

## Checked

All 9 checks pass on a clean checkout with a fresh `npm install` — typecheck, gates, certify, stacks:check, drift, imports:self-test, msbench:self-test, seed:self-test, clean-machine:check. Certify is 139/139 as stated.

I also exercised both new flags directly against `sample-agent-output` (which has `File storage` / `No auth`):

| case | exit |
| --- | --- |
| `--assert-datastore-includes=File` (match) | 0 |
| `--assert-datastore-includes=MySQL` (no match) | 1 |
| `--assert-datastore-includes=` (empty) | 3 |
| `--assert-auth-includes=No` (match) | 0 |
| `--assert-auth-includes=Entra` (no match) | 1 |
| `--assert-auth-includes=` (empty) | 3 |

Exit 3 on the empty needle is the right call — harness fault rather than a silent pass. That plus substring-over-equality with the reasoning written down is the part of this I like most.

Structure is clean across all 8: liveness sentinel present, positive artifact anchor, webview and chat-tool assertions, and gate identity passes so nothing forked a gate.

## Two things

**1. The prose says four semantic assertions; I count three.**

Only `auth-expense-tracker` (Entra), `ai-chatbot-documents` (Blob) and `scheduled-nightly-report` (no frontend) assert their constraint. The other five — `mysql-cms`, `cost-constrained-blog`, `private-internal-hr-tool`, `realtime-whiteboard`, `eu-data-residency-portal` — assert only `requirements.json satisfies the requirements contract`, which is the generic schema check every plan-phase stimulus already runs.

The table in the description actually shows 3/5; it's the sentence above it that says "Four carry semantic assertions. Four record probes." Can we reconcile those? It matters because five of the eight don't test the thing their filename implies — `eu-data-residency-portal` passes with zero mention of the EU anywhere in the artifact. That's fine as regression coverage for "a well-formed artifact came back", but someone skimming the file list will read it as eight scenarios under test when it's three under test and five under observation.

**2. `--assert-datastore-includes` has no asserting consumer in this PR.**

Its only use is inside the `mysql-cms` probe, which carries `assertZeroExitCode: false` and so can't fail a run. The flag is certification-tested but never actually gates anything yet. I get why given mysql-cms ships as a probe, and it arms the moment MySQL lands in the option table — just flagging that it's inert on merge.

## Not checked

The 8 run IDs. Those results are in your local msbench cache, so `5/5` on each is taken on trust from here, and the MySQL product-bug finding rests on the same two runs. The reasoning is specific and plausible — the `requirements.md` line 133 option table not carrying MySQL while `azure-deploy` supports it across 17 files is a good catch — but I can't independently confirm it. Happy to take it as-is; noting it so it's on the record.

## Overall

The discipline here is good, and choosing a probe over an assertion that would be "correct and red about half the time" is exactly right — an assertion that flaky would have got deleted within a month. The network-isolation finding is a genuine argument for probes earning their place.

My only real ask is fixing 4/4 to 3/5 in the description so the coverage claim matches what the files do.

nturinski (approved):

nturinski:
LOL ignore my...verbose comments. Just wanted it note that 5 of the scenarios are still under observation.

PR #1769, opened 2026-09-01
debug-probe: make the live tier (the negative controls) runnable on Windows

nturinski:
You asked to approach `debug-breakpoint` backwards: rather than trying to make it pass, prove it can **fail** — on the grounds that a gate which only ever passes is too easy to fool. That turned out to be the right question, and the answer is not what the 0-for-16 suggested.

## The gate does discriminate — proven end to end

`certify.ts --live` runs real VS Code and real js-debug against the known-good fixture plus six mutations. All seven behave correctly:

```
known-good                       hit                    exit 0
mutation-launch-config-renamed   launchConfigInvalid    exit 1
mutation-app-crashes-on-boot     appFailedToStart       exit 1
mutation-breakpoint-unreachable  breakpointNotHit       exit 1   <- the gate-can-fail proof
mutation-pattern-matches-nothing patternMatchedNothing  exit 3
mutation-port-squatted           probeError             exit 3
mutation-debug-adapter-missing   probeError             exit 3
```

The load-bearing one puts the breakpoint on the `POST` 400 branch while triggering `GET /api/health`. It goes red, and red for the right reason. So the gate is not passing on false positives — it discriminates all six outcomes, and separates product failures (exit 1) from harness faults (exit 3).

## But nobody on Windows could run any of that

```
FAIL  [live] known-good: VS Code did not finish: spawnSync code ENOENT (probe did NOT activate)
FAIL  [live] mutation-launch-config-renamed: ... (probe did NOT activate)
... x7
0/7 certification cases passed
```

`spawnSync('code', …)` works where `code` is a shell script. On Windows it is `code.cmd`, which `spawnSync` cannot execute — so the tier died before a single case ran.

That wording is why it survived: seven identical *"probe did NOT activate"* lines read as a broken probe, not as a runner that never started one. The only way through was already knowing to pass `--vscode=`. Same shape as the `mktemp -t` note in `msbench/README.md` — shipped broken, stayed broken, because every machine was macOS.

## The fix

Resolve `Code.exe` from whatever `code.cmd` is on PATH (`<root>/bin/code.cmd` → `<root>/Code.exe`), then the usual install locations. `Code.exe` rather than the `.cmd` because it is directly executable; shelling out would push a path containing `Microsoft VS Code` through cmd quoting for no benefit.

Resolution is **lazy**. It throws when it finds nothing, and the offline tier is the CI tier and needs no VS Code — resolving eagerly would make `--offline` fail on exactly the machines it exists for. Verified both ways.

On Windows, with no flags:

```
11/11 offline   blame mapping
 7/7  live      real VS Code, real js-debug
18/18
```

README: corrected the stale case counts (4 → 6 mutations, 5 → 7 cases) and recorded the Windows resolution.

## What actually causes the 0-for-16 — not fixed here

The gate is wired `phases: [local]` in `config/gates.yaml`. `config/phases/local.yaml` **never writes `debug-probe.json`** — only `config/phases/debug-breakpoint.yaml` does:

```
config/phases/debug-breakpoint.yaml:  cat > /workspace/debug-probe.json <<'JSON'
config/phases/local.yaml:             (nothing)
```

With no spec the probe idles, no verdict is written, and the grader correctly calls that a harness fault. Every stack run, forever. This README predicted it exactly:

> *"With no `debug-probe.json`, the gate reports exit 3 forever… that is the 'gate that cannot pass' shape displaced into the exit-3 column."*

…and the gate's own `requires: {}` row is annotated `# unreviewed`. The fix is a wiring decision — either the local phase writes a spec derived from a per-stack breakpoint declaration (so the gate finally runs behind real generated output), or it is unwired from `local` and declared a `knownGap`. Deliberately left out of this PR.

### DIFF
--- a/evals/debug-probe/README.md
+++ b/evals/debug-probe/README.md
@@ -150,18 +150,27 @@ A gate that cannot fail is not a gate, so both directions are proven.
 
 ```bash
 node certify.ts --offline     # 11 cases, no VS Code, ~2s — safe for CI
-node certify.ts --live        # 5 cases, real VS Code + real js-debug
+node certify.ts --live        # 7 cases, real VS Code + real js-debug
 node certify.ts               # both
 ```
 
 Useful flags: `--only=<case-id>`, `--verbose`, `--vscode=/path/to/code`.
+
+**On Windows `code` is a `.cmd`**, which `spawnSync` cannot execute, so the live
+tier resolves `Code.exe` instead — from whatever `code.cmd` is on PATH, then the
+usual install locations. Before that it died with `spawnSync code ENOENT` before
+a single case ran and reported it as seven identical *"probe did NOT activate"*
+failures, which reads as a broken probe rather than a runner that never started
+one. The live tier had therefore never been run on Windows at all. Resolution
+happens only when the live tier is actually selected, so `--offline` still needs
+no VS Code.
 
 **Offline** synthesises verdict files and asserts the grader's exit code for
 each, including the ones that must never blame the product: a missing verdict, a
 malformed verdict, an unknown outcome, and a schema-version mismatch. This
 certifies the part that assigns blame, so it runs anywhere.
 
-**Live** stages the known-good fixture plus four mutations and runs real VS Code
+**Live** stages the known-good fixture plus six mutations and runs real VS Code
 against each. The load-bearing one is `mutation-breakpoint-unreachable`, which
 puts the breakpoint on the `POST` 400 branch while triggering `GET /api/health`:
 it must go red, and red for the right reason.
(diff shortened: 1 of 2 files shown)


nturinski:
Folded into #1774, which needs this fix to run its live certification cases and so carries it directly. Closing to avoid two PRs touching `certify.ts`.

The change and its reasoning are unchanged in #1774: on Windows `code` is a `.cmd` that `spawnSync` cannot execute, so the live tier died with `spawnSync code ENOENT` before a single case ran — and reported it as seven identical *"probe did NOT activate"* failures, which reads as a broken probe rather than a runner that never started one.

PR #1770, opened 2026-09-01
Stage fixtures: harvest a representative scaffold from a real run, and fix the first thing it found

nturinski:
Fixtures the harness owns, one per project stage, harvested from real runs — so gates can be exercised against a realistic workspace for free, and live runs are spent confirming rather than discovering. This is the first stage.

## The problem

The certification fixtures are broad — **18 of 21** wired gates can be exercised offline — and **unrepresentative**. `reference-node-fullstack` is a single package whose `src/server.js` serves its own `public/`. Every project the scaffold agent emits is an npm-workspaces monorepo: `services/functions` behind `func`, `services/web` behind its own Vite dev server, `services/shared` hoisted to the root `node_modules`.

So a gate can be green in certification and near-zero in production, with nothing connecting the two numbers:

| Gate | certification | production |
|---|---|---|
| `runtime-frontend-api` | green | **0 / 17** |
| `runtime-frontend` | green | 1 / 11 |
| `runtime-crud` | green | 1 / 11 |

The `dependenciesInstalled` defect fixed in #1759 is the same shape — it could only occur under workspace hoisting, no fixture had workspaces, so certification couldn't see it and two paid runs found it instead.

## What this adds

`evals/msbench/harvest-stage.ts` reconstructs a fixture from a finished run's `patch.diff`. The container commits the base workspace, so the diff is exactly what the agent wrote; `node_modules`, `.github/agents` (152 staged instruction files) and `.eval` are dropped on top since none are agent output. `--check` re-harvests and reports drift.

`stage-scaffold` comes from run [`2026083170541011`](https://msbenchapp.azurewebsites.net/run-analysis/2026083170541011) — the verified-good run where `project-builds`, `runtime-app-starts` and `runtime-health` all passed. 58 files: `services/web` (29), `services/functions` (17), `services/shared` (6), all three `.azure` artifacts, and a root `package.json` declaring `workspaces: ["services/*"]`.

Measured through the real code paths:

| | `stage-scaffold` | `reference-node-fullstack` |
|---|---|---|
| npm workspaces | `["services/*"]` | none |
| frontend directory | `services/web` (separate) | none — app serves it |
| runtime target | `functionsHostUnavailable` | `node src/server.js` |
| safety scan | 59 files, **1 violation** | 19 files, 0 violations |

## The first thing it found

That last cell. On its first scan the fixture tripped `hardcodedSecret` on:

```
AZURE_STORAGE_CONNECTION_STRING=DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM0…
```

That is the **published Azurite emulator key** — it ships in Microsoft's own documentation, only ever authenticates against a local emulator, and is exactly what a correct local-dev scaffold should emit. `safety-boundaries` called it a product failure.

A live false red in a security gate, found in minutes, that the idealised fixtures could not surface because none of them configure a storage emulator.

Allowlisted as the key itself rather than as `devstoreaccount1`: suppression is span-scoped since #1767, and the account name sits earlier in the line so it would not overlap the `AccountKey=` the rule matches. Verified both ways — the fixture scans clean at 0 violations, and a non-emulator `AccountKey` in the same shape is still reported.

## The design constraint

A stage fixture is an **input**, never an expected answer. `harvest-seed.ts` settled this and it carries over verbatim:

> *"the plan is an input, not an expected answer, so a stale one makes scaffold trials fail **loudly** and cannot make a broken scaffolder look green. Fail-safe, not fail-open."*

## Recorded, not fixed here

- **`safety-boundaries` and `webview-parseable` have no entry in `OFFLINE_VALIDATORS`** — which is why they read as uncovered, not for want of a fixture. `safety-boundaries` is a security gate that has now shipped two silent-green defects (#1755, #1767) and certification has never once executed it. Adding the registry entries is the natural follow-up.
- **`debug-plan` passes as a standalone grader against this fixture and fails inside certification** (`unexpectedStatus, unexpectedServiceCount, unexpectedEmulator, missingChecklist`). The certified path and the executed path disagree — the drift the shared-contract design exists to prevent.
- **`service-fidelity`, `datastore-fidelity` and `project-plan` all fail against real agent output.** Either the run's plan and tree genuinely disagree, or the gates are too strict. The fixture now makes that answerable for free.

## Next stages

`stage-plan`, `stage-local-dev` (adds `.vscode/**` + `debug-probe.json`) and `stage-deploy` follow the same path. `stage-local-dev` is the one that would let `debug-breakpoint`'s live tier run against a realistic project instead of only the known-good reference — and would let frontend-dev-server support for `runtime-frontend`/`-api` be built offline instead of a paid run per iteration.

`drift`, `typecheck`, `imports:self-test`, `certify` (127/127), `stacks:check`, `gates`, `lint`, `clean-machine:check`, `stage:check` all pass.


nturinski:
Superseded by #1774, which rebuilds this work on current `feat/CoR`.

Two reasons this is closed rather than merged:

**The harvesting machinery was not load-bearing.** I removed `harvest-stage.ts`, the `stage-scaffold` fixture and its npm scripts from the top of the stack and re-ran every suite — `drift`, `typecheck`, `imports:self-test`, `certify`, `stacks:check`, `gates`, `lint`, `clean-machine:check` all still pass. Nothing in the debug-breakpoint work references them, and `stage:check` was in no CI workflow.

**It had gone stale and become unsafe.** This branch predates #1767, so merging it would have reverted the safety-boundaries liveness fix.

What was worth keeping is carried into #1774: the `stage-local-dev` fixture and the Azurite emulator-key allowlist, which fixed a live false red in the security gate.

Harvesting is paused deliberately — the priority is proving these tests can run at all. `harvest-stage.ts` and `stage-scaffold` can come back when we actually want to harvest; the reasoning is preserved in this PR's history.

PR #1771, opened 2026-09-01
stage-local-dev: stop debug-breakpoint blaming the product for Azure Functions projects

nturinski:
Second stage fixture. Stacked on #1770 (needs `stage-scaffold`) and includes #1769 (`resolveVsCode`, without which the live tier cannot run on Windows and this change is certified by a live case).

It earned itself immediately: pointing the probe at a realistic project showed `debug-breakpoint` **fabricating a product failure against every correctly generated Azure Functions app**.

## What it found

```
resolved services/functions/src/functions/health.ts:26
startDebugging returned false          <- 64 seconds later
outcome: appFailedToStart              <- exit 1, a PRODUCT failure
```

A Functions launch configuration is `request: attach` with `preLaunchTask: "func: host start"`. The probe starts nothing and installs no extensions, so there is no process to attach to. **Removing `preLaunchTask` changed nothing**, which is what identifies `request` rather than the task provider as the cause.

This is the same shape as the missing-debug-adapter defect already documented in the README — a configuration this environment cannot execute, reported as a project that does not work — and it was left one field away. The probe checked `type` and never `request`.

Now pre-checked before launch, alongside the adapter check, returning `probeError` (exit 3). Verified on the same fixture: `appFailedToStart` exit 1 → `probeError` exit 3, in seconds rather than 64.

Certified by a new live case `mutation-attach-config-is-declined`, which takes the known-good fixture to the attach shape and requires exit 3.

```
8/8   live      (was 7/7)
129/129 offline (was 127/127)
```

## What the fixture is

`stage-scaffold` plus `.vscode/launch.json`, `.vscode/tasks.json`, `debug-probe.json`.

Those three are **hand-written**, and that is the design constraint. They are derived from what the harvested `.azure/vscode-debug-plan.md` already specifies — "Task Tracker API (debug)" on `:7071`, "Task Tracker Frontend (debug)" on `:5173`, the compound config, `GET /api/health` — so they are a faithful realisation of a real agent-written plan rather than an invention.

But they are still ours, so the fixture **deliberately does not certify `debug-config` or `debug-artifacts`**. Those grade `.vscode/**` against the plan; certifying them here would grade our own fixture and call it a product verdict.

*(Aside: the breakpoint pattern I first wrote, `status:\s*'healthy'`, matched nothing — the real source says `status = 'healthy'`. Exactly the kind of thing a realistic fixture surfaces for free.)*

## Why it could not be harvested

`local.yaml` runs one graded turn under `azure-debug-plan`, which writes the plan and **stops at the approval gate** — which is what `debug-gate` certifies. But `debug-config`, `debug-artifacts` and `debug-breakpoint` are wired into that same phase and need what `azure-debug-generate` produces *after* approval.

So `debug-gate` passing and `debug-config` failing are the same fact graded with opposite polarity, three gates are structurally unsatisfiable in the phase they are wired to, and the corpus has nothing to harvest: **1/10, 1/10 and 0/10** passes respectively.

Fixing it needs a two-turn local phase, which `build-config.ts` currently forbids:

> `config/phases/X.yaml declares both turn-instead and turn-after; only one may replace the stack prompt.`

Left for its own change.

## Three structural findings, same conclusion

This is the third finding about this gate's wiring, after the missing `debug-probe.json` in the local phase (#1769) and the phase stopping at the approval gate above. All three say the same thing: **the gate is sound and is wired where it cannot be asked a question.**

It also means the gate cannot currently answer for the Functions stack at all, since every correctly generated Functions project takes the attach path. That is a `knownGap`, not a red.

`drift`, `typecheck`, `imports:self-test`, `certify`, `stacks:check`, `gates`, `lint`, `clean-machine:check`, `stage:check` all pass.


nturinski:
## Correction: attach configurations *are* driveable

The preflight in `11c57e2` declined every `request: attach` configuration. That was too broad, and it would have written off the entire Azure Functions stack on a false premise.

I tested the premise instead of assuming it. Against an attach config whose `preLaunchTask` is a `type: shell` background task starting `node --inspect=9229` — the same shape as `func: host start`, with nothing that needs an extension:

```
preLaunchTask "serve" resolves in this environment
startDebugging returned true
trigger connected after 1 attempt(s)
stopped: reason=breakpoint session=Remote Process [0] << Attach Under Background Task
outcome: hit
```

VS Code runs the `isBackground` preLaunchTask, waits for its problem matcher to report ready, and attaches. **`request` was never the problem.**

## What actually breaks

A task type this environment cannot provide. An Azure Functions project declares `preLaunchTask: "func: host start"` with `"type": "func"`, and that provider comes from `ms-azuretools.vscode-azurefunctions` — which the agent's own `vscode-debug-plan.md` lists as a prerequisite, and which is not installed in the probe environment.

The preflight now fetches available tasks and declines only when the config's `preLaunchTask` is not among them. On `stage-local-dev`:

```
probeError: launch configuration "Task Tracker API (debug)" declares preLaunchTask
"func: host start", which this environment cannot resolve — the extension that
provides that task type is not installed. ... Tasks available here: api: build,
web: dev server, install, build - services/web/tsconfig.json, ...
durationMs: 1004
```

One second instead of 64, and specific enough to act on: every *other* task in that workspace resolves, so the diagnosis is the `func` type alone.

## Two live cases, not one

The positive control is what stops this mistake being made again silently:

| case | outcome |
|---|---|
| `mutation-attach-with-resolvable-task-is-driven` | `hit`, exit 0 |
| `mutation-prelaunch-task-unresolvable` | `probeError`, exit 3 |

Live tier **9/9**.

## This changes the conclusion about Functions

My previous comment said the gate "cannot currently answer for the Functions stack at all." That was wrong. It needs two things, and the custom image `msbench-1.1.0` already carries one of them:

- `func` on PATH — **already present** in the custom image
- `ms-azuretools.vscode-azurefunctions` installed in the probe environment — not yet

So making `debug-breakpoint` answer for Functions projects is now a scoped, tractable change: install that extension alongside the probe (`base.yaml`'s `installExtensions` already concatenates across config layers, which is how the probe itself rides in). Worth doing as its own PR once this lands, since it needs a run to confirm.

Until then the gate declines rather than inventing a verdict — same ruling as the missing-adapter check, same reason.

All suites still green: `drift`, `typecheck`, `imports:self-test`, `certify`, `stacks:check`, `gates`, `lint`, `clean-machine:check`, `stage:check`.

nturinski:
Superseded by #1774, which rebuilds this work as a single change on current `feat/CoR`. See #1770 for why the stack was replaced rather than merged.

PR #1773, opened 2026-09-01
Inverted breakpoint control for Functions: assert NOT hit, so a failure is the proof

nturinski:
Stacked on #1772. An **inverted** negative control for the Functions debug path, plus the two findings it produced.

## Why the assertion is inverted

Asserting the breakpoint *is* hit cannot distinguish success from silence. This suite has been caught by that shape four times: a gate wired into a phase that never writes its `debug-probe.json` (#1769), seven graders wired but never staged (#1755), a safety gate whose file count was satisfied by the harness's own staged instructions (#1767), and every gate in a `RATE_LIMIT` run. In all four, **nothing happened and everything passed**.

So this stimulus asserts the breakpoint is **NOT** hit:

```yaml
- comment: INVERTED CONTROL; passes only if NO breakpoint was hit, so a failure here is the proof
  exec: '! node .../validate-debug-breakpoint.ts'
```

A green could mean many kinds of nothing. A **red can only mean** the grader ran, exited 0, and a debugger stopped on a line of a Functions project in the container. Absence of action produces a pass; only the thing we want produces a failure.

Paired with a second assertion reading `outcome` from the verdict artifact, because `!` inverts exit 1 and exit 3 alike and cannot tell a miss from a decline.

## Finding 1 — the extension works in the container

Run [`2026090178170878`](https://msbenchapp.azurewebsites.net/run-analysis/2026090178170878) confirms #1772 end to end:

```
resolved services/functions/src/functions/health.ts:12
preLaunchTask "func: host start" resolves in this environment   <- previously declined every Functions project
```

`func`, `azurite`, `psql` and a built `dist/` all present.

## Finding 2 — `local.settings.json` can never be harvested

The run then spent its full 240s budget:

```
outcome: appFailedToStart
detail:  startDebugging(...) never resolved within the probe budget —
         an unfinishable preLaunchTask is the usual cause
```

`func start` will not run without `services/functions/local.settings.json` — it declares `FUNCTIONS_WORKER_RUNTIME`, `AzureWebJobsStorage` and the app settings the generated code reads.

It is also the file the scaffold agent is **instructed to gitignore** from the first commit, because it holds connection strings.

Both are correct, and together they mean **a harvested Functions fixture is always unrunnable**: `harvest-stage.ts` reads the agent's `patch.diff`, and a gitignored file is not in it. No amount of re-harvesting fixes that.

Written by hand here, with a fixture README recording which files are inputs and why each cannot be harvested. `languageWorkers__node__arguments` opens the inspector on 9229 — the port `launch.json` attaches to, and the link that makes `request: attach` work at all.

## Also here

The `functions-breakpoint` phase seeds the fixture, starts Azurite and PostgreSQL, installs `func` if absent, and pre-runs `npm install` + the TypeScript build. The task chain still runs — it just finds its work done. That is the difference between measuring "can this be debugged" and "does npm finish in 240 seconds".

The breakpoint sits on `getServices()`, the first statement of the health handler, not on `status = 'healthy'` — the latter only executes when both emulators answer, so a stopped emulator would make it unreachable and the verdict would read as "the debugger cannot stop here" rather than "the datastore is down".

## Status

Whether a breakpoint is now reached is **unproven** — that is the next run. What is proven is that the previous blocker is gone and the next one is named. The control is designed so that run's result will be unambiguous either way.

`drift`, `typecheck`, `certify`, `gates`, `lint`, `stage:check` all pass.

### DIFF
--- a/evals/debug-probe/README.md
+++ b/evals/debug-probe/README.md
@@ -359,10 +359,40 @@ outcome: hit
 Certified by `mutation-attach-with-resolvable-task-is-driven`, which exists so
 that mistake cannot be made again silently.
 
-**So the Functions stack is answerable here** — it needs
-`ms-azuretools.vscode-azurefunctions` installed in the probe environment and
-`func` on PATH, which the custom image (`msbench-1.1.0`) already carries. Until
-the extension is installed the gate declines rather than inventing a verdict.
+**So the Functions stack is answerable here.** It needs two things, and the
+custom image (`msbench-1.1.0`) already carries one:
+
+| Needed | Provides | Status |
+| --- | --- | --- |
+| `ms-azuretools.vscode-azurefunctions` | the `func` **task type**, so `preLaunchTask` resolves | added to [`msbench/config/base.yaml`](../msbench/config/base.yaml) |
+| `func` on PATH | the Functions host itself | already in the custom image |
+
+Measured on `stage-local-dev` with the extension installed into an isolated
+extensions dir — the preflight goes from declining to passing:
+
+```
+preLaunchTask "func: host start" resolves in this environment
+trigger port 127.0.0.1:7071 is free
+```
+
+**A breakpoint hit in a Functions project is not yet proven**, and the remaining
+distance is honest to state: the task chain is
+`func: host start` → `api: build` → `install`, so the project must install and
+compile inside the probe budget, and this fixture's breakpoint sits on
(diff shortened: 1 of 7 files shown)


nturinski:
Superseded by #1774, which rebuilds this work as a single change on current `feat/CoR`. See #1770 for why the stack was replaced rather than merged.

PR #1774, opened 2026-09-01
Prove the debug-breakpoint gate can run against a real Azure Functions project

nturinski:
Replaces the #1770 → #1773 stack with one change, rebuilt on current `feat/CoR`. Everything here serves one question: **can `debug-breakpoint` be asked its question about the stack the agent actually generates?**

Answer: yes, and it now has. Run [`2026090180705733`](https://msbenchapp.azurewebsites.net/run-analysis/2026090180705733) hit a breakpoint in a real Functions project — `outcome: hit — breakpoint hit at services/functions/src/functions/health.ts:12`.

## Why the stack was replaced

Two reasons, both verified:

- **The harvesting machinery was not load-bearing.** I removed `harvest-stage.ts`, the `stage-scaffold` fixture and its npm scripts from the top of the stack and re-ran every suite — all eight still pass. Nothing in the debug work references them, and `stage:check` was in no CI workflow.
- **The stack had gone stale and was dangerous.** It was branched before #1767 landed, so merging it would have **reverted the safety-boundaries liveness fix**.

Size, for context — the bulk is the fixture, not logic:

```
fixture: 8790 lines  (5844 of it package-lock.json)
logic:    544 lines
```

## The fixture

`grader-certification/stage-local-dev`, reconstructed from run [`2026083170541011`](https://msbenchapp.azurewebsites.net/run-analysis/2026083170541011): npm workspaces over `services/{functions,web,shared}`, a Functions API behind `func`, a React frontend on its own Vite dev server.

Every existing fixture is an idealised single package that serves its own frontend — a shape the agent never produces. That is how a gate certifies green and scores near zero in production (`runtime-frontend-api`: green in certification, **0 of 17** in runs).

Its `.vscode/**`, `debug-probe.json` and `services/functions/local.settings.json` are hand-written **inputs**; the fixture README records which and why. It therefore does **not** certify `debug-config` or `debug-artifacts` — grading our own inputs would be a verdict about us.

## Three defects it found

**1. The gate blamed the product for every Functions project.** A Functions config is `request: attach` behind `preLaunchTask: "func: host start"`; the probe started nothing, so `startDebugging` returned false → `appFailedToStart`, exit 1, 64 seconds to say it.

My first fix declined *all* attach configs — wrong, and it would have written off the stack. Attach is driveable:

```
preLaunchTask "serve" resolves in this environment
startDebugging returned true
stopped: reason=breakpoint
outcome: hit
```

So the check is about the **task**, not `request`. The probe now fetches available tasks and declines only when the named one is absent — one second instead of 64. **Two** live cases, because the positive control is what stops the over-broad version coming back.

**2. The `func` task type needs an extension.** `ms-azuretools.vscode-azurefunctions` is now installed in the probe environment — the agent's own debug plan already lists it as a prerequisite. It **depends on the extension under test**, so installing it can replace the VSIX being evaluated; verified that VS Code leaves a satisfied dependency alone, and the entry is ordered last with a comment saying why it must stay there.

**3. `local.settings.json` can never be harvested.** `func start` won't run without it, and the agent is instructed to gitignore it because it holds connection strings. Both correct — together they mean any reconstructed Functions fixture is unrunnable, because a gitignored file isn't in a patch.

## The assertion was inverted, then flipped back

`functions-breakpoint-node.yaml` first shipped asserting the breakpoint was **NOT** hit, so a **failure** was the proof. That is not a stylistic choice: asserting the positive cannot distinguish success from silence, and this suite has been caught by that shape four separate times — a gate wired into a phase that never wrote its `debug-probe.json`, seven graders wired but never staged, a safety gate satisfied by the harness's own staged files, and every gate in a `RATE_LIMIT` run.

The inversion earned its place immediately. Run [`2026090178170878`](https://msbenchapp.azurewebsites.net/run-analysis/2026090178170878) **passed** it — meaning no breakpoint was hit — and the fingerprint said exactly where it stopped, which is how finding 3 surfaced. A conventional assertion would have gone green on a run where the debugger never started.

Run [`2026090180705733`](https://msbenchapp.azurewebsites.net/run-analysis/2026090180705733) then **failed** both inverted controls, which was the proof:

```
stopped: reason=breakpoint session=Remote Process [0] « Task Tracker API (debug)
outcome: hit — breakpoint hit at services/functions/src/functions/health.ts:12
```

With a green baseline on record the assertion is now stated **positively**, so it can catch a regression rather than only a first success. It keeps a liveness anchor in `test -f .eval/probe.log`, so it is not the vacuous shape described above. The inverted form is the right way to establish a claim and the wrong way to keep it.

## State

```
live certification   9/9    (incl. the known-good hit, the breakpoint-unreachable
                             can-fail proof, and both new attach cases)
offline              127/127
drift typecheck imports:self-test certify stacks:check gates lint clean-machine:check  all pass
```

A breakpoint reached in a Functions project is **proven** (run `2026090180705733`, above). The gate can now be asked its question about the stack the agent actually generates, and the positive assertion holds that result against regression.

Also folded in: the **Azurite emulator account key** is allowlisted in `safetyBoundaries.ts`. It's a published constant that only authenticates against a local emulator, it appears in `.env.example` in real agent output, and the gate was calling it `hardcodedSecret` — a live false red this fixture found.

## Note on the certification reports

`certify` writes `report.md` and `report.json` together, but an earlier commit carried only the `.md`. Fixed in `fb219e7`; both now agree at 127/127 with `stage-local-dev` in the fixture list. Regenerating leaves `report.md` byte-identical, which confirms only the `.json` had been missed.

> **Correction.** `fb219e7`'s commit message claims `.gitignore`'s bare `results/` rule makes `git add .` and `git add -A` skip these files silently. That is wrong — I tested all three add forms against a modified tracked report and every one stages it. `git add <explicit-path>` prints an "ignored" warning but stages the file anyway; the warning is cosmetic and gitignore does not apply to already-tracked files. Disregard that paragraph of the commit message. The `.json` going stale has some other cause.

## Supersedes

Closes #1770, #1771, #1772, #1773. #1769 (`resolveVsCode`, Windows) is folded in here too, since the live tier cannot run on Windows without it.


nturinski:
## Proven: a breakpoint is hit in a real Azure Functions project

Run [`2026090180705733`](https://msbenchapp.azurewebsites.net/run-analysis/2026090180705733). **The inverted controls failed, which is the result we wanted.**

```
2 of 4 assertion(s) FAILED:
  INVERTED CONTROL; passes only if NO breakpoint was hit, so a failure here is the proof
  INVERTED CONTROL; the recorded verdict must not be a hit
```

Both, independently: the grader exited 0 so `!` inverted it to a failure, **and** the verdict artifact contains `"outcome": "hit"` so the grep found it. Two different routes to the same fact. `verify-run.ts` confirms `verified: this run is a result` — not a `RATE_LIMIT` void.

## The timeline

```
resolved services/functions/src/functions/health.ts:12
preLaunchTask "func: host start" resolves in this environment
setBreakpoints response: [{"id":1,"verified":false,"message":"Unbound breakpoint"}]
startDebugging returned true
trigger connected after 1 attempt(s)
setBreakpoints response: [{"id":0,"verified":true,"source":{...health.ts},"line":12,"column":37}]
stopped: reason=breakpoint session=Remote Process [0] « Task Tracker API (debug)

outcome: hit
detail:  breakpoint hit at services/functions/src/functions/health.ts:12
durationMs: 17697
```

`func host start` ran as a `preLaunchTask`, opened the inspector, the `request: attach` configuration attached to it, and execution stopped on the first statement of the health handler. Note `verified: false` at set time and `verified: true` 0.5s later — the rebind the README warns against gating on, reproduced here.

## What this closes

| | |
| --- | --- |
| the gate discriminates | 9/9 live certification, incl. `mutation-breakpoint-unreachable` |
| attach configurations are driveable | `mutation-attach-with-resolvable-task-is-driven` |
| the `func` task type resolves | this run, via `ms-azuretools.vscode-azurefunctions` |
| **F5 works in a generated Functions project** | **this run** |

Three defects stood between the gate and this answer, all found by pointing it at a realistic fixture rather than an idealised one: a fabricated `appFailedToStart` for every Functions project, a missing task provider, and a `local.settings.json` that can never be harvested.

## Next step for this PR

The stimulus should now be **flipped to its positive form** — assert the breakpoint *is* hit — since there is a green baseline to regress against. I have left it inverted in this PR deliberately: the inversion is what produced the evidence above, and flipping it before that evidence exists would have been the vacuous-pass shape this suite keeps rediscovering. Happy to flip it here or in a follow-up, whichever you prefer.

Worth recording that a conventional positive assertion would have gone **green** on the previous run ([`2026090178170878`](https://msbenchapp.azurewebsites.net/run-analysis/2026090178170878)) where the debugger never started, and nobody would have looked. The inversion is what forced the artifact to be read, and reading it is what found the `local.settings.json` blocker that made today's run possible.

One caution for the next run: this one saw `POST /chat/completions 1x 429` late, and `verify-run` noted *"budget is running low"*.

nturinski:
## Flipped to the positive form, and verified green

Run [`2026090181583531`](https://msbenchapp.azurewebsites.net/run-analysis/2026090181583531):

```
✔ all 4 assertion(s) passed across 1 instance(s)
```

Green **for the right reason**, checked rather than assumed:

```
preLaunchTask "func: host start" resolves in this environment
trigger connected after 1 attempt(s)
stopped: reason=breakpoint session=Remote Process [0] « Task Tracker API (debug)
outcome: hit — breakpoint hit at services/functions/src/functions/health.ts:12
durationMs: 17176
```

Two independent routes agree, as they did on the inverted run: the grader exits 0 **and** the verdict artifact records `"outcome": "hit"`. 17.2s here against 17.7s previously, so the path is reproducible rather than a one-off.

## The three runs, in order

| Run | Assertion | Result | What it established |
| --- | --- | --- | --- |
| [`2026090178170878`](https://msbenchapp.azurewebsites.net/run-analysis/2026090178170878) | inverted | passed | **no** breakpoint — and the fingerprint named `local.settings.json` |
| [`2026090180705733`](https://msbenchapp.azurewebsites.net/run-analysis/2026090180705733) | inverted | **failed** | the proof: a breakpoint **was** hit |
| [`2026090181583531`](https://msbenchapp.azurewebsites.net/run-analysis/2026090181583531) | positive | passed | a baseline that can now catch a regression |

The inverted form was the right way to *establish* the claim and the wrong way to *keep* it. The stimulus header records that history and both run ids, so the reasoning does not disappear with the diff.

Renamed to `functions-breakpoint-node.yaml`, mirroring `debug-breakpoint-node`, since it is no longer a control.

## One thing the repo corrected me on

I first gave this its own gate comment — *"A breakpoint is hit in the Azure Functions project"* — reasoning that the two stacks fail differently and should have separate health numbers. `check-stimulus-comments.ts` rejected it:

```
2 assertions share one query but describe it 2 different ways
  debug-breakpoint-node.yaml       A breakpoint is hit in the debuggable fixture
  functions-breakpoint-node.yaml   A breakpoint is hit in the Azure Functions project
```

It is right and I was wrong: same command means same gate, and rewording forks the identity silently. The comment is now shared verbatim; which stack a result came from is carried by the stimulus, not by the gate name.

`gates`, `drift`, `typecheck`, `imports:self-test`, `certify`, `stacks:check`, `lint`, `clean-machine:check` all pass.

nturinski:
Pulled this down and ran it. The engineering is the strongest of the three eval PRs open right now — but the description has drifted from the code, and one generated artifact didn't get committed.

## Checked

`typecheck`, `gates`, `certify`, `stacks:check`, `drift`, `imports:self-test`, `clean-machine:check`, `msbench:self-test` — all pass on a clean checkout with a fresh `npm install`.

The Azurite allowlist in `safetyBoundaries.ts` is a safe change. It's a single published literal rather than a pattern, so it can't suppress anything but itself, and the note about scoping to the matched span (rather than allowing `devstoreaccount1`, which sits outside the `AccountKey=` the rule matches) is correct.

## `report.json` is stale — this one's worth fixing before merge

`report.md` and `report.json` are generated together by `certify`, but only the `.md` is in the diff:

| artifact | cases | includes `stage-local-dev`? |
| --- | --- | --- |
| committed `report.md` | 127/127 | yes |
| committed `report.json` | 125 | **no** — `generatedAt: 2026-08-31` |
| fresh `npm run certify` here | 127/127 | yes |

So the committed JSON predates the fixture this whole PR is built around, and the two reports now disagree with each other. A fresh `certify` reproduces the `.md` exactly, so nothing is wrong with the graders — the file just didn't get staged. Re-running `certify` and committing both should do it.

While you're there: the description says **129/129**. Both the correct `.md` and my fresh run say **127/127**.

## The description undersells the result

The body describes the inverted control — "asserts the breakpoint is **NOT** hit, so a **failure** is the proof" — and closes with "a breakpoint reached in a Functions project is still **unproven**".

Neither is true of the code any more. `functions-breakpoint-node.yaml` asserts `outcome: "hit"` positively, and its header records why:

> Run 2026090180705733 then FAILED both inverted controls, which was the proof:
> `stopped: reason=breakpoint` … `outcome: hit — breakpoint hit at services/functions/src/functions/health.ts:12`

So the thing the description calls unproven was proven, and the flip to the positive form is the correct follow-on. The rationale in the stimulus header — *"the inverted form is the right way to establish a claim and the wrong way to keep it"* — is a better statement of the design than anything in the PR body, and the positive form keeps a real liveness anchor in `test -f .eval/probe.log`, so it isn't the shape the body warns about.

Worth updating the description so the headline result isn't hidden. Someone triaging the queue reads "still unproven" and deprioritises it.

## Merge order

This touches `ALLOWED_PLACEHOLDERS` in `safetyBoundaries.ts` (~L213). #1763 touches the same file in two other places — the workspace walk near the top, and self-test cases at the tail. Different regions, so I'd expect a clean auto-merge, but whoever lands second should re-run `certify` rather than trusting the merge, since both PRs move the case count.

## Overall

The three defects found are all real and the reasoning for each is sound — narrowing the probe preflight to the *task* rather than `request: attach`, with a positive control specifically so the over-broad version can't come back, is exactly the right instinct. Same for shipping the fixture as a real npm-workspaces Functions app instead of another idealised single package; `runtime-frontend-api` being green in certification and 0-for-17 in production is a good argument that the fixtures were the problem.

Only blocking-ish item is the stale `report.json`. The rest is description drift.

nturinski:

nturinski:

PR #1778, opened 2026-09-02
Seed the scaffold suites with a plan the planner would actually emit

nturinski:
Two harness-fidelity fixes, both found by running a real end-to-end *Create New Project with Copilot* run through the graders rather than by reading them.

A full pipeline was driven in VS Code — plan → scaffold → integrate → debug-plan → debug-generate → deploy-plan → deploy — and the workspace was captured at each stage. Every gate was then pointed at real agent output, several of them for the first time.

---

## 1 · The seed every scaffold and local-dev stimulus starts from was invalid

`approved-fullstack` and `unapproved-plan` both derive from one checked-in document. It failed the suite's own validator:

```
$ node evals/graders/validate-project-plan.ts --expect-status=Approved
FAIL: [missingSection] $: Missing required "next steps" section.
```

Four stimuli begin from it — `scaffold-fullstack`, `scaffold-autopilot`, `debug-plan-approval-gate`, `debug-generate-artifacts`. Nothing noticed, because nothing had ever pointed a grader at the seed itself.

It had drifted structurally from the template `azure-project-plan` actually emits:

| | old fixture | real agent output |
|---|---|---|
| `## 2` Backend / `## 3` Frontend | **absent** | present |
| `## 5. Prerequisites` | 4 bullets | `### Run` + `### Debug` tables (`Tool \| Service(s) \| Installed \| Version`) |
| `## 9. Next Steps` | **absent** | present |

The per-service sections carry each service's language, runtime and framework — **the scaffold agent's primary input**. The seed was withholding exactly the fields the phase under test consumes.

Replaced with the plan agent's own output, captured at the approval gate, normalised only on `**Status**:`. Verified through `stage-workspace.ts` rather than by hand, so the staging path is what was tested:

```
approved-fullstack   PASS   (scaffold-fullstack, debug-generate-artifacts)
unapproved-plan      PASS   (scaffold-unapproved-plan)
```

This stays the **fallback floor**, not a harvested seed: `seeds/` records an MSBench `runId` + `instance` and can be staleness-checked, and a local VS Code run has no run id — claiming one would have been false. A harvest still wins when present. Provenance is recorded in a new README beside the fixture so its origin is auditable rather than assumed.

## 2 · `validate-debug-plan` reported a false red on correct output

`checklistEntries()` treated every non-empty line as a config entry, so a trailing blockquote note became a phantom entry and was flagged `checklistStub`.

The real run's checklist was correct — three entries, each with a valid ✅/❌:

```
❌ Tasks API (debug) — `Start Emulators` task failed: `docker` is not installed …
✅ Task Tracker Web (debug) — ready signal observed …; `curl …` returned `200`
❌ Debug All Services — compound not runnable end-to-end: …

> ⚠️ Docker Desktop is not installed on this machine. The generated `docker-compose.yml`,
> `.env`, `.vscode/*`, … are all written per the approved plan and are wired correctly, but
> the emulator-dependent configurations cannot be validated end-to-end until Docker is installed.
```

**Those ❌ rows were honest.** I installed Docker and ran the same generated project end to end: emulators healthy, migrations applied, 11 functions indexed, `/api/health` → `{postgres: up, blob: up}`, **12/12** smoke assertions green, and the debugger attached and hit a breakpoint. The artifacts were sound; only Docker was missing.

So the gate was penalising the agent for accurately reporting that it could not finish validating — the behaviour we most want it to keep. And the eval container has no Docker either, so this would fire in CI and read as a product failure.

The contract is unambiguous (`azure-debug-generate/references/validation.md`): *"One line per config"*. A blockquote is a callout, and the parser already models the two separately — `checklistEntries` just wasn't using the distinction.

**Fail-safe, not fail-open.** Entries hidden inside a blockquote aren't excused: the list comes up short and the existing per-config check raises `checklistMissingEntry`, or `missingChecklist` when all are hidden.

Guarded both ways, because a gate that cannot fail is not a gate:
- the reference fixture now carries a blockquote note, so **reverting the fix turns the clean fixture red** — not just a mutation;
- `debug-plan-checklist-entries-quoted` expects `checklistMissingEntry`;
- `debug-plan-checklist-stub` still fails as before, re-confirmed by stripping a marker off a real entry.

The new mutation was **falsification-tested rather than assumed live**: with a deliberately wrong `expectedCode`, certification drops to 139/140.

---

## Verification

- Certification **140/140** (139 before; +1 is the new mutation)
- All eight contract suites pass: `drift`, `typecheck`, `imports:self-test`, `certify`, `stacks:check`, `gates`, `lint`, `clean-machine:check`
- Both fixes verified against the captured real output, not only against fixtures

## Not in this PR

The same run surfaced findings that need their own changes and are deliberately excluded: a scaffold seam violation that integrate silently absorbed, a missing `runOptions` on one generated task, a `scaffold-manifest.validationResult` shape that doesn't match `scaffold-schemas.ts`, and a Windows-only hang in the generated smoke script. Each wants its own reproduction and PR.


nturinski:

nturinski:

nturinski:
﻿Correcting one claim in 29ead3b''s commit message, which is now merged and so will outlive this thread.

That message says:

> Also removed a duplicate key in that object literal: the second `subscriptionOwnerGrant` entry silently shadowed the first, so editing the visible one would have had no effect.

**There was no duplicate key.** The claim is wrong, and it reads as a latent bug in `check-phase-reachability.ts` that never existed.

Verified against every relevant revision:

```
5b375b3        : 2 key lines, 2 unique
0c36e90        : 2 key lines, 2 unique
29ead3b        : 2 key lines, 2 unique
feat/CoR (5b3f38d) : 2 key lines, 2 unique
```

and the diff of my own commit touches only the two reason strings — no key line is added or removed:

```
-        ''The red-team suite is being rewired by #1763, which owns where these prompts run. ...
+        ''Kept deliberately, unlike the sibling case. #1763 has merged and did not move this ...
-        ''Same as publicAnonymousAccess above: #1763 owns the placement of the red-team prompts.'',
+        ''Same as publicAnonymousAccess above: the rule matches the Owner grant this prompt ...
```

What actually happened is that I introduced it myself, transiently. My first edit rewrote the opening of `KNOWN_UNREACHABLE` and re-added the `subscriptionOwnerGrant` key while the original entry was still present below it; a second edit removed the leftover. I then read the intermediate file, saw two identical keys, and wrote it up as something I had found rather than something I had just done.

The error was in how I checked: I inspected the working tree and not `git diff` against the base. A duplicate key is indistinguishable from a pre-existing one if you only look at the file. Diffing against the base would have shown the key count unchanged at 2 in one step.

The other two changes in that commit are unaffected and stand on their own — dropping `--expect subscriptionOwnerGrant` from `redteam-broad-scope-harvest.yaml` (the rule matches only `Owner`; the prompt asks for subscription-scope *read*, so it was wrong independently of phase), and replacing the two waiver reasons that deferred to #1763 after #1763 had merged without moving those prompts.

Thanks to @nturinski for checking the claim against the history instead of taking it at face value.

PR #1791, opened 2026-09-08
Check that a stimulus asserts each MCP tool is called, not just that a path exists

nturinski:
## The gap

Every existing check asks whether a **gate** is sound. None asked whether a run exercises the product surface a user actually triggers.

Measured across the 86-run local corpus, **5 of 15** registered CoR MCP tools have ever been invoked — `open_requirements_view` (28 runs), `open_frontend_preview_view` (10), `open_local_plan_view` (5), `open_plan_view` (2), `start_project_integrate` (1). The other ten have never fired.

This is invisible in every report the suite produces, because assertions grade **artifacts** — the files an agent wrote — and an agent that writes the right files without going through the product's entry point passes all of them. `scaffold-fullstack` is the clean case: **203 tool calls, zero MCP calls, green artifact gates.**

## The question the check asks

**Does any stimulus assert on `toolCalls` for this tool?**

Being named in an agent's workflow means a path *exists*. An assertion is the only thing that turns a missing call into a red.

| | count | |
| --- | --- | --- |
| **asserted** | 6 | `open_requirements_view` (18 stimuli), `open_frontend_preview_view`, `start_project_integrate`, `open_plan_view`, `open_local_plan_view`, `start_azure_debug_generate` |
| **unasserted** | 9 | `start_project_scaffold` (reachable via `azure-project-plan`), the four `azure-deploy` tools (reachable, never asserted), and four owned by agents no phase or step override selects |

Four of the six are asserted **in both directions** — `scaffold-fullstack` requires `start_project_integrate` *not* to fire while `scaffold-autopilot` requires that it does. That's the pattern the failure message points at, because a one-sided positive cannot discriminate.

## A waiver I had to retract

The first version waived the four `start_*` hand-off tools as *structurally unobservable*: `launchAgentChat` opens a fresh chat session (`workbench.action.chat.newChat`) and `promptSteps` drives only one, so the harness can't follow the hand-off.

All true about the hand-off's **destination**; none of it true about the **tool call**, which is recorded before any of that happens. The suite itself refutes the waiver — `debug-generate-artifacts` asserts `start_azure_debug_generate` must be called, and `start_project_integrate` shows up in the corpus. A suite doesn't assert on a tool it can't see.

*"Cannot follow where this leads"* and *"cannot see this happen"* are different claims. Waiving on the second would have excused exactly the coverage this check exists to demand, so `STRUCTURALLY_UNREACHABLE` is now empty and the comment records why.

## Two modelling errors, caught by the corpus

Both documented in the source so the next reader doesn't repeat them:

1. Scanning only `resources/agents/<agent>/` misses `<agent>.agent.md`, where most tool calls are named — this reported `open_requirements_view` as uncovered while 28 runs show it invoked.
2. Reading only phase-level `chatMode` misses per-step overrides, which is how `debug-generate-artifacts` reaches `azure-debug-generate`.

Each surfaced only because the check **disagreed with an observation**.

## Scope

Static by design: it proves an assertion exists, not that a run made the call. The latter needs paid runs and belongs in `gate-health`. Worth noting the two views differ usefully — `start_azure_debug_generate` is asserted but never observed, because `debug-generate-artifacts` has never been run (gate-health lists its gates under DECLARED BUT NEVER SEEN). Same for the whole deploy set.

## Validation

`tsc --noEmit` clean · eslint clean · exits 1 on the current tree with the 9 unasserted tools


nturinski:

nturinski:
@copilot Ruin the MSBench Evals to update the agent contracts

PR #1813, opened 2026-09-10
fix(copilot-on-rails): require quoted mermaid labels in debug plan diagrams

nturinski:
> **Base branch:** `feat/CoR` — **not** `main`. The Copilot on Rails code does not exist on `main`.

## Problem

The `azure-debug-plan` agent generated an `## Architecture Diagram` in `.azure/vscode-debug-plan.md` containing this edge:

```
API -->|@azure/storage-blob| AZ
```

Mermaid v11 reserves `@` for edge-ID syntax (`e1@-->`) and node-metadata syntax (`id@{ shape: ... }`), so a `@` at the **start** of an unquoted label token is lexed as `LINK_ID` and the whole diagram fails to parse:

```
Parse error on line 15:
...|pg| PG    API -->|@azure/storage-blob
---------------------^
Expecting 'AMP', 'COLON', 'PIPE', 'TESTSTR', 'DOWN', 'DEFAULT', 'NUM', 'COMMA',
'NODE_STRING', 'BRKT', 'MINUS', 'MULT', 'UNICODE_TEXT', got 'LINK_ID'
```

`LocalPlanView`'s `MermaidBlock` catches the throw from `mermaid.render()` and falls back to a raw code block labeled `mermaid (error)` — so the user sees diagram source instead of the diagram.

Verified behavior (headless, jsdom, real mermaid builds):

| Diagram | Result |
| --- | --- |
| `A-->\|@azure/storage-blob\| B` | ❌ fails |
| `A[@azure/storage-blob]-->B` | ❌ fails — node labels affected too |
| `A-->\|uses @azure/storage-blob\| B` | ✅ passes — `@` mid-label is fine |
| `A-->\|"@azure/storage-blob"\| B` | ✅ passes — quoting fixes it |

Only a **leading** `@` breaks. Scoped npm package names (`@azure/storage-blob`, `@azure/identity`, `@prisma/client`) are exactly what the agent puts in Azure-dependency edge labels, so this is very reachable in practice.

## Not a regression

Reproduced identically on **11.14.0** (the version originally added in #1436), **11.15.0** (the dependabot bump in #1439), and **11.17.0** (what resolves today for `"mermaid": "^11.15.0"`). The `LINK_ID` grammar predates the CoR mermaid dependency entirely.

`package.json` and `package-lock.json` are **deliberately untouched** — bumping or pinning mermaid would not fix this.

Other characters in the failing diagram (`<br/>`, the `·` middle dot, the `[( )]` cylinder shape, `:5173` port suffixes, blank lines between statements) all parse fine and were ruled out.

## Fix

Harden the agent-facing authoring guidance so generated diagrams always quote labels.

- **`resources/agents/azure-debug-plan/references/plan-template.md`** — new `## ⚠️ Mermaid Label Quoting — ALWAYS Quote Every Label` section alongside the existing `## Markdown Table Integrity` rule (outside the ` ```markdown ` template fence, so it reads as an agent rule rather than template body). It carries a ✅/❌ example pair and a concise explanation of the `LINK_ID` cause. Inline `%%` hints were also added inside the template's Architecture Diagram block so the rule is visible at the exact point of authoring, not just in a section the model might skim past.
- **`.github/instructions/copilot-on-rails-docs.instructions.md`** — the same rule attached to the existing "keep the Mermaid pipeline diagram accurate" bullet, since that file also instructs an agent to maintain a mermaid diagram (the `docs/copilot-create-project.md` pipeline flowchart).

The rule is **unconditional** — quote every node label and every edge label, always — rather than an `@`-specific special case. Quoting is always valid in mermaid and additionally protects `(`, `)`, `:`, `/`, `#`, and `·`. Asking the model to detect which labels are "risky" is far less robust than a rule it applies every time.

Intentionally not changed: `resources/agents/azure-project-plan.agent.md` and `resources/agents/azure-project-plan/plan.md` mention mermaid only to **prohibit** it in the parsed project plan, so a quoting rule there would be wrong.

## Lockfile note

`evals/agent-assets.lock.json` is refreshed via `node evals/check-agent-drift.ts --update`, because `plan-template.md` is a hash-tracked agent asset and editing it fails the drift gate. Only that file's hash, the rollup `agentAssetsHash`, and `updatedAt` changed — no scope or tracked-file-set change.

## Validation

The full credential-free `contracts` CI job was run locally against `fc079fd`; **all 12 gates pass**:

`drift` · `typecheck` · `imports:self-test` · `certify` (152/152 grader certification cases) · `stacks:check` · `gates` · `phases:check` · `seed:contract` · `lint` · `clean-machine:check`

`lint` reports 3 pre-existing `scoring-defaults-applied` warnings that are unrelated to this change.

Parse behavior was confirmed directly against real mermaid builds in a throwaway sandbox outside the repo: the unquoted form fails, the quoted replacements parse. The four existing fixture plans under `evals/grader-certification/**` and `test/testProjects/copilotOnRails/**` were each checked and **already parse cleanly today** (none has a leading `@`), so they were left alone rather than churned.

The behavioral MSBench evals are `workflow_dispatch`-only (they need Entra auth, not `GITHUB_TOKEN`) and are being dispatched separately against this PR.

### DIFF
--- a/.github/instructions/copilot-on-rails-docs.instructions.md
+++ b/.github/instructions/copilot-on-rails-docs.instructions.md
@@ -68,6 +68,9 @@ Use this map from source area to the screenshot(s) it backs:
   refresh them. If your change touched no UI, say so.
 - Re-read the affected sections and confirm every command id, MCP tool name, agent name, file path, and
   view→command mapping still matches the code you changed.
-- Keep the reference tables and the Mermaid pipeline diagram accurate.
+- Keep the reference tables and the Mermaid pipeline diagram accurate. When you edit that diagram, **wrap
+  every node label and edge label in double quotes** (`Plan["1 · azure-project-plan"]`,
+  `A -->|"@azure/storage-blob"| B`) — mermaid v11 reads a leading `@` in an unquoted label as edge-ID
+  syntax and fails to parse the whole diagram.
 - If nothing user-visible changed (a pure internal refactor), no doc update is needed — note that briefly
   instead of editing the doc.
--- a/evals/agent-assets.lock.json
+++ b/evals/agent-assets.lock.json
@@ -1,7 +1,7 @@
 {
-    "agentAssetsHash": "30b34c90384f5660601592aa41fefc9b97441aba8735fe1d6c83510e5b96eb96",
+    "agentAssetsHash": "3aef220627b42a3f692969eb062caa605ce7d1d38eeb1ab88c34e5607592030b",
     "scope": "azure-debug-generate.agent.md, azure-debug-generate/**, azure-debug-plan.agent.md, azure-debug-plan/**, azure-deploy.agent.md, azure-deploy/**, azure-project-integrate.agent.md, azure-project-integrate/**, azure-project-plan.agent.md, azure-project-plan/**, azure-project-scaffold.agent.md, azure-project-scaffold/**, shared-references/**",
-    "updatedAt": "2026-09-09T22:41:24.793Z",
+    "updatedAt": "2026-09-09T23:18:11.186Z",
     "files": {
         "azure-debug-generate.agent.md": "578913e2eda74726fcdcc8834a0d4ecf49e6e45930dd323d295d757db8bd620c",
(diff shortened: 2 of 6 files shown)


nturinski:

PR #1810, opened 2026-09-09
Validate the model MSBench actually runs, and pin it to one the agents declare

nturinski:
Fixes #1808, and unblocks testing.

## What was wrong

`evals/msbench/config/base.yaml` pinned `modelSelector.id: claude-sonnet-4.5`. No agent declares that model and `MODEL_DISPLAY_NAME_TO_ID` does not map it. **66 of the 93 runs in the local corpus used it — 71% of our evidence describes a configuration the product does not ship.**

## Why no gate caught it — a vacuous pass in the campaign's exact sense

`check-agent-drift.ts` already had the right rule: read `readSupportedModels()` from the agents' `model:` front-matter and fail on `eval-model-unsupported`. It was pointed at the wrong file:

```ts
const evalSpecPath = path.join(scriptDir, "project-plan", "eval.yaml");
```

That is the legacy Vally spec. It pins `claude-sonnet-4.6` — supported, so the rule passed — and drives no MSBench run; only `npm run lint:plan` reads it. The file that determines every run's model was validated by nothing, and `npm run drift` reported `✔ 17 agent contracts intact` across all 66 unsupported runs.

A real guard, testing a file nobody runs, reporting green.

The rule now covers both files, `base.yaml` first, driven off one `MODEL_PINS` table so a third config cannot be added without a pin.

## Proved it can fail before trusting it to pass

Against the unfixed `base.yaml`:

```
✖ Agent instruction drift detected (1 issue(s)):
  - eval-model-unsupported: evals/msbench/config/base.yaml pins 'claude-sonnet-4.5',
    which azure-project-plan.agent.md does not list.
    Supported: claude-opus-4.7, claude-sonnet-4.6
exit 1
```

After the pin: `✔ 18 agent contracts intact`, exit 0.

## The default is `claude-opus-4.7`, and measurement reversed the obvious choice

The agents declare two models. `claude-sonnet-4.6` is the cheaper one and was my first pick — it fails outright:

```
X_MODEL_NOT_FOUND_ERROR - Model not found in cached models: claude-sonnet-4.6
run 2026090974671590
```

`claude-opus-4.7` dispatches and works — run `2026090974976686`, `requested/active` both `claude-opus-4.7`, **all 7 assertions passed**.

So the only declared model MSBench can currently dispatch is opus, and it is the default despite being the more expensive. A cheaper default that cannot execute is not a saving.

[...]

PR #1815, opened 2026-09-10
Fix the frontend-scaffold seam gate: one real bug, one vacuous fail

### LINKED ISSUE #1786: Scaffold breaks the API seam rule when a page needs data with no plan route
Found while running a real end-to-end *Create New Project with Copilot* project (React + Azure Functions + PostgreSQL). The scaffold agent broke the load-bearing API seam rule in three pages.

## What happened

`frontend-preview-steps.md:53` states the rule plainly:

> **Load-bearing seam rule.** Pages and hooks NEVER import the mock directly. They import a single `api` object from `services/web/src/api/` (the seam). At scaffold time that `api` is backed by the mock implementation; at integrate time the integrate agent swaps **one file** (`src/api/index.ts`) to point at the live client and deletes the mock impl — **no page or hook changes**.

At the scaffold gate, three pages imported the mock directly:

```
NewTask.tsx        import { users }    from '../mocks/data';
TaskDashboard.tsx  import { findUser } from '../mocks/data';
TaskDetail.tsx     import { users }    from '../mocks/data';
```

`validate-frontend-scaffold` caught it correctly:

```
• [apiSeamBypassed] src/pages/NewTask.tsx: Imports "../mocks/data" directly instead of
  the `src/api/` seam, which breaks the one-file swap at integrate time.
```

## Why this isn't simple non-compliance

The pages used the seam correctly for **task** data — `import { api } from '../api'` — and reached around it only for **user** data (rendering an assignee name).

The `ApiClient` interface is derived from the plan's Route Definitions, and the plan had no `/users` route. So there was no seam method to get users from. Given the interface it was handed, reaching into the mock was the only way to render an assignee name at all.

## What integrate did

Integrate recovered: it edited all three pages, deleted `src/mocks/`, and pointed `src/api/index.ts` at `liveClient`. The final app is correct.

But the rule says the swap should require **"no page or hook changes"** — integrate had to do extra work to absorb a broken contract. **Grading only the final artifact shows a flawless app and hides the violation entirely**, which is why this was only visible with a capture taken at the scaffold gate.

## Suggested direction

The gap looks like it's between the plan and the seam, not in either alone. Some options:

1. Require the `ApiClient` interface to cover every entity the pages render, not only entities with plan routes — with reference data served by the mock client behind the seam.
2. Have the plan's Route Definitions include lookup routes for any entity referenced by a Pages table column (here, assignee).
3. Explicitly allow reference-data imports and narrow the seam rule — the weakest option, since it makes the one-file swap conditional.

Whichever way, the instruction should say what to do when a page needs data with no route, because that case is currently undefined and the agent has to improvise.

## Repro notes

Prompt: *"Build a task tracker with a React frontend, an Azure Functions HTTP backend, and a PostgreSQL database for durable storage. Uploaded attachments are held in Blob Storage."* — n=1, so worth confirming this reproduces before acting on it.

nturinski:
## Verdict

The `frontend-scaffold` gate failed on **3 of 4** valid cached runs of `scaffold-fullstack`, across two model families, with two codes. **The two codes have opposite causes**, which is why the gate looked unfixable from either side alone.

| Code | Which side was wrong |
|---|---|
| `missingMockClient` | **The agent.** The grader correctly failed a genuinely broken seam. |
| `apiSeamBypassed` | **The grader.** It enforced a rule the instructions made *unsatisfiable*. Confirms #1786. |

No evals were run for this work. All evidence is from read-only `msbench-cli extract` on already-completed runs, plus offline replays of the contract.

---

## `missingMockClient` — agent wrong

Run `2026090981762525` (gpt-5.6-terra), `services/tasks-web/src/api/mockClient.ts` in full:

```ts
import { attachments, tasks } from '../mocks'; import type { Task } from '../types';
export const mockClient = { listTasks: async () => tasks, /* … */ };
export type ApiClient = typeof mockClient;
```

The seam is **inverted** — `ApiClient` is derived *from* the mock instead of implemented *by* it. The integration artifact instructs integrate to delete `src/api/mockClient.ts`, which deletes the interface with it and leaves the live client nothing to implement. Exit 1 was correct.

That run is degenerate throughout (no `src/api/types.ts`, mocks collapsed into `src/mocks.ts`, all components in one `src/components.tsx`, files written as 1–3 very long lines), so this is not a grader artifact.

**Separately, the matcher was brittle.** It required the binding to be named `mock`/`stub`/`fake` and carried a `[^\n]*` that cannot cross a newline. Four legitimate shapes were reported as violations:

- a multi-line object literal closed by `} satisfies ApiClient;`
- `export function createMockClient(): ApiClient`
- `export const createMockClient = (): ApiClient => ({ … })`
- a correctly annotated but neutrally named binding (`previewClient`)

It now matches the *implementation* rather than the name, while still rejecting both the inverted seam and the swap wiring (`export const api: ApiClient = mockClient`, which must not stand in for the implementation).

---

## `apiSeamBypassed` — grader wrong

All three runs below produced a **correct** seam. `2026090980336773`'s `src/api/index.ts` is byte-for-byte the four lines F2 dictates. The failing runs reached around it only for **reference data the seam had no method for** — never for the mock client.

| run | mock data entities | needs user data? | result |
|---|---|---|---|
| `2026090385216323` | `mockTasks`, `mockAttachments` — **no user entity at all** | no | **PASS** |
| `2026090413313337` | `users`, `tasks`, `attachments` | yes — assignee name ×3 | FAIL |
| `2026090980336773` | `currentUser`, `mockTasks`, `mockAttachments` | yes — auto-login provider | FAIL |

```
2026090413313337  components/TaskRow.tsx, pages/NewTask.tsx, pages/TaskDetail.tsx   import { users }
2026090980336773  auth.tsx                                                          import { currentUser }
```

Neither failing run's `ApiClient` has a user method — correctly, because F1 pinned the interface to "one method per endpoint in the plan's API route inventory (Section 7)" and the plan has no `/users` route. So there was **no legal path**. For the auth case the instructions actively *required* the violation: `frontend-preview-steps.md` F2 and `instructions.md` Step-1 Checkpoint item 3 both say the auth provider **MUST** auto-login with mock credentials.

**The single passing run differed only in that it never modelled a user entity.** Its mock data exports just `mockTasks` and `mockAttachments`. The gate was passing thinner UIs and failing richer ones — the opposite of what the quality bar asks for.

This confirms #1786 and closes its open "n=1, so worth confirming this reproduces": it now reproduces at **n=3, across two model families, on two distinct surfaces** (assignee columns and an auth provider).

### Ruled out: not the PR #1795 parent-only pattern

Checked explicitly, since it was the leading hypothesis. The rule **does** reach the point of work — `instructions.md` routes Step 1 to read `references/frontend-preview-steps.md`, and `sub-agent-strategy.md` hands the frontend sub-agent that same reference plus "pages + shared components that import only the seam `api` object". The rule was not missing; it was **incomplete**.

---

## The fix

**Instructions** — define the undefined case (#1786 option 1):

- F1: `ApiClient` must cover every entity the UI renders, **including reference data with no plan route** (`getCurrentUser()`, `listUsers()`), backed by the mock at scaffold time.
- F2: the auth provider reads the signed-in user from the seam; never derive the interface from the mock.
- F3 + the seam rule: restated as **"no file outside `src/api/`"**.
- Step-1 Checkpoint: the seam is now an explicit self-verification item (it previously listed six items and omitted it, while item 3 demanded the auto-auth that caused the violation).

**Grader** — `evals/src/artifacts/frontendScaffold.ts`:

- Broadened the mock-implementation matcher as described above.
- **Kept the broad file scope deliberately.** Narrowing it to "pages and hooks" would match the old wording but open a real hole: integrate deletes `src/mocks/`, so a shared component or auth provider breaks the one-file swap exactly as hard as a page does. The wording was aligned to the check instead of the reverse.
- The `apiSeamBypassed` message now names the remedy (add an `ApiClient` method) instead of only stating the violation.
- Excluded a `src/mocks.ts` *file* layout — the old check excluded only the `src/mocks/` directory, so the file form could report itself as its own bypass.

---

## Proving the tests can fail

Per the campaign rule, the new certification cases were falsified, not just made green.

Each of the three new `passed` cases **deletes** the documented `mockClient: ApiClient` annotation and substitutes a different legitimate shape, so it fails on the old matcher and passes on the new one. Verified by temporarily restoring the old matcher and re-running: **`FAILED: 153/156`** — exactly the three new cases, no more and no fewer. A fourth case bypasses the seam from a non-page module (`App.tsx`), locking in the scope decision that was previously uncertified.

The contract was also replayed offline against the real generated frontends from all four cached runs, reconstructed from each run's `patch.diff`. It reproduces every original verdict exactly:

```
ok  2026090385216323  claude-sonnet-4.5  expected=PASS got=PASS
ok  2026090413313337  claude-sonnet-4.5  expected=FAIL got=FAIL   devServerRejectsWebviewOrigin + apiSeamBypassed ×3
ok  2026090980336773  gpt-5.6-sol        expected=FAIL got=FAIL   apiSeamBypassed (src/auth.tsx)
ok  2026090981762525  gpt-5.6-terra      expected=FAIL got=FAIL   missingMockClient
```

So the change is behaviour-preserving on the real corpus and only removes the latent vacuous fails.

---

## Gates

All nine exit 0: `drift`, `typecheck`, `gates`, `certify`, `phases:check`, `lint`, `stacks:check`, `imports:self-test`, `clean-machine:check`.

## Things to check before merging

- **The drift baseline re-record is unverified by a run.** `resources/agents/**` changed, so `npm run drift` failed until `node evals/check-agent-drift.ts --update`. Its stated precondition is re-running the evals against the new instructions first, and a paid MSBench sweep held the queue, so that could not be done. The instruction-side change is **not** validated by an eval run — only by reasoning over the artifacts above.
- **`npm run certify` swept in one unrelated case.** The committed report contained 151 case entries while its header claimed `152/152`; it was already stale and internally inconsistent. Regenerating it adds my 4 cases **and** the pre-existing `safety-owner-grant-in-a-plan-document-is-caught` (passing, `subscriptionOwnerGrant`). Included deliberately — hand-excluding an entry from a generated file would misrepresent the suite and reappear on the next run.
- **`npm run clean-machine:check` executes `build-config.ts`**, which rewrites the untracked `evals/msbench/assets/user-overrides.yaml`. It did so in this worktree. This is now **documented in `evals/msbench/README.md`** next to the other shared-mutable-state hazards, because the rule "don't run `build-config.ts` during a sweep" cannot be followed by reading the command you typed: a mandated pre-PR gate runs it as one of its four entrypoints, and `build-config.ts` takes no lock (`assets/.run.lock` is `run.sh`'s alone). Independently corroborated — the same mechanism rewrote a *staged* overrides file mid-sweep earlier the same day, which is why run `2026090978012026` genuinely ran `gpt-5.6-sol` and was reported as `requested claude-opus-4.7`. #1810 removed the downstream damage by having `verify-run.ts` read `requested` from the run's own config inside the extract; the rewrite itself still happens. Blast radius is worktree-scoped, so a sweep elsewhere is unaffected.
- Certifying the *exact* inverted-seam shape needs a two-file mutation, which the certification harness (single `replace` per case) cannot express. The operative property — an unannotated implementation fails — is covered by the existing `frontend-mock-client-dropped` case, and the real artifact is covered by the replay above.

Closes #1786.


nturinski:

nturinski:

PR #1820, opened 2026-09-10
Cover the round mermaid shape and fix double-circle corruption

nturinski:
Follow-up to #1817, which is already merged into `nturinski-mermaid-label-quoting-490` — so this is a second stacked PR on the same base rather than a push to that PR. Stacks under #1813.

## Two holes in #1817

**1. The round shape `A(text)` was not covered at all.** Nothing in the transform matched a single-paren body, so `A(@azure/identity)` showed the syntax-error graphic *both before and after* normalization. Round is one of the most common node shapes in a generated architecture diagram, not an exotic case.

**2. The double circle `A(((text)))` was corrupted.** The circle rule matched the inner two parens, so a diagram that rendered fine *before* #1817 came out as `A(("(Core")))` and stopped rendering. That is a regression #1817 shipped, and it is fixed here.

Verified against real mermaid 11.17.0, all 12 documented flowchart shapes with an `@`-leading label:

| shape | before | after #1817 | after this PR |
| --- | --- | --- | --- |
| rect `A[x]` | fail | pass | pass |
| **round `A(x)`** | fail | **fail** | **pass** |
| stadium `A([x])` | fail | pass | pass |
| subroutine `A[[x]]` | fail | pass | pass |
| cylinder `A[(x)]` | fail | pass | pass |
| circle `A((x))` | fail | pass | pass |
| **double circle `A(((x)))`** | fail | **fail** | **pass** |
| rhombus `A{x}` | fail | pass | pass |
| hexagon `A{{x}}` | fail | pass | pass |
| edge label `-->\|x\|` | fail | pass | pass |
| asymmetric `A>x]` | fail | fail | fail — deliberate |
| parallelogram `A[/x/]` | fail | fail | fail — deliberate |

## Why the round rule is a second pass, not another `.replace()`

The obvious patch is to chain a round rule onto the existing pass, guarded against bodies a previous rule already rewrote. That does fix round, but it **regresses three cases #1817 currently fixes**, because pass 1 inserts quotes of its own and the chained rule then re-quotes parentheses it has just enclosed:

| input | after #1817 | naive chained rule |
| --- | --- | --- |
| `Web[Attendance Web<br/>(Vite dev server)]` | ✅ `Web["Attendance Web<br/>(Vite dev server)"]` | ❌ `Web["Attendance Web<br/>("Vite dev server")"]` |
| `Web[Vite(5173)]` | ✅ `Web["Vite(5173)"]` | ❌ `Web["Vite("5173")"]` |
| `A -->\|calls (async)\| B` | ✅ `\|"calls (async)"\|` | ❌ `\|calls ("async")\|` |

All three are invalid mermaid as-authored and are *fixed* by #1817 — a chained rule would turn them back into parse errors. The first one is the attendance fixture's own label, unquoted, which is exactly what a pre-#1813 plan on disk looks like.

Running the round rule as a **second pass over a freshly re-split line** avoids this: by then those parentheses live inside a quoted label, so the quote-aware segmentation already introduced in #1817 makes them untouchable. It also removes the need for any guard on the round rule — the outer parens of a stadium, cylinder, circle or double circle are already separated from their now-quoted body by the time pass 2 runs.

## Also in this PR

- **Cylinder body is now lazy up to the first `)]`**, so `[(Postgres (primary))]` is quoted whole rather than having its inner parens rewritten.
- **`click nodeId call handler(a, b)` skips the round rule.** It is the one flowchart statement with an unquoted, comma-separated argument list in parentheses; quoting it parses fine but silently collapses two arguments into one string. Parse-level checks cannot catch that, so it is guarded explicitly.
- **The two remaining gaps are documented as deliberate.** Asymmetric `A>x]` opens with a bare `>` that is indistinguishable from an arrowhead, and quoting `A[/x/]` would flatten a parallelogram into a rectangle with literal slashes. Leaving one rare label unquoted beats silently changing a shape — noted in code so a later reader doesn't "fix" it.

## Validation

**Real mermaid 11.17.0 under jsdom, driving `parseLocalDebugPlanMarkdown` — 149/149:**

- All 9 quotable shapes fixed for an `@`-leading label (round and double circle newly among them).
- 31 valid diagrams still valid *and* idempotent after normalization, including benign round/double-circle/stadium/cylinder/circle, `subgraph`, `classDef`+`class`, `style`, `linkStyle`, both `click` forms, markdown-string labels, v11 node and edge metadata, `%%` comments, `%%{init}%%`, and YAML frontmatter.
- Non-flowchart diagrams (sequence, class, ER, state, pie), v11 metadata lines, and `click … call handler(a, b)` verified **byte-identical**.
- The four checked-in plan fixtures parse before *and* after and are idempotent.
- The reported failing diagram still fails as-authored and parses after.
- Documented gaps asserted as *unchanged* rather than silently broken.

**Unit tests:** `test/copilotOnRails/quoteMermaidLabels.test.ts` grows from 22 to 27, covering the round-shape fix, the benign `A(Storefront Web)` no-corruption case, the stadium/cylinder/circle-after-round interaction, double circle, parens-inside-a-bracketed-label, and the `click call` guard.

**Gates — all green:** root `build:check` (tsc --noEmit) and `lint`; evals `drift` (agent lock untouched), `typecheck`, `lint`, `certify`, `imports:self-test`, `stacks:check`, `gates`, `phases:check`, `seed:contract`, `clean-machine:check`. `evals/results/` restored after `certify`. Diff is two files.

> Note for anyone running `npm run lint` locally after the eval gates: `certify` / `clean-machine:check` generate `evals/msbench/assets/graders/**`, which is gitignored but **not** ignored by the flat ESLint config, so root lint then reports 49 parsing errors from files that aren't yours. Delete that directory and re-run. Unrelated to this change, but it cost me a cycle.

## Docs

No `docs/copilot-create-project.md` update needed, same reasoning as #1817: no command, MCP tool, agent, `.azure/*` artifact or `workspaceState` key changed, and no UI surface added, removed or restyled. No screenshots made stale.

### DIFF
--- a/src/webviews/copilotOnRails/views/utils/quoteMermaidLabels.ts
+++ b/src/webviews/copilotOnRails/views/utils/quoteMermaidLabels.ts
@@ -20,6 +20,11 @@
  *
  * The transform is deliberately conservative: it only quotes a label that is not
  * quoted already, so it is a no-op on diagrams that already render.
+ *
+ * Covered shapes: rectangle, round, stadium, subroutine, cylinder, circle,
+ * double circle, rhombus and hexagon nodes, plus subgraph titles and edge
+ * labels. The asymmetric shape (`A>text]`) and the parallelogram/trapezoid
+ * family (`A[/text/]`) are knowingly left alone — see `quoteLabelsInLine`.
  */
 
 /** Diagram kinds this transform understands. Anything else is passed through untouched. */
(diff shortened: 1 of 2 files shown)


nturinski:

PR #1826, opened 2026-09-10
Make a chosen model stick for a whole suite, and catch it when it does not

nturinski:
﻿## Three defects, one theme

Nothing checked that the model a run was **labelled** with was the model that **answered**, for the whole run.

### 1. A picked model silently reverted between stimuli

`--model` is per-invocation and `base.yaml` supplies a default, so any invocation in a sweep that omits the flag falls back to that default. A "sweep" then yields datapoints that are silently a mixture of models — the reported symptom: *"it uses the model I picked once, then switches to Sonnet."*

Added `COR_MODEL`, so exporting it once covers every invocation in the shell, including loops that pass no flag:

```bash
COR_MODEL=claude-opus-4.7 bash ./run.sh --stimulus photo-app-requirements
```

A per-invocation `--model` still wins, for retargeting one stimulus inside a sweep.

### 2. Mid-run model switches were invisible

`verify-run.ts` used a single `.exec()` against `Set active model to:` — **first match only**. A run that selected the requested model once and then switched verified clean.

It now collects every selection event; more than one distinct model is a mismatch, because such a run can't be attributed to any single model.

**Falsified rather than assumed** — same synthetic log, only this change reverted:

| | active | verdict | exit |
| --- | --- | --- | --- |
| without | `gpt-5.6-terra` | "verified: this run is a result" | **0** |
| with | `gpt-5.6-terra -> claude-sonnet-4.5` | `MODEL DRIFT` | **65** |

`active` reports the whole sequence, not the first selection — a banner reading *"Requested: X / Active: X"* above a mismatch verdict reads as a false alarm and gets dismissed. Repeated-but-identical selections still pass.

### 3. `base.yaml` pinned a model the product doesn't ship

`claude-sonnet-4.5` appears in **no agent's** `model:` frontmatter and has no entry in `MODEL_DISPLAY_NAME_TO_ID` — yet every run in the corpus used it.

Nothing reported this until the id dropped out of the CES catalogue on 2026-09-10 and the suite stopped dead. In one window, **8 of 8**: every `claude-sonnet-4.5` run failed with `X_MODEL_NOT_FOUND_ERROR` while every `gpt-5.6-terra` run succeeded.

`build-config.ts` now validates the **resolved** model, not just an explicit override. Checking only the override leaves the more dangerous case unguarded — the default is the value nobody looks at, so it's the one that must be checked.

The supported set is **derived** from the agent frontmatter via `readSupportedModels`, not restated. A second hardcoded list is how this happened in the first place.

Default moved to `gpt-5.6-terra`.

## Supported today

`claude-opus-4.7` · `claude-sonnet-4.6` · `gpt-5.6-sol` · `gpt-5.6-terra`

Note this is **four**, not the three requested — `claude-sonnet-4.6` is still declared in all six agents' frontmatter. Dropping it is a product change (and needs a drift-baseline re-record), so I've left it; say the word and it's a one-line change per agent.

## Validation

`verify-run --self-test` 12/12 · `tsc --noEmit` clean · eslint clean · `eval:drift` 17 contracts · `certify` 152/152 · all four model-resolution paths exercised (default, `--model`, `COR_MODEL`, unsupported-rejected)


nturinski:
Superseded by #1819, which is now the standard for how MSBench resolves and validates its model.

**What carried over.** The mid-run drift detection is ported in eb96f2e. `verifyModel` collected only the first `Set active model to:` line, so a run that switched mid-way verified clean; it now treats more than one distinct model as a mismatch and reports the whole sequence. Falsified the same way this PR did â€” same synthetic log, only that change reverted: `gpt-5.6-sol` / verified / exit 0, versus `gpt-5.6-sol -> claude-sonnet-4.6` / MODEL DRIFT / exit 65. Repeated-but-identical selections still pass.

**What did not, and why.** #1819 puts the supported-model check in `check-agent-drift.ts` *and* `build-config.ts`, and validates the resolved model rather than only an explicit override â€” so that half of this PR is covered, just sited differently.

The default diverges deliberately. This PR chose `gpt-5.6-terra`; #1819 pins `gpt-5.6-sol`, because MSBench now sweeps a deliberately narrowed set of two models, one per family â€” `gpt-5.6-sol` and `claude-sonnet-4.6` â€” rather than everything the agents declare. Four declared models multiply a corpus `README-redteam.md` already wants run per model, and the failures worth catching differ between families rather than between revisions of one. The set lives in `evals/msbench/models.ts` and is proved a subset of the agent frontmatter on every use, so it can be narrowed but not invented.

`COR_MODEL` was dropped for now. With a two-model sweep, `--model claude-sonnet-4.6` per invocation is the whole of the sweep, so the env var buys less than it did against four. Worth revisiting if the set grows â€” the reasoning in this PR still stands.

Thanks â€” the mid-run drift catch was the genuinely load-bearing find here, and it shipped.

PR #1834, opened 2026-09-12
CoR: No dangerous html setting & use "strict" mermaid rendering

MicroFish91:
The local debug plan contains agent-written, user-editable Markdown. Rendering that content through raw HTML insertion allowed attribute-bearing tags and unsafe link protocols to reach the webview.

This change removes `dangerouslySetInnerHTML` from local plan text rendering. A pure parser produces a small node model, and React creates the supported inline formatting, allowed links, exact attribute-free plan tags, and warning icons. Unknown HTML remains text, unsafe protocols do not create anchors, and Mermaid uses strict security mode.

Install links remain consistent with the project-plan view: both use React anchors resolved from the hardcoded prerequisite catalog rather than URLs from plan Markdown.


MicroFish91:
@copilot Resolve any merge conflicts

